PR #39, opened 2026-06-16
fix(security): close read-auth suffix-exemption bypass + arm token-mint guard

aks129:
Urgent follow-up to the merged read-auth gate (#38), which shipped with two bypasses. Mirrors the reviewed fixes from the `security/hardening` branch (#35) **byte-for-byte** so there's no conflict when #35 is reconciled.

## 1. Suffix-exemption bypass — **fixed, effective immediately**
`enforce_tenant_id` exempted discovery endpoints with `endswith('/metadata')` / substring `in` checks. A FHIR read whose resource id is literally `metadata` — `GET /r6/fhir/Patient/metadata` — matched the suffix and **bypassed auth entirely**. Replaced with:
- exact-path match for `/metadata` and `/health`
- prefix match for genuinely-namespaced sub-trees (`/internal/`, `/.well-known/`, `/oauth/`, `/mcp-apps/`, `/demo/`)

A resource read is `/{prefix}/{Type}/{id}` (single `Type` segment), so it can never match a namespaced prefix. Zero caller impact.

## 2. Token-mint oracle — **guard shipped dormant**
`POST /internal/step-up-token` minted a tenant-bound token for **anyone, unauthenticated** — a trivial read-auth bypass (mint a token → read any tenant). Added an `INTERNAL_TOKEN_MINT_SECRET` guard: when set, the caller must present a matching `X-Internal-Secret` (constant-time), else `403`. **Unset ⇒ unchanged (open)**, so deploying this is a no-op.

⚠️ **Activation is NOT in this PR.** The endpoint has 15+ callers — including the browser dashboard (`static/js/r6-dashboard.js`), which cannot hold a secret. Setting `INTERNAL_TOKEN_MINT_SECRET` without reconciling those callers would break the dashboard. Full activation (server-side token issuance / consumer updates) is handled in the `security/hardening` branch. **Until that lands, the token-mint bypass remains open in prod.**

## Tests
`tests/test_read_auth.py`: `Patient/metadata` & `Observation/health` now `401`; exact `/metadata` `/health` still public; mint open when secret unset, `403`/`200` when set. **650 Python tests pass.**

### DIFF
--- a/r6/routes.py
+++ b/r6/routes.py
@@ -12,6 +12,7 @@ Validation: structural checks for required fields. Falls back when external
 validator unavailable. No StructureDefinition or terminology binding validation.
 """
 
+import hmac
 import json
 import logging
 import os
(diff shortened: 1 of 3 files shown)


aks129:
Superseded by #35, which now contains these fixes plus a public-tenant-aware token-mint guard (no demo-dashboard breakage) and the full consent/hardening bundle. Consolidating on #35.

PR #71, opened 2026-07-10
feat(adapters): LangChain example script (bridge already supports it)

### LINKED ISSUE #63: Adapters: LangChain example script (bridge already supports it)
adapters/healthclaw_bridge.py already converts the 27-tool manifest to OpenAI/Gemini formats, and docs/recipes/any-agent-framework.md names LangChain as a supported path — but there's no runnable example like adapters/examples/openai_agent.py. Add adapters/examples/langchain_agent.py: load the manifest, wrap each tool as a LangChain StructuredTool dispatching through HealthClawClient, run a small agent loop against the public desktop-demo tenant. Mirror the style/size of the existing examples (~60 lines, argparse, reference-only, not run in CI).

aanishs:
Closes #63.

Adds `adapters/examples/langchain_agent.py`, mirroring the existing `openai_agent.py` / `gemini_agent.py` (~65 lines, argparse, reference-only, not run in CI): loads the manifest, wraps each tool as a langchain-core `StructuredTool` dispatching through `HealthClawClient`, and runs a bounded tool-calling loop.

## Notes

- **langchain-core v1+** accepts the manifest's JSON Schema directly as `args_schema`, so no pydantic model generation is needed. (Verified empirically: dict schemas are rejected on 0.3.29 and accepted on 0.3.44+ / 1.x, so the docs say `>= 1.0` to keep the guidance unambiguous.)
- Tool dispatch uses `tool_map.get(name)` and returns an explicit error string for an unknown tool name instead of raising `KeyError`.
- The example binds tools via `langchain-openai`'s `ChatOpenAI`; the `to_langchain_tools` helper is model-agnostic, so swapping in another LangChain chat model is a one-line change.
- `docs/recipes/any-agent-framework.md` gains a LangChain section, the three-doors table points at the example, and the stale tool count is corrected **23 → 29** to match `adapters/tools.manifest.json`.

## Scope boundary (deliberate)

Not hardened past the existing OpenAI/Gemini examples — no network timeouts, response-size caps, `X-Human-Confirmed` threading, or arg sanitization. Those are properties of the shared `HealthClawClient` bridge, not this ~65-line reference script, and the issue asks to mirror the existing examples' style/size. Happy to open a separate issue/PR on bridge hardening if useful.

## Verified

Against the public `desktop-demo` tenant:
- all **29** manifest tools wrap cleanly
- a wrapped `fhir_search` invocation returns a redacted, disclaimer-injected Bundle through `/mcp/rpc`
- the unknown-tool guard returns the error string rather than raising

`ruff` clean; full repo suite green (971 Python, 92 Node) after rebase onto current `main`.

Signed off under DCO.

### DIFF
--- a/adapters/examples/langchain_agent.py
+++ b/adapters/examples/langchain_agent.py
@@ -1,0 +1,68 @@
+"""Minimal example: HealthClaw tools on a LangChain tool-calling loop.
+
+Requires: `pip install "langchain-core>=1.0" langchain-openai` (v1 accepts a
+JSON-Schema dict as args_schema directly) and OPENAI_API_KEY. Reference
+example (not run in CI). Guardrails enforced by HealthClaw server-side.
+
+    python adapters/examples/langchain_agent.py \
+        --mcp-base https://mcp-server-production-5112.up.railway.app \
+        --tenant desktop-demo --prompt "List the patient's active conditions"
+"""
+import argparse
+import json
+import sys
+
+sys.path.insert(0, __file__.rsplit("/adapters/", 1)[0])
+from adapters.healthclaw_bridge import (  # noqa: E402
+    load_manifest, HealthClawClient,
+)
+
+
+def to_langchain_tools(manifest, hc):
+    """Manifest -> LangChain StructuredTools relaying through HealthClawClient."""
+    from langchain_core.tools import StructuredTool
+    tools = []
+    for t in (manifest["tools"] if isinstance(manifest, dict) else manifest):
+        def _run(_name=t["name"], **kwargs):
+            return json.dumps(hc.call(_name, kwargs))[:12000]
+        tools.append(StructuredTool.from_function(
+            func=_run, name=t["name"], description=t.get("description", ""),
+            args_schema=t.get("inputSchema", {"type": "object", "properties": {}}),
+        ))
+    return tools
+
+
+def main():
+    ap = argparse.ArgumentParser()
+    ap.add_argument("--mcp-base", required=True)
+    ap.add_argument("--tenant", required=True)
+    ap.add_argument("--step-up-token")
+    ap.add_argument("--prompt", required=True)
+    ap.add_argument("--model", default="gpt-4o")
+    args = ap.parse_args()
+
+    from langchain_core.messages import HumanMessage, ToolMessage
+    from langchain_openai import ChatOpenAI
+
+    hc = HealthClawClient(args.mcp_base, args.tenant, args.step_up_token,
+                          agent_id="langchain-example")
+    tools = to_langchain_tools(load_manifest(), hc)
+    tool_map = {t.name: t for t in tools}
+    llm = ChatOpenAI(model=args.model).bind_tools(tools)
+    messages = [HumanMessage(args.prompt)]
+
+    for _ in range(6):  # bounded tool loop
+        ai = llm.invoke(messages)
+        messages.append(ai)
+        if not ai.tool_calls:
+            print(ai.content)
+            return
+        for tc in ai.tool_calls:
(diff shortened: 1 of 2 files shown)


aks129 (approved):
Reviewed and verified locally — this is exactly how I hoped #63 would land.

What I checked before merging:
- All **29 manifest tools wrap cleanly** as StructuredTools with the raw JSON-Schema dicts on langchain-core 1.x, and dispatch through the wrapper works (ran your `to_langchain_tools` against the manifest with a stub client — your empirical note about the 0.3.29 vs 1.x schema behavior checks out, and pinning the docs to >=1.0 was the right call).
- Mirrors `openai_agent.py` faithfully: same bridge import idiom, same bounded loop, same 12000-char truncation. The `tool_map.get` + explicit unknown-tool error is a small upgrade over the siblings — agents recover from that string; they don't recover from a KeyError.
- The **23 → 29 doc fix** in `any-agent-framework.md` is correct per the manifest — and caught drift our own guard misses (it only scans the site templates). Genuinely useful beyond the PR's scope.
- Full Python suite green on the branch (971).

The explicit scope boundary in the description is appreciated — matching the existing examples rather than gold-plating a reference script was the right read of the issue. If you do want to take a swing at bridge-level hardening (timeouts, response caps, `X-Human-Confirmed` threading), an issue sketching the shape would be very welcome.

(The red Vercel check is fork-preview authorization on our side, not your code.)

Thanks for a genuinely well-made contribution — between this and the Medplum search/fetch work, you clearly get how agent-facing health tooling should behave. More than welcome here anytime.

aanishs:
Thank you! I really like the direction of this project, would love to help any way I can.

aks129:
Really glad to have you — and thank you again. Between the search/fetch contract thinking on the Medplum side and shipping the LangChain adapter here, you clearly have the instincts for agent-facing health tooling, which is exactly the hard part.

Where the project is headed (the working plan lives in [docs/ROADMAP.md](https://github.com/aks129/HealthClawGuardrails/blob/main/docs/ROADMAP.md), issues are the canonical log):

- **Clinical intelligence:** broaden the lab interpreter (TSH, liver panel, vitamin D), unit conversion with cited factors, trend/delta across successive results; risk-adjusted care-gap cadences.
- **CMS-0057-F prior-auth (2027 mandate):** position the existing SDC `$populate`/`$extract` engine as the guardrailed documentation layer for Da Vinci DTR. Design doc is in `docs/design/`; the honest open question there is the CQL bridge.
- **Agent ergonomics:** this is where your recent thinking lands directly — schema-aware, corrective, access-scoped tool errors so agents self-correct instead of guessing. If you wanted to prototype that on our `/mcp` search/fetch surface, it would be a genuinely high-value contribution and dovetails with the Medplum work.
- **Reach:** more framework adapters (LlamaIndex next to your LangChain one), the per-agent quickstarts, and the consumer connect flow (identity-verified Fasten -> read-scoped token -> any agent).

We're aiming for **GA by end of year**, and the two things that most move us toward it are exactly what you're good at: **real-world testing and usability**. Concretely, the most useful help right now:
- Point an agent (LangChain, OpenAI, Claude) at the live `desktop-demo` tenant and tell us where the tool descriptions, error messages, or schemas make the model stumble. Every rough edge you hit is one a real user would.
- Try the [quickstarts](https://github.com/aks129/HealthClawGuardrails/tree/main/docs/quickstarts) as if you'd never seen the repo and open issues where a step is unclear or a default surprises you.

`good first issue` labels mark the low-friction starts, but honestly the highest-leverage thing is unfiltered feedback on where the agent experience breaks. Open issues liberally — that is exactly the signal we need to get to GA. Thanks for jumping in.

PR #81, opened 2026-07-12
fix(proxy): surface upstream FHIR errors instead of empty results and fake 404s

### LINKED ISSUE #74: Upstream proxy converts FHIR errors into empty search results (and auth failures into 404s)
In upstream-proxy mode, `FHIRUpstreamProxy` collapses every upstream error into a success-shaped response:

- `search()` returns `_empty_bundle()` for **any** non-200 — including 400 (bad search param), 401 (expired client-credentials token), 403, 429, 5xx, and network timeouts (`r6/fhir_proxy.py` L186–198). The facade then returns HTTP 200 with `total: 0`.
- `read()` returns `None` for any non-200 that isn't a real 404 (L170–184), and the facade turns `None` into `404 {resource} not found` (`r6/routes.py` L588–592). An expired token or an access-policy denial is reported to the caller as "this resource does not exist."

## Repro (Medplum-in-front recipe)

With `MEDPLUM_BASE_URL` configured per `docs/recipes/healthclaw-in-front-of-medplum.md`:

```
# Direct to Medplum: 400 + OperationOutcome "Unknown search parameter: datetime"
# (Medplum rejects unknown params at query-build time)
GET https://api.medplum.com/fhir/R4/Observation?datetime=ge2024-01-01

# Through HealthClaw: 200 {"resourceType":"Bundle","type":"searchset","total":0,"entry":[]}
GET $BASE/r6/fhir/Observation?datetime=ge2024-01-01
```

The facade forwards all query params upstream (`r6/routes.py` L761), so Medplum correctly rejects the bad one — and the proxy eats the rejection. The same collapse applies to any upstream FHIR server that returns a non-200.

## Why this matters

1. **Agents get a confident wrong answer.** "0 results" and "your query was rejected" require opposite reactions (stop vs. fix-and-retry). Through the MCP layer this compounds: `fhir_search`'s `_mcp_summary.note` tells the agent "No Observation resources found matching criteria" and the connector `search` tool returns `{"results": []}` — for a query that never executed. Upstream servers return OperationOutcomes precisely so clients can self-correct; the guardrail layer currently discards them.
2. **The audit trail records events that didn't happen.** The upstream search path emits `record_audit_event(..., detail='search (upstream): 0 results')` (`r6/routes.py` L789–793) even when the upstream returned 400/401/500. An immutable audit log asserting a successful zero-result search that was actually a rejected or failed request undercuts the audit guarantee.
3. **Failures are undebuggable from the client side.** A 401 from a stale token, a 403 from a Medplum access policy, and a genuinely empty patient record are all indistinguishable to the caller.

## Proposed fix

Happy to open the PR:

1. **Propagate `(body, status_code)` from `search()`/`read()`** — the same shape `create()`/`update()` already return — instead of collapsing to `_empty_bundle()`/`None`. `read()` keeps returning not-found semantics only for a true upstream 404.
2. **Surface failures as a *sanitized* OperationOutcome, never the raw upstream body.** The facade builds a bounded OperationOutcome from an allowlist — `issue[].severity`, `issue[].code`, and diagnostics truncated and stripped to safe content — plus the original HTTP status. Raw passthrough is not safe here: `_rewrite_urls()` only substitutes the configured base URL and would not catch PHI echoes, stack traces, or other internal URLs in upstream diagnostics. Non-JSON failures (timeouts, HTML error pages) get a synthesized OperationOutcome carrying the status class.
3. **Fix the adjacent pre-existing leak in the same pass:** `create()`/`update()` currently return upstream response JSON *without* `_rewrite_urls()` (`fhir_proxy.py` L209, L226), so upstream URLs already leak on write errors today. Route their error bodies through the same sanitizer.
4. **Preserve upstream bundle metadata on success:** the facade rebuilds each entry as `{fullUrl, resource}` only (`routes.py` L768+), dropping `entry.search` — including any `search.mode="outcome"` warning entries an upstream emits. Carry `entry.search` through so upstream warnings survive.
5. **Audit failures as failures:** `record_audit_event` already supports an `outcome` field — failed upstream calls should record `outcome='failure'` with the status (e.g. `detail='search (upstream): rejected 400'`), not a zero-result success.
6. **Tests** in `tests/test_medplum_in_front.py` (which today mocks only successful reads and write gates): a mocked upstream 400 OperationOutcome surfaces through the facade as 400 + sanitized OperationOutcome; a mocked 401 on read does **not** surface as not-found; diagnostics containing an internal URL or patient name do not transit; audit records `outcome='failure'`.

Context: found this while testing the Medplum-in-front recipe after our discussion on medplum/medplum#9616 — this is the same "agents recover from structured errors, not silent failures" point applied one layer down.

aanishs:
Fixes #74.

## What was happening

`FHIRUpstreamProxy` collapsed every upstream failure into a success-shaped response: `search()` returned an empty bundle for any non-200 (a rejected query read as "no results"), and `read()` returned `None` for any non-200 except 404, which the facade reported as `404 not found` (an expired proxy token read as "this resource does not exist"). The audit trail recorded failed searches as zero-result successes (outcome code 0), and 401-failed reads produced no AuditEvent at all — full live transcript in #74.

## What this changes

1. **`search()`/`read()` return `(payload, status)`** — the tuple shape `create()`/`update()` already use. `read()` returns `(None, 404)` only for a true upstream 404.
2. **Every non-2xx becomes a *sanitized* OperationOutcome.** Only the FHIR value-set tokens `severity` and `code` survive from the upstream body; `details.text` is **synthesized from the code**, and the upstream's own free text is never forwarded. This is the load-bearing PHI decision: upstream error text (`details.text`, `diagnostics`) can carry patient names or internal hostnames that no URL-scrubber removes reliably — a `code`-only contract is the one that provably can't leak. Parameter-level correction ("unknown param X, did you mean Y") is produced on the **local** search path instead, where the query is parsed here and the param can be named without echoing anything upstream. Also: malformed issue arrays/fields are tolerated (type-guarded before value-set membership); bodies over 1MB and all 5xx bodies are never parsed; a 2xx body that is unparseable or not a JSON object → 502 `processing`; a search 200 that isn't a Bundle → 502 (else the route would synthesize a misleading empty searchset). `create()`/`update()` error bodies — previously returned raw and un-rewritten — go through the same sanitizer.
3. **Status mapping** — one deliberate judgment call to review:
   - Caller-attributable statuses pass through: 400, 404, 405, 406, 409, 410, 412, 422, 429.
   - **401/403 depend on who owns the upstream credential.** SHARP per-request proxies forward the *caller's* SMART token (`caller_auth=True`), so 401/403 pass through — the caller must see them to re-authenticate. On env-configured proxies (`FHIR_UPSTREAM_URL`/Medplum client-credentials) they map to **502** with issue code `security`: the *proxy's* credentials failed, and a passthrough 401 would tell an agent its own facade token is bad and trigger wrong re-auth loops. Happy to adjust either side of this split.
   - Network failures and 5xx map to 502 / `transient`; upstream logs disclose only `type(exc).__name__` and the resource type — no resource ids, no paths (REVIEW_STANDARDS #1).
4. **Failed accesses are audited**: failed upstream reads — including not-found — searches, creates, and updates now emit AuditEvents with `outcome='failure'` and a status-only detail (`search (upstream): rejected HTTP 400`) — no query params or PHI in detail strings. This closes the gap where 401-failed reads left no audit trace (REVIEW_STANDARDS #3: every access emits an AuditEvent).
5. **`entry.search` preserved with a `mode`/`score` allowlist** when the facade rebuilds searchset entries, so upstream `search.mode="outcome"` warning entries survive the guardrail pass — and those embedded OperationOutcome entries are themselves sanitized through the same allowlist, since `apply_redaction()` doesn't inspect issue-level free text.

## Verification

- **994 tests pass** (repo was at 985 before this branch); **69/69 affected tests pass**; ruff clean on touched files; all touched Python parses with the Python 3.11 grammar.
- **31/31 live public-server integration tests pass** against HAPI FHIR R4 and SMART Health IT R4 (2026-07-12). The exact 400/401 failure cases remain deterministic in the mock-Medplum harness below.
- New regression tests: rejected search surfaces as 400 with synthesized safe text (not an empty bundle); 401 read is not a 404 and IS audited; 404 read IS audited; hostile names, hosts, IPs, and encoded URLs planted across upstream details/diagnostics/extensions never transit (error path AND success-bundle outcome entries); SHARP 401 passthrough; malformed issue shapes and malformed 200 bodies don't crash; oversized bodies not parsed; 5xx bodies never parsed or forwarded; `entry.search` allowlisted; exception-type-only disclosure.
- The diff was also adversarially reviewed across four rounds before opening. Findings from earlier rounds (SHARP credential semantics, malformed-shape crashes, free-text scrubbing being structurally insufficient, missing 404-read audit, success-bundle outcome-entry leak, unbounded body parse) are all addressed above; the final round found no new defects.
- **Live before/after** against the same mock-Medplum harness from #74 (this branch built from the repo Dockerfile, `FHIR_UPSTREAM_URL` set):

| Request | Upstream returned | Facade before (#74) | Facade after (this PR) |
| --- | --- | --- | --- |
| `Observation?datetime=ge2024-01-01` | 400 OperationOutcome "Unknown search parameter: datetime" | **200**, `total: 0` | **400**, OperationOutcome `code: invalid` (synthesized message; upstream text not forwarded) |
| `Patient/pt-1` | 401 OperationOutcome | **404** "not found" | **502**, `code: security`, "Upstream authentication/authorization failed" |
| `Observation?code=4548-4` (control) | 200 searchset | 200, redacted + disclaimed | 200, unchanged |
| Audit trail | — | failed search logged as Success (0); failed read not logged | both logged as Serious failure (8); control logged as Success |

## Known residual (disclosed, not hidden)

On **successful** proxied bundles, `_rewrite_urls()` is a no-op when `FHIR_LOCAL_BASE_URL` is unset (the default), so an upstream `fullUrl`/`Bundle.link`/`meta.source` pointing at an internal host is passed through on the happy path. This is **pre-existing** behavior (predates this PR and affects every proxied read/search today) and lives in the redaction layer, not the error path — so I've deliberately left it out of scope rather than expand an error-handling PR into `redaction.py`. Flagging it as a candidate for its own issue; happy to take that separately.

## Scope boundaries

- The MCP orchestrator's handling of these (now non-empty) error responses is deliberately untouched — that's the `isError`/OperationOutcome passthrough discussion in #75, separate concern.
- `operation()` (used by `$stats`/`$lastn`/`$validate` passthrough) already returned `(body, status)` and is unchanged; routing its error bodies through the sanitizer would be a sensible follow-up but is out of scope here.
- No changes to write semantics, step-up, clinical redaction profiles, or local mode.

### DIFF
--- a/r6/fhir_proxy.py
+++ b/r6/fhir_proxy.py
@@ -116,6 +116,221 @@ def _fetch_medplum_token(client_id: str, client_secret: str) -> str:
 # Timeout for upstream requests (seconds)
 _UPSTREAM_TIMEOUT = float(os.environ.get('FHIR_UPSTREAM_TIMEOUT', '15'))
 
+# ---------------------------------------------------------------------------
+# Upstream error sanitization
+#
+# A failed upstream call must surface as a (sanitized) OperationOutcome with
+# the real status — never as an empty bundle or a fake not-found. Raw upstream
+# error bodies are NOT passed through: diagnostics can embed PHI, stack traces,
+# or internal URLs that _rewrite_urls() would not catch.
+# ---------------------------------------------------------------------------
+
+# Statuses that describe the CALLER's request; these pass through unchanged.
+# Upstream 401/403 are deliberately absent: the caller never authenticates to
+# the upstream directly, so those mean the PROXY's credentials or upstream
+# policy failed — surfaced as 502 so the caller doesn't re-auth in a loop.
+_PASSTHROUGH_STATUSES = frozenset({400, 404, 405, 406, 409, 410, 412, 422, 429})
+
+# FHIR issue-type codes allowed through from upstream OperationOutcomes.
+_SAFE_ISSUE_CODES = frozenset({
+    'invalid', 'structure', 'required', 'value', 'invariant',
+    'security', 'login', 'unknown', 'expired', 'forbidden', 'suppressed',
+    'processing', 'not-supported', 'duplicate', 'multiple-matches',
+    'not-found', 'deleted', 'too-long', 'code-invalid', 'extension',
+    'too-costly', 'business-rule', 'conflict', 'transient', 'lock-error',
+    'no-store', 'exception', 'timeout', 'incomplete', 'throttled',
+    'informational',
+})
+_SAFE_SEVERITIES = frozenset({'fatal', 'error', 'warning', 'information'})
+_MAX_SANITIZED_ISSUES = 5
+# Refuse to parse oversized upstream error bodies (memory-exhaustion guard).
+_MAX_ERROR_BODY_BYTES = 1_000_000
+
+# Upstream free text (issue[].details.text / diagnostics) is NEVER forwarded:
+# it can carry patient names, identifiers, or internal hostnames that no URL
+# regex can reliably strip. We keep only the issue `code` (a bounded FHIR
+# value-set token) and synthesize the human-readable message ourselves from
+# this map. Specific, parameter-level correction ("unknown param X, did you
+# mean Y") is produced on the LOCAL search path, where we parse the query and
+# can name the parameter without echoing upstream text.
+_ISSUE_CODE_MESSAGE = {
+    'invalid': 'The upstream FHIR server rejected the request as invalid.',
+    'structure': 'The upstream FHIR server rejected the request structure.',
+    'required': 'The upstream FHIR server reported a missing required element.',
+    'value': 'The upstream FHIR server rejected a submitted value.',
+    'not-supported': 'The upstream FHIR server does not support this request.',
+    'security': 'Upstream authentication or authorization failed.',
+    'forbidden': 'The upstream FHIR server forbade this request.',
+    'not-found': 'The upstream FHIR server reported the resource was not found.',
+    'conflict': 'The request conflicted with the current state upstream.',
+    'duplicate': 'The upstream FHIR server reported a duplicate.',
+    'too-costly': 'The upstream FHIR server refused the request as too costly.',
+    'throttled': 'The upstream FHIR server is rate-limiting requests.',
+    'processing': 'The upstream FHIR server could not process the request.',
+    'transient': 'The upstream FHIR server had a transient error.',
+    'timeout': 'The upstream FHIR server timed out.',
+    'exception': 'The upstream FHIR server encountered an internal error.',
+}
+
+
+def _message_for_code(code: str) -> str:
+    return _ISSUE_CODE_MESSAGE.get(code, 'The upstream FHIR server returned an error.')
+
+
+def _issue_code_for_status(status: int) -> str:
+    """Map an HTTP status to a FHIR issue-type code."""
+    mapping = {
+        400: 'invalid', 401: 'security', 403: 'security',
+        404: 'not-found', 405: 'not-supported', 409: 'conflict',
+        410: 'deleted', 412: 'conflict', 422: 'processing',
+        429: 'throttled',
+    }
+    if status in mapping:
+        return mapping[status]
+    return 'transient' if status >= 500 else 'processing'
+
+
+def _sanitize_issue_list(raw_issues, fallback_code: str) -> list:
+    """Allowlist-sanitize an OperationOutcome issue array.
+
+    Keeps only ``severity`` and ``code`` from upstream (both bounded FHIR
+    value-set tokens) and synthesizes ``details.text`` from the code — the
+    upstream's own free text is never forwarded. Tolerates malformed shapes
+    (null, dict, string, non-dict entries, unhashable fields) by dropping or
+    defaulting them; a hostile or broken upstream must not crash the error
+    path or leak text through it.
+    """
+    issues = []
+    if not isinstance(raw_issues, list):
+        return issues
+    for issue in raw_issues[:_MAX_SANITIZED_ISSUES]:
+        if not isinstance(issue, dict):
+            continue
+        severity = issue.get('severity')
+        code = issue.get('code')
+        # Membership tests must be type-guarded: a list/dict severity or code
+        # is unhashable and would raise inside `in`.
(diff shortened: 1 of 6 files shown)


aks129:
This is an excellent PR — thank you for the depth. Calling out specifics:

- The core fix is correct and I verified it end to end: `read()`/`search()` now return `(payload, status)`, a true upstream 404 is the only thing that yields `(None, 404)`, and every other non-2xx surfaces as a sanitized `OperationOutcome` with its real status. Both call sites in `r6/routes.py` handle the new tuple — no stray consumer left on the old contract.
- The `sanitize_upstream_error` allowlist (keep only `severity`/`code`, synthesize `details.text` from a fixed map, never forward upstream free text) is the right design, and your reasoning — that URL-scrubbing free text isn't reliable — is exactly right. And you didn't just claim it: `test_medplum_search_rejection_surfaces_not_empty_bundle` plants a patient name and an internal hostname in the upstream body and asserts neither leaks. That's real verification.
- I confirmed the audit-trail fix by querying `AuditEventRecord` myself — a 401 read that previously left *no* trace now writes a `failure` row with a status-only detail. That was the sharpest part of #74.
- Full suite: 1169 passed vs 1146 on `main` — +23 tests, zero regressions. `ruff` clean.

The red `lint` check is **not your code** — it's 4 `F401`s in files you never touched, fixed on `main` by a commit that landed after your base. Since `maintainer-can-modify` is on and it rebases cleanly, **I've rebased your branch onto current `main` for you** so CI re-runs green. Merging once it's clean.

Really appreciate the adversarial self-review rounds and the before/after evidence — that's what makes review fast on our end. Welcome, and thanks for taking on a genuinely subtle bug. 🙏

aanishs:
Thanks @aks129! Appreciate the kind words. I've opened a wider discussion about error handling and a potential path forward: https://github.com/aks129/HealthClawGuardrails/discussions/75

Let me know if this fits

PR #100, opened 2026-07-13
Forms rail (5/9): DocumentReference persistence with embedded PDF bytes

aks129:
## Summary
- Adds `r6/sdc/documents.py::persist_intake_document(tenant_id, subject_ref, pdf_bytes, *, title=None, questionnaire_response_id=None)`: builds a FHIR R4 `DocumentReference` (`status: current`, LOINC 34133-9 "Summarization of episode note" as the document type, `subject`, `date`) whose `content[0].attachment` carries `contentType: application/pdf`, `title`, `size`, and — the point of this task — `data`: the **base64-encoded PDF bytes** (FHIR `Attachment.data` is base64), not just a byte count. Persists tenant-scoped as an `R6Resource`, mirroring `r6/smbp/routes.py::_persist_document_reference`, and emits an audit event.
- When `questionnaire_response_id` is passed, links the DocumentReference back to the source `QuestionnaireResponse` via both `context.related` and `relatesTo` (code `transforms`), so the structured QR and the rendered PDF can be traced to each other.
- Adds `get_document_pdf_bytes(tenant_id, docref_id) -> bytes | None`: loads the DocumentReference (tenant-scoped) and decodes `content[0].attachment.data` back to the exact bytes — this is the round-trip getter Task 7's delivery route will use. Returns `None` if the DocumentReference doesn't exist for that tenant or has no embedded data.

## Test plan
- [x] `tests/test_sdc_documents.py` (13 new tests): retrievable by id under tenant, `contentType`/`size` correctness, `data` is valid base64 that decodes to the exact input bytes, `get_document_pdf_bytes` round-trips exact bytes, missing-docref returns `None`, tenant isolation (same docref id under a different tenant resolves to nothing, both via the getter and via a direct scoped query), `context.related`/`relatesTo` present and correct when `questionnaire_response_id` is given, both omitted when it isn't, and multi-document non-shadowing.
- [x] `uv run python -m pytest tests/ -q` — 1236 passed.
- [x] `uvx ruff check r6/ tests/ scripts/ main.py app.py` — clean.
- [x] `uv.lock` churn reverted before commit (not included in the diff).

### DIFF
--- a/r6/sdc/documents.py
+++ b/r6/sdc/documents.py
@@ -1,0 +1,93 @@
+"""Persist a completed intake PDF as a FHIR DocumentReference.
+
+persist_intake_document() embeds the actual PDF bytes (base64, per FHIR's
+Attachment.data) in content[0].attachment.data — not just a byte count —
+so the rendered PDF (r6/sdc/pdf.py::render_questionnaire_response_pdf) can
+be retrieved later for delivery (Task 7) without re-rendering.
+
+Mirrors r6/smbp/routes.py::_persist_document_reference's DocumentReference
+shape, generalized to carry the bytes and to optionally link back to the
+QuestionnaireResponse the PDF was rendered from.
+"""
+
+import base64
+import json
+from datetime import datetime, timezone
+
+from models import db
+from r6.models import R6Resource
+from r6.audit import record_audit_event
+
+# LOINC "Summarization of episode note" — a generic document-summary type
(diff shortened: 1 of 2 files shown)


aks129:
Superseded by #102 — the whole forms rail is going in as one combined PR for a single review (per maintainer request). This PR's commits (DocumentReference persistence, T5) are included there.

PR #108, opened 2026-07-13
$conformance: grade error fidelity as property seven

### LINKED ISSUE #96: $conformance: grade error fidelity as property seven with local and mock-proxy profiles
Context: follow-up to the accepted [error-fidelity direction](https://github.com/aks129/HealthClawGuardrails/discussions/75). The maintainer explicitly asked for these probes in the existing `r6/conformance/` harness so `$conformance` emits error fidelity as property seven, with a visible failing grade before every fix lands.

## Why this belongs in the scorecard

The six current properties prove PHI redaction, audit, step-up, human-in-the-loop, tenant isolation, and disclaimers. They all grade protections on successful paths. A deployment can still turn a rejected query into “0 results” and receive a clean A.

Error fidelity is the missing property: when the data or tool layer fails, does the system return a truthful, corrective, sanitized failure—or a confident wrong answer?

Current `main` is already partially improved:

- #81 should make the proxy-rejection profile pass: status preserved, OperationOutcome sanitized, audit marked failure.
- Local search still ignores unknown parameters and unsupported modifiers.
- MCP read helpers still flatten non-OK bodies, and returned error objects are not consistently marked `isError`.

That mixed state is exactly what a scorecard should show.

## Required harness change

`r6/conformance/probes.py` currently models each property as boolean-only (`ProbeResult.passed`) and calculates the report grade from the count of passed properties. Error fidelity needs an explicit per-property grade:

- **F:** failure is silent or success-shaped (for example, an ignored filter yielding unfiltered 200 results);
- **C:** failure is signaled but opaque or misclassified (for example, a status-only string or a tool error without `isError`);
- **A:** status and category are truthful, the result is corrective and sanitized, MCP uses `isError` where applicable, and audit reflects the failure.

Extend `ProbeResult`/serialization/rendering with a property grade and profile coverage while preserving the existing boolean contract: the property counts as passed only at A. JSON, text, and shields output must all expose the new property honestly.

## Profile 1 — local (deterministic everywhere)

Run through the same FHIR facade adapter used by the existing in-process probes:

| Check | Request | A | Current failure signal |
| --- | --- | --- | --- |
| Strict unknown parameter | `Observation?datetime=x` + `Prefer: handling=strict` | 400 OperationOutcome naming the key and supported set | 200 unfiltered results |
| Lenient unknown parameter | `Observation?datetime=x` | 200 + self link excluding it + `search.mode="outcome"` warning | 200 with no warning |
| Unsupported modifier | `Observation?code:exact=x` | 400 OperationOutcome in every mode | silently ignored |
| Search audit truth | rejected/ignored search | failure/warning represented without claiming applied filters | success-shaped audit/response |

When an MCP endpoint is supplied to the harness, add the transport check to this profile:

| Check | Request | A | Current signal |
| --- | --- | --- | --- |
| MCP tool failure | invalid FHIR resource/search through `tools/call` | `isError: true` + sanitized OperationOutcome | status-only content with no `isError` |

If no MCP endpoint is configured, report that check as `not_run` and label the property coverage `local-fhir-only`; do not silently count it as a pass.

## Profile 2 — mock upstream proxy (deterministic, never a live third party)

Exercise the full Flask proxy route with an injected/mock upstream:

| Mock response | A |
| --- | --- |
| Search 400 + OperationOutcome | facade returns the mapped status + sanitized OperationOutcome; audit outcome is failure |
| Read 401 | response distinguishes authentication failure from absence; never fake 404 |
| 500 / timeout | truthful 5xx-class OperationOutcome; failed access audited |
| 400 diagnostics containing a synthetic patient name + internal URL | neither value reaches response, log, or audit detail |

The proxy fixture must be injected through a harness/test seam. Do not mutate process-global proxy environment inside a live `$conformance` request, race production traffic, call a real FHIR server, or depend on third-party uptime.

If the running deployment has no safely configured proxy-probe target, `$conformance` reports `proxy: not_run` and grades the property on the local profile, labeled `local-only`. The test suite still runs the mock-proxy profile on every change.

## Baseline and sequencing

This issue should land before the remaining transport/search fixes so the scorecard records the real baseline:

- proxy profile: expected A after #81;
- local search: expected F until the strict/lenient issue lands;
- MCP check: expected C until the transport adapter lands.

With the current six properties passing and error fidelity below A, the baseline PR will move the overall scorecard from A to B (6/7 passing) and make `$conformance` return 503. That is intentional and should be stated in the PR: the property is not “passing” merely because its probes were added. Tests should assert the measured baseline and output contract; follow-up PRs move the property grade rather than deleting or weakening probes.

## Acceptance criteria

- [ ] `PROPERTIES` contains `error_fidelity`, emitted by `$conformance` as property seven.
- [ ] `ProbeResult`, JSON, and text output expose the error-fidelity grade and exactly which profiles/checks ran; shields output exposes the grade and profile coverage while remaining valid against the Shields endpoint schema.
- [ ] Local strict, lenient, modifier, and audit-truth probes are deterministic and use only synthetic data.
- [ ] Optional MCP coverage is explicit: executed when configured, otherwise `not_run`, never an implied pass.
- [ ] Mock-proxy tests cover 400, 401, 500/timeout, audit outcome, and hostile diagnostics without any live third-party dependency.
- [ ] The proxy profile uses dependency injection/test fixtures and cannot alter global production proxy configuration.
- [ ] F/C/A grading boundaries have pure unit tests.
- [ ] Existing six properties and their checks remain unchanged except for generic serialization/rendering support.
- [ ] The known baseline is documented; follow-up fixes produce observable grade movements.

## Scope boundaries

- No real patient data and no live third-party FHIR calls.
- No new error protocol; grade the OperationOutcome/MCP contracts from the accepted direction.
- No bundling the B/C fixes into this issue—the point is to measure them independently.

## Blocked by

None — this is intended to land before the remaining fixes so their grade movement is visible.

I’m happy to drive this one, including the mock-upstream harness. cc @aks129 as requested in discussion #75.

aanishs:
## What & why

Adds **Error Fidelity** as the seventh guardrail property in the existing
`r6/conformance/` harness, following the direction accepted in
[Discussion #75](https://github.com/aks129/HealthClawGuardrails/discussions/75).

The previous six-property scorecard covered successful paths. It could still
award an A while a rejected search was presented as a confident empty result.
This change makes failure-path truthfulness measurable without bundling the
separate local-search or MCP transport fixes.

Closes #96.

## What changed

- Adds conservative F/C/A error-fidelity grading; the property passes only at A.
- Runs deterministic local probes for strict and lenient unknown parameters,
  unsupported modifiers in default/strict/lenient modes, corrective evidence,
  self-link truthfulness, bounded results, and failure/warning audit outcomes.
- Adds optional authenticated Streamable HTTP MCP coverage. Configured profiles
  run independently, report `run`/`not_run`, and cannot be hidden by another
  profile crashing.
- Adds an injected mock-upstream profile covering sanitized 400, auth failure,
  500, timeout, audit outcomes, and hostile name/internal-URL leakage without a
  live third-party dependency or process-global proxy mutation.
- Exposes grade, coverage, profiles, and exact check names in `ProbeResult`, JSON,
  and text. Shields output exposes grade and coverage while retaining only the
  standard endpoint-badge fields.
- Updates the public scorecard and documentation from six properties to the
  honest initial baseline: Grade B (6/7), Error Fidelity F.

## Intentional baseline

The new property is deliberately failing on current `main`:

- local F: unsupported filters are still silently accepted;
- optional MCP C: the tool reports a failure but does not yet return the fully
  corrective `isError: true` OperationOutcome contract;
- injected proxy A: the sanitized proxy error work already on `main` satisfies
  that profile.

Accordingly, `$conformance` returns 503 at B. Follow-up fixes should move the
measured profile grades rather than weaken or delete probes.

## Safety and scope

- Synthetic, subject-bounded searches only; no real patient data.
- No live upstream FHIR calls.
- No changes to write semantics, step-up, redaction, or the underlying
  local-search/MCP behavior being measured.
- Hostile response evidence is checked fail-closed across parsed and raw bodies;
  unsafe `datetime` → `_lastUpdated` suggestions and invented supported
  parameters cannot earn A.

## Verification

- `1392 passed, 1 skipped` locally on CPython 3.11.
- `1390 passed, 1 skipped` in Linux/ARM64 on the Mac Mini's Colima runtime.
- MCP service: `tsc --noEmit` clean; 142/142 Jest tests pass on Node 20.
- Real Colima Compose stack: migration and seed exited 0; Flask and MCP health
  returned 200; missing/wrong/correct MCP bearer auth returned 401/401/200.
- Live CLI with protected MCP returned the intentional B/F baseline with
  `local+mcp` coverage and MCP C.
- Synthetic hostile names, internal hostname, and transport token appeared zero
  times in Flask or MCP container logs.
- `ruff`, Python 3.11 AST parsing, and `git diff --check` are clean.
- Final independent standards, Issue #96/Discussion #75, and Fable adversarial
  reviews found no remaining issues; sibling-evidence smuggling is covered by
  regression tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,8 @@ properties pass while local search still accepts unsupported filters. The same
 harness runs against the Flask test client as a **CI baseline**
 (`tests/test_guardrail_conformance.py`), so grade movement is explicit rather
 than hidden. `--json` emits a machine-readable report; `--mcp-url` also probes
-MCP `tools/call` error signaling. Library API:
+MCP `tools/call` error signaling. For an authenticated MCP deployment, set
+`MCP_AUTH_TOKEN` or pass `--mcp-auth-token`. Library API:
 `from r6.conformance import LiveProbeClient, ProbeContext, run_conformance`.
 
 ## Install as a Claude Plugin
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ AI Agent ──▶ MCP Server ──▶ Guardrail Proxy ──▶ Any FHIR Serve
 ## Prove it: guardrail conformance
 
 The guardrails are **verifiable, not marketing.** A runnable harness probes any
-deployment with synthetic data and emits a scorecard across all six properties —
+deployment with synthetic data and emits a scorecard across all seven properties —
 run it against your own instance (or ours):
 
 ```bash
@@ -123,10 +123,11 @@ python scripts/guardrail_conformance.py \
 
 ```text
 HealthClaw Guardrail Conformance — https://app.healthclaw.io [tenant=desktop-demo]
-  Grade: A   (6/6 properties)
+  Grade: B   (6/7 properties)
   [PASS] PHI Redaction            [PASS] Human-in-the-Loop
   [PASS] Immutable Audit Trail    [PASS] Tenant Isolation
   [PASS] Step-Up Authorization    [PASS] Medical Disclaimers
+  [FAIL] Error Fidelity — F (local-fhir-only)
 ```
 
 Or hit the **one-URL self-test** on any running deployment — no token needed, it
@@ -136,9 +137,13 @@ self-tenants internally and returns 200 at Grade A (503 otherwise):
 curl "https://app.healthclaw.io/r6/fhir/\$conformance?format=text"
 ```
 
-The same harness runs against the Flask test client as a **CI gate** (`tests/test_guardrail_conformance.py`),
-so a guardrail regression fails the build. `--json` emits a machine-readable
-report. Library API: `from r6.conformance import LiveProbeClient, ProbeContext, run_conformance`.
+The initial error-fidelity baseline is intentionally B: the original six
+properties pass while local search still accepts unsupported filters. The same
+harness runs against the Flask test client as a **CI baseline**
+(`tests/test_guardrail_conformance.py`), so grade movement is explicit rather
+than hidden. `--json` emits a machine-readable report; `--mcp-url` also probes
+MCP `tools/call` error signaling. Library API:
+`from r6.conformance import LiveProbeClient, ProbeContext, run_conformance`.
 
 ## Install as a Claude Plugin
 
@@ -215,7 +220,7 @@ Tool names use underscores (not dots) for Claude Desktop / MCP client compatibil
 | `fhir_lastn` | Most recent N observations per code |
 | `fhir_interpret_labs` | Lab reference-range interpretation (`$interpret`) — decision support, not diagnosis |
 | `care_gaps` | Preventive-care gaps (`$care-gaps`) — screenings/immunizations that may be due, from the patient's own records |
-| `guardrail_conformance` | Run the guardrail conformance self-test — graded A–F scorecard across all six properties |
+| `guardrail_conformance` | Run the guardrail conformance self-test — graded A–F scorecard across all seven properties |
 | `fhir_permission_evaluate` | R6 Permission access control evaluation |
 | `fhir_subscription_topics` | List available SubscriptionTopics |
 | `questionnaire_populate` | SDC `$populate` — pre-fill a Questionnaire for a subject |
--- a/docs/demos/forms-rail-run-of-show.md
+++ b/docs/demos/forms-rail-run-of-show.md
@@ -33,7 +33,8 @@ consent. That is the whole pitch — safety you can watch fail closed.
    (Penicillin) so the review page has something to confirm.
 3. Mint a step-up token: `POST /r6/fhir/internal/step-up-token {"tenant_id": "desktop-demo"}`.
 4. Open the guardrail scorecard in a browser tab you can flip to:
-   `GET /r6/fhir/$conformance?format=text` — Grade A. (Credibility anchor.)
+   `GET /r6/fhir/$conformance?format=text` — the intentional Grade B baseline:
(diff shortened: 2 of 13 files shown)


aks129 (approved):
Reviewed the code, not just the green checks — this is a strong contribution, @aanishs. 🎯

**What I verified:**
- The `error_fidelity` probe is designed exactly right: it grades failure *behavior* (a nonexistent resource must return a typed error, never a fake-empty/404; hostile URLs and PHI-class values must not be reflected back in errors), which is the failure-path counterpart to the six happy-path guarantees. Great follow-through from #81.
- It handles both JSON and SSE (`text/event-stream`) MCP Streamable-HTTP responses — the transport nuance that trips most probes.
- The grading (`A`/`C`/`F`, weakest-executed-profile wins) is **additive**: it reports as a suffix (`error fidelity A, full`) rather than folding into the 6/6 pass count, so it surfaces the new dimension **without silently dropping a deployment's existing A–F grade**. That's the right call for a graded 7th property, and it means this won't regress the CI conformance gate or the live `$conformance` scorecard.
- Full CI green: python, postgres, node, playwright, compose-smoke, compliance-gates, lint.

**Ready to merge.** Thank you for making the guardrail contract prove itself on the failure paths too — 'the agent gets a clean rejection and recovers, instead of hearing silence and fabricating no data' is exactly the property the field is missing.

Issue #138, opened 2026-07-16
Connector registry — implement the marketplace (spec #125)


aks129:
Part of #134. Implement the pluggable connector registry from the design spec (`docs/superpowers/specs/2026-07-16-careagents-connector-marketplace-design.md`, #125).

- `careagents/connectors/`: `Connector` protocol + `StartResult` (Connect/Import/Soon); registry + `register_all`.
- Migrate existing `sample` and `fasten` into connector classes (no behavior change).
- `Connection` gains `connector` id + small `meta` (e.g. provider).
- Hub: marketplace grid rendered from the registry; live tiles first, then import, then honest "coming soon" with notify-me.
- Test: adding a connector is one file + one registry line, no template edits.

**Acceptance:** the hub shows a real connect-anything menu driven by the registry; sample/fasten behave exactly as today through it.

Issue #140, opened 2026-07-16
Wearables connector: Apple Health via Open Wearables (live)


aks129:
Part of #134. The Aug 18 Apple Health headline. Open Wearables 0.6.3 ingests Apple + Health Connect data and owns the phone bridge — so Apple Health is a live path with no native code on our side.

- Stand up / point at an Open Wearables sidecar with Apple (+ Oura) providers enabled and client IDs configured.
- Implement the `wearable` connector (depends on the registry issue + the client reconciliation): provider picker (Apple Health, Oura, …), OAuth kickoff via `/wearables/oauth/start` with a CareAgents-minted tenant token, pending→active poll as samples land.
- Grey out providers whose client IDs aren't configured, with a clear reason.

**Acceptance:** in the hub, "Connect Apple Health" runs the real OW flow and the connection flips to active once data ingests; the agent can answer over that data.

aks129:
Closed in a lean triage (23 Sep): blocked upstream (Open Wearables OAuth/developer auth), so any work here is speculative. Reopen when the provider path is live.

PR #149, opened 2026-07-18
feat(careagents): connector marketplace — pluggable registry + Apple Health (closes #138 #140)

aks129:
The other Aug-18 headline: CareAgents' connection step becomes a **universal front door** over a pluggable connector registry.

## What's here
- **`careagents/connectors.py`** — a `CATALOG` of 7 sources with honest tiers (`live` | `import` | `soon`) and a `start()` dispatcher that returns a plan (create tenant / route to a connect URL / record waitlist intent). Adding a source = one entry + (for a live flow) one branch. No template edits.
- **`GET /api/connections/catalog`** + generic **`POST /api/connections/<id>`** replace the hardcoded sample/fasten endpoints — same response shapes, so existing flows are unchanged. Coming-soon tiles **record intent, never dead-end**.
- **Hub marketplace grid** (home.html/js/css) replaces the two buttons; a tile handles provider pick → connect_url open → pending poll.

## Apple Health / wearables — honest by design
Apple Health, Oura, Whoop, Garmin, Fitbit, Strava are surfaced as wearable providers (Open Wearables owns the phone bridge, so **no native code**). But HealthClaw's wearables OAuth currently **503s** — Open Wearables' authorize needs developer-session auth that isn't wired upstream yet. So the wearable tile is **gated on `CARE_WEARABLES_ENABLED`**: live OAuth only where the sidecar is actually wired; otherwise it shows "coming soon" with the provider list visible. We never route users into a 503.

## Tests
catalog lists Apple Health; wearable soon-by-default / live-when-enabled; coming-soon records intent; unknown connector 404. **Full suite: 1480 passed.**

Implements #125's spec.

### DIFF
--- a/careagents/app.py
+++ b/careagents/app.py
@@ -21,6 +21,7 @@ from flask import (Flask, Response, jsonify, redirect, render_template,
                    request, session, url_for)
 
 from careagents.accounts import (AccountService, AuthError, new_binding_code)
+from careagents import connectors
 from careagents.agent import run_turn, run_turn_to_message
 from careagents.config import Config
 from careagents.healthclaw import HealthClawClient, HealthClawError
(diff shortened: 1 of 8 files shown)


aks129:
Also includes **#145**: careagents now accepts an `ANTHROPIC_OAUTH_TOKEN` (Bearer + oauth beta header) as an alternative to an API key — so the demo can run on real Claude (claude-sonnet-5) instead of rate-limited Gemini. Drop the token (or a plain `ANTHROPIC_API_KEY`) into careagents.cloud's env to activate; `llm.py` already prefers Anthropic.

PR #152, opened 2026-07-19
fix(mcp): preserve error fidelity across transports

### LINKED ISSUE #93: MCP tool errors: preserve sanitized OperationOutcomes and set isError consistently across transports
Context: follow-up to the accepted [error-fidelity direction](https://github.com/aks129/HealthClawGuardrails/discussions/75). The upstream proxy half has landed in #81.

## What is already fixed

- #81 makes the Flask proxy preserve failure status and return a bounded, sanitized OperationOutcome instead of an empty bundle or fake 404.
- #83 puts every backend fetch behind a timeout and converts `BackendTimeoutError` into a structured `{error: "backend_timeout", ...}` result.

Those fixes make the backend error useful. The MCP layer still loses or misclassifies it.

### 1. Read-path helpers discard the body

`readResource`, `searchResources`, `validateResource`, `getContext`, `connectorSearch`, and `connectorFetch` in `services/agent-orchestrator/src/tools.ts` still turn a non-OK response into a status-only object such as:

```ts
return { error: `Search failed with status ${resp.status}` };
```

The Flask response may already contain the safe corrective information the agent needs: FHIR issue code, whether this is authentication versus absence, and locally authored guidance for malformed search input. That information is discarded one hop after the guardrail produced it.

### 2. Returned failures are success-shaped on every MCP transport

- Streamable HTTP and SSE serialize every returned object into `content` without `isError`.
- Stdio is only a partial reference: its `catch` path sets `isError: true`, but an ordinary returned `{error: ...}` object follows the success path there too.
- Non-timeout API failures that throw still become JSON-RPC `-32603` in the HTTP transports.

The [MCP tools specification](https://modelcontextprotocol.io/specification/2025-06-18/server/tools) distinguishes protocol errors from tool-execution errors: API, input-validation, and business-logic failures belong in a tool result with `isError: true`, so the model can see the corrective content and retry. Unknown tools and malformed JSON-RPC remain protocol errors.

### A. Preserve a bounded backend failure

Add one helper for non-OK backend responses, used first by the read-path methods above:

- If the body is an OperationOutcome, retain the bounded FHIR error shape produced by the Flask guardrail boundary: `resourceType`, at most N issues, allowlisted issue fields, and a response-size cap.
- Do not reintroduce raw upstream text. #81 already replaces upstream free text with locally synthesized text; this layer must not bypass that contract.
- If the response is non-JSON, malformed, or not an OperationOutcome, synthesize a minimal failure containing the HTTP status and a stable error code.
- Preserve the HTTP status in the tool error envelope.

### B. Add one transport adapter

Create a shared adapter that classifies an executor result and renders a `CallToolResult`:

- tool/API failure → `content` plus `isError: true`;
- success → normal `content` with no error flag;
- unknown tool / malformed protocol request → JSON-RPC protocol error.

Use the same adapter from stdio, SSE, and Streamable HTTP. Do not duplicate heuristics in three handlers.

### C. Preserve bridge compatibility

`POST /mcp/rpc` keeps its existing JSON-RPC envelope and continues returning the raw executor payload. Its read failures become richer, but it is not wrapped in MCP `content`/`isError`; that wrapping belongs only to actual MCP transports.

## Acceptance criteria

- [ ] A backend 400 OperationOutcome through `fhir_search` reaches the model with its sanitized issue array and original status; it is not reduced to `"Search failed with status 400"`.
- [ ] The same tool failure through stdio, SSE, and Streamable HTTP returns `isError: true` with equivalent content.
- [ ] The #83 `backend_timeout` result is also marked `isError: true` on all three transports.
- [ ] A non-JSON backend failure becomes a bounded synthesized tool error and never leaks an HTML page, stack trace, URL, token, or raw exception text.
- [ ] Unknown tools and structurally malformed MCP requests remain JSON-RPC protocol errors, not tool results.
- [ ] `/mcp/rpc` keeps its current envelope and success behavior.
- [ ] Transport-level tests cover returned errors as well as thrown exceptions; `tools.test.ts` alone is not sufficient.

## Scope boundaries

- Read-path body preservation first; do not redesign action/write semantics in this issue.
- `commitWrite()` response handling is an adjacent wart, not part of this slice.
- No new error vocabulary: reuse FHIR OperationOutcome and the stable timeout/error codes already emitted by the backend.

## Blocked by

None — #81 and #83 are already merged.

Happy to collaborate or pick this up if it remains unclaimed. cc @aks129 as requested in discussion #75.

aanishs:
Closes #93.

## What changed

- Preserve non-2xx read-path failures as a bounded error envelope containing the original HTTP status and a sanitized FHIR `OperationOutcome`.
- Apply the shared failure parser to `context_get`, `fhir_read`, `fhir_search`, `fhir_validate`, and the connector-compatible `search` / `fetch` tools.
- Use one MCP result adapter for stdio, legacy SSE, and Streamable HTTP so returned tool failures and `backend_timeout` consistently set `isError: true`.
- Keep unknown tools and malformed MCP calls as JSON-RPC protocol errors (`-32601` / `-32602`).
- Keep `/mcp/rpc` compatibility: it still returns the raw executor payload, now with the richer bounded failure.
- Pass the already-parsed Express body into the SDK SSE handler; without this, real SSE `tools/call` requests failed because the request stream had already been consumed.

## Security contract

The MCP layer never forwards arbitrary backend error text. It retains only allowlisted FHIR severity/code tokens, caps the body at 64 KiB and issues at five, drops diagnostics/extensions/extra fields, and synthesizes human text locally. The only preserved corrective text is the exact finite grammar generated by the local general-search and `AuditEvent` search contracts; hostile lookalikes fall back to code-derived text.

Malformed JSON, HTML, oversized bodies, thrown exceptions, URLs, tokens, stack text, and PHI-shaped hostile values are covered by adversarial tests and do not reach MCP content.

## Verification

- Agent orchestrator (Node 22): **168 passed**
- TypeScript: `tsc --noEmit` clean
- Full Python suite: **1,481 passed, 1 skipped**
- Ruff: clean
- Real transport tests exercise Streamable HTTP, SSE wire traffic, and a spawned stdio server for returned failures, thrown failures, timeouts, and unknown tools.
- DCO sign-off present; tool/version counts unchanged.

### DIFF
--- a/services/agent-orchestrator/src/backend-failure.ts
+++ b/services/agent-orchestrator/src/backend-failure.ts
@@ -1,0 +1,257 @@
+import type { Response } from "node-fetch";
+
+const MAX_BACKEND_ERROR_BYTES = 64 * 1024;
+const MAX_OPERATION_OUTCOME_ISSUES = 5;
+
+const SAFE_SEVERITIES = new Set(["fatal", "error", "warning", "information"]);
+const SAFE_ISSUE_CODES = new Set([
+  "invalid",
+  "structure",
+  "required",
+  "value",
+  "invariant",
+  "security",
+  "login",
+  "unknown",
+  "expired",
+  "forbidden",
+  "suppressed",
+  "processing",
+  "not-supported",
+  "duplicate",
+  "multiple-matches",
+  "not-found",
+  "deleted",
+  "too-long",
+  "code-invalid",
+  "extension",
+  "too-costly",
+  "business-rule",
+  "conflict",
+  "transient",
+  "lock-error",
+  "no-store",
+  "exception",
+  "timeout",
+  "incomplete",
+  "throttled",
+  "informational",
+]);
+
+const LOCAL_SEARCH_PARAMETERS = [
+  "patient",
+  "code",
+  "status",
+  "_lastUpdated",
+  "_count",
+  "_sort",
+  "_summary",
+  "context-id",
+] as const;
+const AUDIT_SEARCH_PARAMETERS = ["context-id", "entity-type", "_count"] as const;
+const LOCAL_SEARCH_PARAMETER_SETS = [
+  LOCAL_SEARCH_PARAMETERS,
+  AUDIT_SEARCH_PARAMETERS,
+] as const;
+const SAFE_UNSUPPORTED_SEARCH_KEYS = new Set(["date", "datetime"]);
+const SAFE_MODIFIER_TOKENS = new Set([
+  "above",
+  "below",
+  "contains",
+  "exact",
+  "identifier",
+  "in",
+  "iterate",
+  "missing",
+  "not",
+  "not-in",
+  "of-type",
+  "text",
+  "type",
+  "frobnicate",
+]);
+const FIXED_LOCAL_SEARCH_MESSAGES = new Set([
+  ...[...new Set([...LOCAL_SEARCH_PARAMETERS, ...AUDIT_SEARCH_PARAMETERS])].map(
+    (parameter) => `Repeated ${parameter} parameters are not supported.`
+  ),
+  "_count must be a non-negative integer.",
+  "_sort must be _lastUpdated or -_lastUpdated.",
+  "_summary only supports count.",
+  "Patient reference must match Patient/{id}.",
+  "_lastUpdated must be a valid ISO datetime with optional ge/le/gt/lt prefix.",
+  "context-id must be a valid FHIR id.",
+]);
+
+const ISSUE_CODE_MESSAGES: Record<string, string> = {
+  invalid: "The FHIR backend rejected the request as invalid.",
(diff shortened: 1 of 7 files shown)


aanishs:
CI note: the two red statuses are not regressions from this branch.

- `playwright-tests`: the PR fails only `e2e/tests/landing.spec.ts:23`, which still expects a `Try the Live Dashboard` link. The landing page on the exact base commit (`19a55a9`) now exposes `Try CareAgents` instead, and that base commit’s own main CI run fails the same Playwright job: https://github.com/aks129/HealthClawGuardrails/actions/runs/29669657762
- `Vercel`: deployment requires authorization from a member of the maintainer’s Vercel team.

All checks that exercise this change are green, including Node tests, Python tests, lint, Postgres, compose smoke, compliance gates, secret scan, dependency audit, and the Claude standards review.

PR #169, opened 2026-07-19
test: guard SAFE_MODIFIER_TOKENS drift across Python and TypeScript

### LINKED ISSUE #156: Add a drift guard for the Python/TypeScript SAFE_MODIFIER_TOKENS allowlists
## Problem

Two allowlists must stay identical, in two languages, with nothing coupling them:

- `_SAFE_MODIFIER_TOKENS` — `r6/routes.py:730`
- `SAFE_MODIFIER_TOKENS` — `services/agent-orchestrator/src/backend-failure.ts:57`

Both currently hold the same 14 tokens. They must match because the MCP layer validates the corrective text the Python layer generates: if Python emits a token Node doesn't know, Node correctly falls back to a **generic** message and the specific, useful corrective text is silently lost. The failure is quiet — no test fails, agents just get vaguer errors.

This nearly bit us: #150 (Python side) and #152 (Node side) were reviewed independently against the pre-merge `main` and merged back to back. They happened to agree. Nothing guaranteed that.

## Requirements

- [ ] Add a test that fails when the two sets diverge.
- [ ] It must read **both** sources — parsing the TS file from Python, or exporting the Python set to a fixture the Node suite checks, or generating both from one shared source of truth.
- [ ] Failure message should name the specific tokens that differ and which side is missing them.
- [ ] Mirror the explanatory comment onto the TS side. Python documents `frobnicate` as *"Fixed synthetic token used by the public issue contract"*; the TS list has no comment, which already caused a reviewer to flag it as unexplained dead code.

## Suggested approach

Simplest that works: a pytest that regex-extracts the token set from `backend-failure.ts` and asserts equality with `_SAFE_MODIFIER_TOKENS`. Brittle to formatting, but the file changes rarely and a clear failure message makes the brittleness cheap. A generated shared constant is cleaner if you want to invest more.

Same reasoning applies to the fixed-message grammar (`FIXED_LOCAL_SEARCH_MESSAGES`) if you want to extend the guard.

## Files

- `r6/routes.py:730`
- `services/agent-orchestrator/src/backend-failure.ts:57`
- `tests/` — new test

## Verification

```bash
uv run python -m pytest tests/test_<your_new_test>.py -v
```

Prove it works: temporarily add a token to one side and confirm the test fails with a useful message.

---
Good first issue — self-contained, no guardrail semantics to learn. Read [docs/agent-task-guide.md](../blob/main/docs/agent-task-guide.md) first.

gitcommit90:
## Summary
Guard dual-language `SAFE_MODIFIER_TOKENS` allowlists so Python and TypeScript cannot silently drift.

## Changes
- New pytest that extracts the TS `Set([...])` literals and asserts set-equality with Python `_SAFE_MODIFIER_TOKENS`
- Clear failure when the TS block is missing
- Comment linking the two sources of truth

Closes #156


aks129 (approved):
Verified locally rather than relying on CI (fork PRs do not run the full suite here). This is a clean implementation of #156.

**What I checked:**

- `pytest tests/test_safe_modifier_tokens_drift.py -v` → 2 passed
- **Adversarially, both directions.** Added a token to the TS set only → fails. Removed `iterate` from the TS set → fails. Restored → passes. The guard is not vacuous.
- `uvx ruff check r6/ tests/ scripts/ main.py app.py` → clean
- `npx tsc --noEmit` → clean

**What I liked:**

The `test_safe_modifier_tokens_parser_rejects_missing_block` test is the thoughtful part. The obvious failure mode for a regex-parses-another-language test is that a formatting change silently yields an empty set and the guard passes forever while protecting nothing. Raising `AssertionError` on both "block not found" and "no string literals found," and then testing that behaviour, closes exactly that hole. That was not in the issue — good instinct.

The failure message naming which side each missing token is on is also the right call; that is the information someone needs at 2am when this fires.

And thank you for adding the explanatory comment to the TS side. That token had already been flagged once in review as unexplained dead code in a security allowlist, which is why the issue asked for it.

Merging.

Issue #225, opened 2026-08-01
ux: copy makes three claims the code contradicts


aks129:
For a product whose brand is honesty about what it does, these are the highest-cost bugs in the repo.

1. **`careagents/templates/chat.html:23`** — every agent greets with 'I can read the health records in this space (sample data for now)', including agents bound to **real** Fasten/MEDENT records.
2. **`careagents/templates/home.html:135-137`** — the consent card says 'Leaving: email support@healthclaw.io and we'll delete your data', while the self-serve Disconnect and Delete buttons shipped in #203 sit on the same page. We are understating our own best feature in the one place users read carefully. *(Requires a `CONSENT_VERSION` bump.)*
3. **Connector tiles** advertise an 'import' tier (SHL paste, file upload) whose `start()` returns `{"soon": True}`.

Precedent for the rule being enforced: #170 ('stop advertising an unimplemented summary-only privacy control').

From the 2026-08-01 UX audit.

Issue #182, opened 2026-07-20
Audit-write failure silently rolls back the caller's transaction (endpoint still reports success)


aks129:
## What happened

While preparing the v1.9.0 demo against a stale local SQLite DB (pre-`0003_audit_outcome_detail` schema), every `record_audit_event` raised `OperationalError` (missing column). The except path in `r6/audit.py` then ran a **full** `db.session.rollback()` — wiping the caller's uncommitted work — while swallowing the error. Net effect: `POST /r6/fhir/internal/seed` returned `201 {"created_count": 7}` and persisted **zero** resources and **zero** audit rows, with only a log line to show for it.

## Why it matters

- The docstring promises the opposite: *"Uses a nested transaction (SAVEPOINT) so that audit failures do not roll back the caller's already-committed work."* The code opens `begin_nested()` but the exception handler calls plain `db.session.rollback()`, not `nested.rollback()` — so it takes the whole session down.
- It converts an audit failure into **silent data loss with a success status** — the exact fail-loud principle inversion this project exists to prevent.
- On prod (Alembic-managed Postgres) the trigger is unlikely, but any transient audit-insert failure (constraint, disk, width) would do the same thing.

## Repro

1. Point the engine at a pre-0003 SQLite DB (or any DB where `audit_events` inserts fail).
2. `POST /r6/fhir/internal/seed {"tenant_id": "probe"}` → 201, created_count 7.
3. `GET /r6/fhir/Patient` for that tenant → total 0. Log shows `Failed to record audit event: OperationalError` (42×).

## Suggested direction

- In the except path, roll back **only the savepoint** (`nested.rollback()`), preserving the caller's transaction — matching the docstring.
- Decide the posture explicitly: if audit cannot be recorded, should the guarded operation *fail loud* instead of proceeding un-audited? ("every access emits an AuditEvent" is a graded conformance property — silently proceeding without audit may be the bigger sin.)
- A test that breaks audit inserts and asserts (a) callers' writes survive or (b) the operation fails loudly — never success-with-nothing-persisted.

Found during v1.9.0 release-demo prep, 2026-07-19.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #239, opened 2026-08-01
feat(careagents): the agent remembers — durable chat history (#222)

### LINKED ISSUE #222: ux: the persistent agent has amnesia — chat history is never rendered on reload
The product's core promise is a *persistent* health agent. Today history lives in process memory (`careagents/app.py:49`) and `chat.html` always renders only the canned greeting. A user who returns tomorrow — or reloads, or hits a deploy — sees an empty chat and an agent that forgot them.

**The fix is much smaller than it looks.** `ConversationMessage` already exists (`r6/command_center/models.py`), is tenant-scoped, sits on the HealthClaw side of the PHI boundary, and is already covered by `purge_tenant`. `GET`/`POST /command-center/api/conversations` already exist.

So this is wiring, not architecture — and it keeps the 'CareAgents stores no PHI' invariant intact rather than renegotiating it.

**One known snag:** `recent_conversations` uses `to_dict`, which truncates `text` to 500 chars — needs a full-text path for replay.

From the 2026-08-01 UX + architecture audits.

aks129:
Closes #222. **Stacked on #238** — review that first; this PR's diff is only the persistence work.

The product promises a *persistent* health agent. It kept history in process memory and rendered only a canned greeting, so every deploy, restart, or reload made it forget the person entirely — and the chat page greeted a returning user as a stranger.

## Almost no new infrastructure

History now lives in HealthClaw, per tenant, reusing **`ConversationMessage`** and the existing **`/command-center/api/conversations`** endpoints. The table is already tenant-scoped, already behind the guardrails, and already covered by `purge_tenant` — so "delete my records" removes conversations too, for free.

## Two deliberate design calls

**Not stored in the CareAgents database.** A conversation about someone's records is PHI-adjacent, and this app stores no PHI. Putting transcripts in its tables would have traded that boundary away for a caching convenience.

**Only user and assistant text is persisted — never tool calls or their results.** Those are point-in-time readings of records that may since have changed; replaying them as history would let a stale value masquerade as current truth. A cold process re-reads instead. This also made the implementation considerably smaller.

## Changes

- `healthclaw.py`: `log_message` / `recent_messages`, both non-raising — losing a transcript must never break the conversation in progress.
- `app.py`: `load_history()` rehydrates a cold conversation. The **question is recorded before streaming the answer**, so it survives someone closing the tab mid-reply. Idle eviction (from #238) is now safe rather than lossy.
- `chat.html`: replays the real conversation with a "picking up where you left off" divider. Also drops **"(sample data for now)"**, which was shown to every agent including real-record ones (part of #225).
- Telegram/iMessage share the same store — a conversation started on one surface continues on the other.
- `recent_conversations(full=True)`: the dashboard keeps its 500-char preview, but a rehydrating client needs what was actually said; a truncated replay would quietly rewrite history.

## Tests

Persisted-to-the-tenant, returning-user-sees-their-conversation, cold-process-rehydrates, and storage-outage-doesn't-break-chat. The test helper drains the SSE stream as a browser does — without that the turn never runs to completion.

**1538 passed, 1 skipped.** Ruff (CI-pinned): clean.

### DIFF
--- a/careagents/app.py
+++ b/careagents/app.py
@@ -63,13 +63,26 @@ def create_app(config: Config | None = None,
     # memory for the life of the process.
     hist_seen: dict[str, float] = {}
 
+    def load_history(tenant: str) -> list:
+        """The in-memory conversation, rehydrated from HealthClaw if cold.
+
+        Process memory is a cache, not the record. Deploys, restarts and idle
+        eviction all empty it, and before this the agent simply forgot the
+        person — a strange trait for a product promising a *persistent* health
+        agent. Call under hist_lock.
+        """
+        history = histories[tenant]
+        if not history:
+            history.extend(hc.recent_messages(tenant))
+        touch_history(tenant)
+        return history
+
     def touch_history(tenant: str) -> None:
         """Mark a conversation live; release idle ones once we're holding many.
 
         Deliberately simple: only sweeps when the map is already large, so the
-        common path is a single dict write. Chat history moves to durable
-        per-tenant storage in a follow-up (#222) — this is the bound that keeps
-        the process healthy until then.
+        common path is a single dict write. Evicting is now safe rather than
+        lossy — a dropped conversation reloads from HealthClaw on next use.
         """
         seen_at = time.time()
         hist_seen[tenant] = seen_at
@@ -401,8 +414,11 @@ def create_app(config: Config | None = None,
         if not ctx:
             return redirect(url_for("home"))
         p = PERSONAS.get(ctx["agent"]["persona"], PERSONAS[DEFAULT_PERSONA])
+        # Show the conversation they actually had. Rendering only the canned
+        # greeting made every return visit look like a first visit.
         return render_template("chat.html", me=ctx["agent"], persona=p,
-                               agent_id=agent_id)
+                               agent_id=agent_id,
+                               past=hc.recent_messages(ctx["tenant"], limit=30))
 
     # --- chat API (SSE), scoped to the account's agent -----------------------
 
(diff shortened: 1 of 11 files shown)


aks129:
## Phase 1 acceptance testing — one blocking defect found and fixed

Ran the three Phase 1 PRs (#238 + #239 + #240) merged onto a single integration branch, against a **live local HealthClaw** with a seeded synthetic tenant. Unit tests pass on each branch in isolation; this exercised them together across the HTTP boundary.

| # | Criterion | Result |
|---|---|---|
| AC1 | Terminology labels survive the real HTTP read path | ✅ |
| AC2 | Agent summarizer yields readable names, not `unlabeled record` | ✅ |
| AC3 | Chat turn persists to the tenant | ❌ **failed** → fixed |
| AC4 | Cold process rehydrates the conversation | ✅ |
| AC5 | `delete my records` removes conversations too | ✅ |
| AC6 | No PHI in the CareAgents DB or logs | ✅ |

### AC3 — the defect

**Every chat turn was rejected with HTTP 400 and silently dropped.** The feature would have shipped doing nothing.

HealthClaw's conversation endpoint validates `agent_id` against its own command-center agent registry. CareAgents agent ids are a **different namespace** and aren't in it. Because persistence is deliberately best-effort — losing a transcript must never break the conversation in front of the person — the failure was completely invisible.

```
POST /command-center/api/conversations  agent_id="ag_careagents_123"
  -> 400 {"error":"unknown agent_id: ag_careagents_123"}
```

**The unit tests could not have caught it.** They use a `FakeClient` that accepts any `agent_id`, so they proved the app *calls* `log_message`, not that `log_message` works.

Fixed in fda1a29: the identity travels in `metadata` (unvalidated) instead; a non-201 now logs a warning rather than returning a quiet `False`; two regression tests pin the wire format at the client layer where a fake can't paper over it.

### Verified after the fix

```
AC1  Condition   250.00   -> 'Type 2 diabetes mellitus'
     Observation 4548-4   -> 'Hemoglobin A1c'
     MedicationRequest 860975 -> 'Metformin 500 mg'
AC2  model input: 'Type 2 diabetes mellitus', 'Metformin 500 mg' — 0 unreadable
AC3  log user=True assistant=True
AC4  rehydrated oldest-first, full text preserved
AC5  purge: 200, 2 messages deleted, 0 remain
AC6  CareAgents tables: accounts, agents, connections, email_tokens, passkeys,
     surfaces — no transcript table; no chat text, clinical label, or token in logs
```

**1549 passed, 1 skipped** on the integration branch. Ruff (CI-pinned): clean.

One incidental confirmation: the seed initially 500'd because the local DB wasn't migrated, and the #182 fail-loud audit correctly refused to serve a read it couldn't record. Working as designed.

Issue #258, opened 2026-08-02
CareAgents in production is running stale code — none of the Phase 1/2 work is live, and prod-watch cannot see it


aks129:
## What I found

**Both** CareAgents deployments — the VPS serving `careagents.cloud` and the Railway service — are running code **older than `main`**, and every green signal we have says everything is fine.

Evidence (2026-08-02, unauthenticated):

| Check | Result |
|---|---|
| `scripts/prod_watch.py` | **9/9 passing** |
| `careagents.cloud/auth?enroll=1` | 200, but **0** occurrences of `enroll` in the HTML |
| `careagents-production.up.railway.app/auth?enroll=1` | 200, also **0** occurrences of `enroll` |
| local `careagents/templates/auth.html` | **5** occurrences of `enroll` |
| `careagents.cloud` DNS | `A 187.77.4.50` — still the **VPS**, cutover not done |
| both `/healthz` | `{"provider":"openai", ...}` |

The passkey-enrolment fix (#241) is the cleanest marker: it added the `?enroll=1` path, and neither live deployment has it. So both predate #241, which means **chat history (#222/#239), the passkey fix (#241), and the daily turn cap (#245) are all unshipped** — along with #251/#252/#256 merged since.

This is expected behaviour, not a regression: `docs/development.md` states CareAgents does **not** auto-deploy with the Flask app. Nobody has deployed it since the Phase 1 merges. The engine *is* current — prod-watch confirms the terminology labels are live (5/5 records labelled), because the Flask app *does* auto-deploy on push to `main`.

## Why this matters now

The August webinar demo is the CareAgents journey. Right now a stranger visiting `careagents.cloud` gets the pre-Phase-1 product: no durable chat history, the passkey dead-loop, no spend cap. The work is merged and the story is written, but the artifact is not reachable.

## The monitoring gap (the part worth fixing in code)

`prod_watch.py` passes 9/9 against a **stale deployment**. It checks liveness, readiness, grade, readability, and that sign-in accepts 8 digits — all of which a months-old build also satisfies. Its documented scope limit is "does not cover the signed-in journey"; **version staleness is a second, undocumented blind spot**, and the more dangerous one, because it converts "we shipped it" into a false positive.

[...]

Issue #233, opened 2026-08-01
ci: no CareAgents browser test — all e2e targets a different product


aks129:
All three Playwright specs (`e2e/tests/landing.spec.ts`, `dashboard.spec.ts`, `api.spec.ts`) target the **HealthClaw site and r6 dashboard**. The only CareAgents assertion anywhere in e2e is that the HealthClaw landing page links to careagents.cloud.

Signup, email code, passkey, consent, connect, chat SSE, and review cards have **zero browser coverage**. `tests/test_careagents.py` covers the server well, but the defects found in the 2026-08-01 UX audit — dead passkey link (#223), `prompt()` flows that die in mobile browsers (#224), popup-blocked connect windows — are exactly the class only a browser test catches.

**Minimum viable spec:** email-code login with faked mail → sample connect → agent create → one chat turn.

Also worth fixing the recurring brittleness: #174 and #187 were both stale-locator failures caused by coupling selectors to marketing copy.

From the 2026-08-01 dev-system + UX audits.

Issue #224, opened 2026-08-01
ux: native prompt()/alert() in core flows — unstylable and silently dead in some mobile browsers


aks129:
Non-technical patients are asked to **type a number into a browser prompt** to pick a wearable provider (`careagents/static/home.js:25-29`); Telegram/iMessage codes arrive via `prompt(...)` (`:220, :233`); errors via `alert()`.

These look broken, break the design language, and are suppressed in some in-app/mobile browsers — silently dead-ending the flow on the primary device for this audience.

Also in scope: `.conn-refresh`, `.conn-disconnect`, `.conn-delete`, `.conn-refresh-msg` have **no CSS rules at all** — they render as raw browser-default buttons inside the polished cards, and destructive Delete carries identical visual weight to Refresh.

(The typed-DELETE confirmation is deliberate friction and should stay — but styled.)

From the 2026-08-01 UX audit.

PR #259, opened 2026-08-02
docs: add a ship gate, and stop handing out the unpinned-ruff trap

aks129:
## Why

On Aug 2 I verified production and found that **every CareAgents change from Phase 0 and Phase 1 was undeployed** — durable chat history (#222/#239), the passkey fix (#241), the daily turn cap (#245) — while `scripts/prod_watch.py` reported **9/9 green**.

The monitor was not wrong; it was answering a different question. Liveness, readiness, guardrail grade, and record readability are all satisfied by a months-old build. Details and evidence in #258.

The root cause in *this* document is simpler: the plan had no deploy step. "Done" meant merged, so merged is where the work stopped.

## What changes

- Phase 2 gains a **ship gate**: each phase ends with a deploy, not a merge.
- Phase 2's exit criterion is scoped to the build a stranger would actually reach, rather than to `main`.
- Notes that Phase 3's dry run is meaningless before that deploy lands, and that the passkey cannot be exercised at all until DNS points at the deployed service (WebAuthn is bound to `CARE_RP_ID=careagents.cloud`).
- Checks off the two Phase 2 items that did land (#241, #244) and cross-references the two in flight (#224, #233).

Docs only — no code, no behaviour change. `uv run ruff check .` passes.

## Not included

This PR does not deploy anything. Deploying CareAgents needs explicit authorization and the DNS cutover is owner-owned; both are listed as decisions in #258.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,7 +5,7 @@
 ## Checklist
 
 - [ ] Tests added/updated and `uv run python -m pytest tests/ -q` passes
-- [ ] `pipx run ruff check .` passes
+- [ ] `uv run ruff check .` passes (not `uvx`/`pipx` — those fetch an unpinned ruff)
 - [ ] No PHI in code, tests, fixtures, or logs (synthetic data only)
 - [ ] Touches auth/audit/redaction/tenancy? Note it here so maintainers review against the compliance rules
 - [ ] Node changes: `npx tsc --noEmit && npm test` passes in `services/agent-orchestrator`
--- a/docs/2026-08-01-webinar-delivery-plan.md
+++ b/docs/2026-08-01-webinar-delivery-plan.md
@@ -94,13 +94,35 @@ tab → come back → the conversation is still there.
 
 The bugs that make a live demo or an unsupervised advocate hit a wall.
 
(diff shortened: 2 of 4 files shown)


aks129:
Added a second, unrelated-but-adjacent docs fix to this branch (`9183a3f`), since it is the same failure mode: **a document quietly disagreeing with how the project actually works.**

Four files told contributors to lint with `uvx ruff check .` or `pipx run ruff check .`. CI runs `uv run ruff check .`, which uses the version pinned in the dev group (`ruff>=0.15,<1`); `uvx`/`pipx` fetch the newest published release instead. A newer ruff reports hundreds of findings CI does not — none of them caused by the contributor's change, and all of them a waste to chase. The repo already documents this as a trap that costs an hour, and then handed out the trap in four places.

Fixed here:
- `.github/PULL_REQUEST_TEMPLATE.md:8` — **the worst one**: every pull request in this repo has been asking its author to run the wrong command.
- `docs/agent-task-guide.md:114,136` — the second worst, since `CLAUDE.md` names this file as the first read for real work.
- `docs/extending-the-action-rail.md:115`.

`RELEASING.md:16` and `docs/development.md:17` have the identical defect and are fixed on the `#258` branch, which already edits both files — kept there to avoid a conflict.

Found by Dev while implementing the build marker; it flagged the instances outside its own diff rather than silently widening its PR, which is why they got fixed at all.

PR #267, opened 2026-08-02
feat(careagents,r6): direct FHIR-bundle upload path (#227)

### LINKED ISSUE #227: integrations: ship the file-upload / SHL import tile for real
The `direct` and `shl` connector tiles advertise paste/upload but `start()` returns `{"soon": True}` (`careagents/connectors.py`).

This is the **zero-integration ingest path**: every patient can export a file, and nobody needs an API deal or a vendor relationship for it. It is also the only path that works for patients whose provider Fasten does not cover.

Implementation is small — accept a FHIR bundle / SHC file and reuse the existing `_ingest_one` loop from `/shc/ingest`.

Doing this also makes the currently-false 'import' tier honest (see #225).

From the 2026-08-01 integrations audit.

aks129:
## Summary

Ships the zero-integration ingest tile: a signed-in patient uploads a
FHIR Bundle exported from another app or portal, and the engine ingests
it through the same code path Fasten/SHC take. Turns the previously
placeholder `direct` connector tile into a real end-to-end flow.

## Files changed

- `r6/routes.py` — new `POST /r6/fhir/internal/ingest-bundle` (header-only tenant, streaming cap, per-entry savepoints, PHI-safe errors, direct-upload provenance).
- `r6/fasten/ingester.py` — `_ingest_one` gains optional `agent_id` / `detail` kwargs (defaults preserve Fasten/SHC behavior).
- `careagents/healthclaw.py` — `HealthClawClient.ingest_bundle(tenant, bundle)`; `HealthClawError` carries stable engine `code` + `correlation_id`.
- `careagents/connectors.py` — `direct` tile promoted `soon` → `import` with a real `start()`; `catalog()` marks it `requires_consent=True`.
- `careagents/app.py` — new `POST /api/connections/<conn_id>/upload` (ownership, kind gate, streaming cap, FHIR MIME set, `mark_synced` only on `ingested > 0`, strips internal `tenant_id` from browser response, preserves engine error codes + correlation_id).
- `careagents/templates/home.html` — Upload button on `direct` cards; hidden file chooser; `role=status aria-live=polite` on the result surface.
- `careagents/static/home.js` — file → POST → results loop with code-owned patient copy for every 4xx/5xx path (Honey's copy review); `errors[].correlation_id` surfaced on partial 200s.
- `careagents/static/careagents.css` — `form-ok` / `form-warn` result-surface styles.
- `tests/test_ingest_bundle_endpoint.py` — 26 tests covering every engine release condition.
- `tests/test_careagents.py` — 17 new tests including a full cross-layer integration test (real CareAgents Flask → real engine WSGI, no fakes).

## Release contract met (see thread for the full spec)

**Engine:**
- Header-only tenant selector (body `tenant_id` → 400 `legacy_body_selector`).
- Streaming body read via `request.stream.read(max_bytes + 1)` — kills chunked / no-`Content-Length` bypass.
- MIME accept-list is `application/json` only for the envelope (`application/fhir+json` is refused on this internal endpoint because the body is a `{bundle: ...}` envelope, not a raw FHIR resource; the patient-facing CareAgents route accepts `fhir+json` for the raw Bundle from the browser and wraps into the envelope).
- Per-entry `db.session.begin_nested()` savepoints — a mid-Bundle DB exception rolls back only that row; `ingested` matches persisted rows.
- No exception text ever returned; correlation_id logged server-side and returned as an opaque handle in the response and per-entry `errors[]`. `logger.warning` records only exception class name, never the message or a traceback.
- Audit outcome is `partial` whenever `failed OR skipped` is non-zero.
- Direct-upload path records `agent_id=direct-upload` in the per-resource `AuditEventRecord` — never borrows Fasten's identity.
- `entry.resource` must be a JSON object; flat entry acceptance removed.

**CareAgents:**
- Tenant identity derived only from the authenticated account + owned connection.
- Streaming cap on the app layer too (same defense).
- Widened MIME accept-list for the patient-facing raw Bundle.
- `mark_synced` and the `empty → active` connection flip run ONLY when `ingested > 0` — an all-failed / all-skipped bundle never fakes sync freshness.
- Response strips the engine's `tenant_id` before the browser sees it.
- Engine error codes flow through `HealthClawError.code`; UI renders code-owned patient copy for every 4xx/5xx path (`payload_too_large`, `content_type_required`, `invalid_json`, `not_a_bundle`, `invalid_bundle`, `too_many_entries`, `wrong_connector_kind`, `ingest_error`, `commit_failed`).
- Support correlation_id surfaced on failure; raw backend messages never rendered.
- `role=status aria-live=polite` on the result surface for screen readers.
- Honey's UX-copy review: blocker `invalid_bundle` copy replaced; two optional refinements (`content_type_required` and skip summary) both applied.

## Caps

- 5 MiB body / 500 entries by default. Env overridable (`INGEST_BUNDLE_MAX_BYTES` / `INGEST_BUNDLE_MAX_ENTRIES`). Matches the strongest production precedent (Azure FHIR: 500 entries, 28 MB) per `RESEARCH/HEALTHCLAW_207_227_SAFETY_LIMITS_2026_08_02.md`.

## Test plan

- [x] Full Python suite at exact head `090f0293`: **1623 passed, 2 skipped**.
- [x] Full Node suite (services/agent-orchestrator jest): **176 passed**.
- [x] `tests/test_ingest_bundle_endpoint.py` — **26 passed** (all engine release conditions).
- [x] `tests/test_careagents.py` — **93 passed** including 17 new direct-upload tests.
- [x] Cross-layer integration test writes a Patient into a real engine tenant via the real CareAgents endpoint and verifies it is readable back via the engine's search API.
- [ ] Awaiting GitHub Actions gates (Python, Postgres, Node, Playwright, Compose, lint, CodeQL, dependency, secret).
- [ ] Awaiting independent review by @bagupsh at exact head `090f0293`.
- [ ] Awaiting @crista blessing at the same exact head.

## Follow-up (out of scope)

- **SHL wrapped-manifest import** — same endpoint, different kind. The `shl` tile stays `soon` for this PR; the encrypted-manifest decoder is the separate slice.

Closes #227.

### DIFF
--- a/careagents/app.py
+++ b/careagents/app.py
@@ -41,6 +41,27 @@ logger = logging.getLogger(__name__)
 # strongest privacy control in the one place people read carefully.
 CONSENT_VERSION = "2026-08-01"
 
+# Local hard cap for the file-upload path (#227). The engine's
+# `internal/ingest-bundle` is the source of truth for the value; this
+# ceiling stops an oversized (or chunked / no-Content-Length) request
+# from ever spending more than max_bytes+1 in our process before we
+# refuse it. 5 MiB matches the engine default so a request either fails
+# here quickly or lands cleanly.
+_UPLOAD_MAX_BYTES = 5 * 1024 * 1024
+
+# FHIR R4 §3.2 SHALL support `application/fhir+json`; the two legacy
+# variants ride the same accept-list so a real FHIR client, a plain-JSON
+# hand-crafted body, and older exporters all land.
+_UPLOAD_MIME_TYPES = frozenset({
+    "application/fhir+json",
+    "application/json",
+    "application/json+fhir",
+})
+
+# In-memory conversation bounds. Each worker caches the chats it serves, so
+# these cap memory and keep a long-running process from degrading (#218).
+MAX_LIVE_CONVERSATIONS = 200
+CONVERSATION_IDLE_SECONDS = 6 * 3600
 def create_app(config: Config | None = None,
                client: HealthClawClient | None = None,
                accounts: AccountService | None = None) -> Flask:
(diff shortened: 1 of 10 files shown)


aks129:
Rebuilt against current `main` (`6b02f76`). Two conflicts, both **semantic rather than positional** — the kind that would have merged cleanly on a re-run and still been wrong:

- **`careagents/app.py`** — this branch added the #227 upload constants immediately next to `MAX_LIVE_CONVERSATIONS` / `CONVERSATION_IDLE_SECONDS`, and `main` **deleted** those when durable runs replaced the in-process conversation cache (#257). Taking this branch's side wholesale would have resurrected dead constants; taking main's would have dropped the upload limits. Kept the upload constants, dropped the bounds — verified nothing in `careagents/` still references them.
- **`careagents/templates/home.html`** — both sides appended distinct markup at the same point: this branch's hidden file input for the upload path, and #224's provider picker / code card / delete confirmation. Both kept.

Also fixed two lint errors this branch was carrying into a CI-gated check: `F401` for unused `pytest` and `r6.routes` imports in `tests/test_ingest_bundle_endpoint.py`. CI runs `uv run ruff check .` and would have failed the PR on them.

```
1762 passed, 6 skipped, 2 xfailed
ruff: All checks passed!
```

Now `MERGEABLE`. One note for whoever reviews: I have **not** reviewed the #227 upload feature itself — this was a conflict rebuild plus a lint fix, not a code review of the direct-upload path. It touches a write path that accepts a caller-supplied FHIR bundle, so it deserves a real review against the guardrail invariants (tenant scoping, audit, redaction on the way back out) before merge.

Issue #287, opened 2026-08-03
Don't delete any single id-validation test — they only work as an ensemble


aks129:
Recorded by QA during #267 verification, so it isn't lost to a future tidy-up.

There are now four tests constraining resource-id validation, and QA's mutation testing showed **they are only sufficient together**. Each catches something the others don't, and the newest one has a self-reference that makes it individually bypassable:

| Test | File | Catches |
|---|---|---|
| `test_resource_id_pattern_is_a_charset_control_not_a_length_control` | `test_ingest_resilience.py` | pattern ceiling ≠ column width (both directions), charset widening — **sole catcher** for ceiling→256 and charset+`_` |
| `test_fasten_ingest_still_accepts_real_epic_shaped_long_ids` | `test_ingest_resilience.py` | hardcodes an 86-char id — immune to coupled mutations |
| `test_resource_id_column_fits_real_ehr_ids` | `test_ingest_resilience.py` | absolute floor: `R6Resource.id >= 128` |
| `test_audit_resource_id_column_fits_real_ehr_ids` | `test_ingest_resilience.py` | `AuditEventRecord.resource_id >= R6Resource.id` |

**The trap:** the first test derives its expected width *from* `R6Resource.id`. So a mutation that narrows the pattern **and** the column together (255→64 on both) passes it. QA verified that case is caught — but only by the two absolute-valued tests (the hardcoded 86-char id, and the `>= 128` floor). Likewise, narrowing the *audit* column alone — the column the 2026-07-08 live incident actually rolled back on — is caught only by the audit-width test.

These four overlap enough to look redundant. They are not. Deleting any one of them opens a coupled mutation that ships silently, and the failure mode is the dangerous direction: real Epic resources skipped on the live connector rather than a crash.

The chain, all links independently mutation-verified: `pattern ceiling == R6Resource.id` ∧ `audit >= R6Resource.id` ∧ `R6Resource.id >= 128` ∧ *an 86-char id actually survives `_ingest_one`*.

Context: #267, and the 2026-07-08 incident documented in `tests/test_ingest_resilience.py`'s module docstring.

aks129:
Closed in a lean triage (23 Sep): process/meta work that adds machinery without changing what a patient gets. The rule it states is already how PRs are run (rebase on main before merge, mutation evidence in the PR body); specific weak guards are fixed where they are named (#630, #634, #638).

Issue #286, opened 2026-08-03
ingest: a blank resource id is now refused rather than auto-assigned — decide which is correct


aks129:
Found by QA verifying the #267 blocker fixes. **LOW, not blocking**, and deliberately left unpinned by a test because the right answer is a product call.

`_ingest_one` (`r6/fasten/ingester.py`) previously did `resource.get('id') or str(uuid.uuid4())`. The #267 id-validation rewrite made that an explicit `is not None` branch, which changed the falsy cases:

| `id` value | before | after |
|---|---|---|
| `""` | uuid generated, **ingested** | **`invalid_id` — refused** |
| `0` | uuid generated, ingested | ingested as `"0"` |
| `false` | uuid generated, ingested | ingested as `"False"` |
| `12345` (int) | stored as raw int | ingested as `"12345"` |

Repro: `_ingest_one({"resourceType":"Observation","status":"final","id":""}, "test-tenant")` returns `('invalid_id', None)`; on `main` it returned `('ok', <uuid>)`.

This is the same root cause as the id-length regression QA caught earlier — the id check is not gated by `allowed_types`, so any semantic change in it leaks to the Fasten and SHC paths, not just direct upload. Much narrower here: a FHIR resource with a blank id is malformed, so unlike the length bug (65/250 real Epic ids affected) this should be rare.

**Why no test was added:** the old behaviour is arguably the worse one. Fabricating a UUID for a blank id means a re-ingest of the same resource creates a *duplicate row* instead of upserting on `(tenant, type, id)` — so "refuse it" may well be the correct semantics. Cementing either behaviour in a test before someone decides is premature.

**Decide:** refuse a blank id (current), or restore uuid-generation for it? If restoring, note the duplicate-on-reingest consequence. The `false` → `"False"` coercion is a curiosity rather than a realistic input either way.

Related: #267, #281.

Issue #282, opened 2026-08-02
Redaction coverage inventory: several read paths never call apply_redaction


aks129:
Found in the CTO's review of #267, listed as pre-existing and out of scope for that PR — but worth its own tracking since #267 widens who can put raw text in front of these readers.

Paths that read FHIR resources without redacting: \`r6/labs/routes.py:51\`, \`r6/caregaps/routes.py:55\`, \`r6/quality/routes.py:45\`, \`r6/actions/routes.py:234\`, \`r6/actions/rails/form_fill.py:153\`, \`r6/sdc/documents.py:72\`, \`r6/smbp/routes.py:118\`, \`r6/curatr.py:1097\`.

Worth an inventory pass: which of these are genuinely internal/pre-redacted-input only, and which are reachable with caller-influenced content and need \`apply_redaction\` added.

aks129:
**Inventory, 14 Sep 2026, on main after #725/#726/#728.** Each site classified by what leaves it, not by whether `apply_redaction` appears in the file.

| Site | What it reads | What leaves | Class | Action |
|---|---|---|---|---|
| `r6/labs/routes.py` | Observations, DiagnosticReports | trend lines and rows | already redacted (`apply_redaction(copy.deepcopy(obs))` per row, since the #376-era fix) | none |
| `r6/caregaps/routes.py` | Patient, Conditions, Observations | `summary`, `consumerSummary`, `detail` built by `evaluate_care_gaps` | **derived only**: the evaluator reads codes, dates, status and value; results carry rule names, status labels and our own notes, never a coding display or `text` | none; add a marker probe if the evaluator ever starts copying evidence text |
| `r6/quality/routes.py` | Patients, Conditions, Observations | NQF0018 MeasureReport | **derived only**: counts, rates, patient ids; the report shape carries no resource content | none |
| `r6/smbp/routes.py` | Observations for the session's patient | HTML/PDF report from `_components(obs)` and `effectiveDateTime` | **derived only**: systolic/diastolic numbers and dates; label is the patient id, not a name | none |
| `r6/sdc/documents.py` | DocumentReference it just wrote / its attachment bytes | `persist_intake_document` return reaches nobody (form_fill reads `['id']`); `get_document_pdf_bytes` returns the PDF the rail rendered | **own content**: both are our own artefacts, and the PDF's inputs were fixed by #367/#724 (structured name) and human-confirmed rows | none |
[...]

Issue #281, opened 2026-08-02
R6Resource.id is String(255) with no length check — SQLite/Postgres diverge silently


aks129:
Found in the CTO's review of #267, but pre-existing and broader than that PR.

\`r6/models.py:59\` — \`R6Resource.id\` is \`String(255)\`. Nothing validates length before insert. SQLite accepts an overlong value silently; Postgres raises \`StringDataRightTruncation\`. CI's default lane is SQLite, so this class of bug ships invisibly and only surfaces on the Postgres lane's hardcoded test allowlist (\`docs/development.md\`'s known gap) or in production.

Not urgent, not #267-specific — filing so it's tracked rather than lost in that PR's review. A repo-wide id-length validation pass (or a shared FHIR-id validator used at every write path) would close this generally.

aks129:
Every path an id enters `R6Resource.id` or `AuditEventRecord.resource_id` is now bounded before the column: ingest (`_RESOURCE_ID_PATTERN`, ≤255, #288), a client-supplied id on create (`_FHIR_ID_PATTERN`, ≤64), `$ingest-context` ids, and since #726 every `/<resource_type>/<resource_id>` rule (≤255, 400 without echo). The SQLite/Postgres divergence on an overlong value is no longer reachable through the API, and the Postgres CI lane runs the route tests. Closing; reopen if a new entry point appears without a pattern in front of it.

PR #288, opened 2026-08-03
fix(guardrails): close two live vulnerabilities in the direct-upload path (#267 review)

aks129:
Fixes the CTO review findings on #267 / `agent/direct-upload-227`. **This branch does not merge #267's feature — it hardens it.** Whether the feature itself ships before Aug 18 is a separate call; see the recommendation at the bottom.

## Why

#267 added a direct FHIR-bundle upload path, written overnight, never security-reviewed. The CTO reviewed it and returned **BLOCK** with two live vulnerabilities — found by probing a running server, not by reading — plus three more findings.

**B1 — unauthenticated arbitrary write into the demo tenant.** The endpoint gated on `_internal_mint_authorized`, which exempts public tenants (`PUBLIC_TENANTS=desktop-demo` in production). That exemption is right for *minting a token* — a public tenant already bypasses read-auth, so there's nothing extra to protect — and wrong for *authoring content*. An unauthenticated caller could write arbitrary FHIR resources into the exact tenant used for the live demo, content that then flows into an LLM context: stored prompt injection on top of defacement. The ingester also never consulted `_SYSTEM_MANAGED_TYPES`, so `AuditEvent` / `Permission` / `Consent` / `Provenance` / `Bundle` could all be forged.

**B2 — unauthenticated remote crash. 120KB, zero credentials, any tenant.** The body was read and parsed *before* the auth check. CPython's JSON scanner recurses per nesting level, so a 60,000-deep payload raised `RecursionError` — not a `ValueError`, so uncaught → 500. The carefully-built 5MB cap was irrelevant; the crash lever sits three orders of magnitude beneath it.

## What changed

- **`_internal_ingest_authorized`** — requires `X-Internal-Secret` unconditionally, no public-tenant exemption, fail-closed in production. CareAgents already sends the secret on every call, so the legitimate path is untouched.
- **Auth gate moved first** — before content-type sniffing, before the body is read, before anything is parsed. `RecursionError` is now caught explicitly, plus a bounded depth-walk as defence in depth. Same crash fixed on the CareAgents forwarding path.
- **Ingest type allowlist** — `_ingest_one` gained `allowed_types`, defaulting to `None` so Fasten/SHC keep prior behaviour exactly; the upload path passes a clinical-only set.
- **Resource-id validation** — a caller-supplied id was reaching `AuditEventRecord.resource_id`, which `health_compliance.py` exports into the auditor-facing compliance bundle. Now charset-validated before use.
- **Revoked-connection upload bypass** — enforced only in the template; a crafted request still uploaded after disconnect. Now refused server-side.
- **Transport-exception wrapping** on `hc.ingest_bundle`, matching the idiom every other engine call in that file already used.

## The regression I introduced, and QA caught

Worth reading, because it is the most instructive part of this branch.

My id validation used the FHIR spec pattern verbatim — charset **and** its 64-char ceiling — without asking which half was the security control. Only the charset was. But `tests/test_ingest_resilience.py` documents a live 2026-07-08 Epic export where **65 of 250 resource ids exceeded 64 characters**, running to ~109 — which is precisely why `R6Resource.id` was widened rather than truncated. My regex re-imposed that ceiling inside `_ingest_one`, which Fasten and SHC both call and which #267 was never meant to touch.

The failure mode was the dangerous one: **silent data loss on real patient records**, as `invalid_id` skips on the live connector rather than a crash. 1774 green tests missed it, because the existing regression guard for that exact incident writes the ORM row directly and never calls `_ingest_one` — the same shape as the B1 hole, where the test that should have covered it ran past the code that mattered.

Ceiling is now 255, matching both columns so validation and storage cannot disagree.

## Verification

QA verified independently across two rounds — live HTTP probing plus direct SQLite reads, not just the shipped tests:

- **B1**: no secret / wrong secret / empty-string secret → all `403`; correct secret → `200`. Empty-vs-unset `INTERNAL_TOKEN_MINT_SECRET` confirmed consistent between both auth helpers.
- **B2**: deep arrays *and* deep objects, with and without credentials → `400`/`403` in ~0.01s, body never read; server alive after all four.
- **17 id cases** through the live endpoint: 12 refused (including newline injection, CSV formula, null byte, Cyrillic homoglyph, and the 256-char boundary), 5 accepted (including the 109-char real-incident length and the exact 255 column width). Direct DB query confirmed refused ids reached neither `r6_resources` nor `audit_events`.
- **`allowed_types` verified across all 40 supported types**: omitted → refuses nothing (Fasten/SHC genuinely unaffected); upload set → refuses exactly the 5 declared, no gap.
- **Revoked-connection logic traced exhaustively** against every status the codebase actually persists — `direct` connections can only be `empty → active → revoked`, so checking `== 'revoked'` (not `!= 'active'`) is correct against every real state. Requiring `'active'` would have refused every legitimate first upload; I caught that against my own first draft.
- **11 mutations** run against the id tests — all caught, none survived.

```
1774 passed, 6 skipped, 2 xfailed
conformance: 28/28, Grade A
ruff: All checks passed!
```

## Known limits

- No production-mode boot; the fail-closed-in-production branch is verified by code inspection (identical logic to the already-tested `_internal_mint_authorized` equivalent).
- **SQLite only.** The original 2026-07-08 incident was Postgres-only, so a Postgres lane would be the honest final check on column widths.
- No real Fasten/SHC cassette data exists in the repo, so "real Epic id shapes" rests on this repo's own incident record rather than fresh samples.
- Deferred, filed, and explicitly **not** fixed here: #279 (audit `resource_id` exposure in compliance export), #280 (revoked-connection — fixed here, issue tracks the broader pattern), #281 (general id-length validation), #282 (non-redacting read paths), #286 (blank-id semantics), #287 (the id-test ensemble must not be pruned).
- The 50–100MB parse / memory-exhaustion finding from the CTO review is **not** addressed by the depth-walk and remains open. Nothing here should be read as closing it.

## Recommendation — yours to decide

The CTO recommends **deferring #267 itself past Aug 18**, independent of these fixes. The reasoning: fixing the blockers took about an hour, but convincing yourself a caller-supplied-bundle ingest path is genuinely safe took a full review that found five things — and then the fix introduced a HIGH regression caught only by adversarial review. That ratio is the signal. The feature is also not on the webinar's critical path.

These patches are sound and can merge on their own terms. Shipping the *feature* before the webinar is the separate call.

### DIFF
--- a/careagents/app.py
+++ b/careagents/app.py
@@ -335,6 +335,19 @@ def create_app(config: Config | None = None,
                             "message": "This connection does not accept file "
                                        "uploads. Create an 'Upload records' "
                                        "connection to import a FHIR bundle."}), 400
+        # Revocation was enforced only in the template (home.html hides the
+        # upload button once status == "revoked"), so a revoked connection
+        # still accepted uploads on a direct/crafted request — defeating the
+        # disconnect guarantee ("stop new data flowing", accounts.py). A
+        # `direct` connection's normal lifecycle is "empty" (created, no
+        # file yet) -> "active" (after the first successful upload), so the
+        # check must exclude "revoked" specifically rather than require
+        # "active" — requiring "active" would refuse every first upload.
+        if conn["status"] == "revoked":
+            return jsonify({"error": "connection_not_active",
+                            "status": conn["status"],
+                            "message": "This connection has been disconnected "
+                                       "and no longer accepts uploads."}), 409
 
         # Header short-circuit for callers that DO set Content-Length, but
         # never trusted alone — the stream read below is the real bound.
@@ -364,6 +377,17 @@ def create_app(config: Config | None = None,
 
         try:
             bundle = json.loads(raw.decode("utf-8")) if raw else {}
+        except RecursionError:
+            # Same crash lever as the engine's ingest-bundle endpoint
+            # (#267 review): CPython's JSON scanner recurses per nesting
+            # level, and a payload well under _UPLOAD_MAX_BYTES can nest
+            # deep enough to raise RecursionError, which is not a
+            # ValueError and was previously unhandled -> 500. This route
+            # already sits behind @login_required, so the engine-side fix
+            # (moving auth before parse) doesn't apply here; this is purely
+            # the crash fix.
+            return jsonify({"error": "invalid_json",
+                            "message": "nesting too deep to parse"}), 400
         except (ValueError, UnicodeDecodeError):
             return jsonify({"error": "invalid_json"}), 400
         if not isinstance(bundle, dict):
--- a/careagents/healthclaw.py
+++ b/careagents/healthclaw.py
@@ -102,13 +102,22 @@ class HealthClawClient:
         # envelope call, NOT a raw FHIR Bundle post. `application/fhir+json`
         # would mis-label the envelope; the patient-facing route above
         # accepts fhir+json for the raw Bundle from the browser.
-        r = self.http.post(
-            f"{self.fhir}/internal/ingest-bundle",
-            json={"bundle": bundle},
-            headers={"X-Tenant-Id": tenant,
-                     "X-Internal-Secret": self.mint_secret,
-                     "Content-Type": "application/json"},
-            timeout=self.timeout)
+        try:
+            r = self.http.post(
+                f"{self.fhir}/internal/ingest-bundle",
+                json={"bundle": bundle},
+                headers={"X-Tenant-Id": tenant,
+                        "X-Internal-Secret": self.mint_secret,
+                        "Content-Type": "application/json"},
+                timeout=self.timeout)
+        except requests.RequestException as exc:
+            # A transport-level failure (connection refused, timeout) was
+            # previously unwrapped here, propagating out of
+            # upload_connection as an unhandled Flask 500 and breaking the
+            # JSON error contract the UI depends on (#267 review). Every
+            # other engine call in this file wraps requests.RequestException
+            # into HealthClawError; this one didn't.
+            raise HealthClawError("ingest failed", 0) from exc
         try:
             body = r.json()
         except ValueError:
--- a/r6/fasten/ingester.py
+++ b/r6/fasten/ingester.py
@@ -31,14 +31,25 @@ logger = logging.getLogger(__name__)
 _PROGRESS_BATCH = 10  # commit progress every N resources (observability)
 _DOWNLOAD_TIMEOUT = httpx.Timeout(connect=10.0, read=300.0, write=60.0, pool=10.0)
 
-# FHIR id: [A-Za-z0-9\-\.]{1,64} per the spec. A caller-supplied id outside
-# this shape was previously stored verbatim and copied into
+# The CHARSET is the security control here, not the length. A caller-supplied
+# id outside this shape was previously stored verbatim and copied into
 # AuditEventRecord.resource_id (#267 review) — audit `detail` stays PHI-free
-# by convention, but that convention did nothing for the adjacent column,
-# and health_compliance.py exports resource_id into the auditor-facing
-# compliance bundle. Validated before use, not filtered — an id this fails
-# is refused per-entry rather than silently truncated or replaced.
-_RESOURCE_ID_PATTERN = re.compile(r'^[A-Za-z0-9\-\.]{1,64}$')
+# by convention, but that convention did nothing for the adjacent column, and
+# health_compliance.py exports resource_id into the auditor-facing compliance
+# bundle. An id restricted to [A-Za-z0-9-.] cannot carry a name, a date, a
+# path, or markup, which is what that finding was about.
+#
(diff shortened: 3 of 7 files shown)


aks129:
## Rebased, conflicts resolved — and this is now urgent

**Conflict cause:** the branch carried its own copy of the #227/#267 feature commits, and that feature merged to `main` separately as `0d5246b`. Rebased with `git rebase --onto origin/main 6b02f76`, dropping the two duplicates and replaying only the three hardening commits. `MERGEABLE` now, and the diff is exactly 7 files of hardening — the `careagents/static/*`, `home.js` and `home.html` noise is gone.

## The feature merged without its hardening, so both vulnerabilities are live on production right now

Verified against `app.healthclaw.io` with no credential and an unparseable body:

```
$ curl -X POST .../r6/fhir/internal/ingest-bundle \
    -H "X-Tenant-Id: desktop-demo" --data-binary "not-json{"
400 {"error":"invalid_json"}
```

`400`, not `401` — an unauthenticated caller reached the JSON parser. That is B2, live. And because the auth gate never runs for a public tenant, B1 is live too.

Reproduced through the WSGI app as an acceptance test, run identically against `main` and this branch:

| | `main` | this branch |
| --- | --- | --- |
| auth runs before the body is parsed | ❌ 400 invalid_json | ✅ refused |
| deep nesting refused without credentials | ❌ parsed | ✅ refused |
| public-tenant write requires the internal secret | ❌ accepted | ✅ refused |
| refused write leaves no record behind | ❌ **persisted a resource** | ✅ nothing written |
| the authorized path still works | ✅ | ✅ |

**4 failed, 1 passed** on main → **5 passed** here. The fourth row is the one that matters: on `main` an unauthenticated request persisted an `Observation` into the `desktop-demo` tenant. The fifth row is the guard against over-correcting — the legitimate caller still gets a 200.

## Gates

`1919 passed, 6 skipped, 2 xfailed`; `uv run ruff check .` clean; conformance Grade A.

## Note on CI

`playwright-tests` is red here for an unrelated reason — the e2e boot gate has been broken on `main` since Aug 2 and no spec has executed since. Fixed in #292, which should merge first so this PR CI means something.

## Standing recommendation unchanged

The CTO position on shipping #267 *the feature* before Aug 18 has not changed. These patches close what is already exposed; whether the feature stays enabled for the webinar is the separate call.

aks129:
## QA adversarial review: SAFE TO MERGE — with one finding I fixed first

QA verified the rebase by content rather than by trust: the *introduced* diffs (`6b02f76..ad23aee` vs `origin/main..278cc84`) are byte-identical apart from blob hashes and hunk offsets. Zero hardening lost. It also cleared all three blueprint `before_request` hooks to confirm none of them touches the body ahead of the auth gate, and confirmed 255 matches both `R6Resource.id` and `AuditEventRecord.resource_id` plus the 64→255 migration.

It also re-derived the acceptance result independently and got a **stronger** number than mine — **5** security tests fail on `main`, not 4. On `main`, `../../etc/passwd` and a 400-character id both ingested successfully.

### The finding worth stopping for

Mutation testing showed **the fix central property was pinned by nothing.** Moving `_internal_ingest_authorized` back below `json.loads` — literally reintroducing B2 — left the suite green at 1919 passed.

The cause was my own assertion:

```python
assert r.status_code in (400, 403)
```

The `RecursionError` catch also answers 400, so that passed whether auth ran before the parse or after it. An assertion that accepts both the fix and the bug. Same shape as the `-ge 15` guard and the alarm predicate — a control that looked like one thing and quietly did two.

Fixed in `654b0b1`:

- Tightened to `== 403`.
- Added `test_auth_is_the_first_gate_whatever_else_is_wrong_with_the_request` — five requests each failing a *different* downstream check (unparseable body, wrong content type, absent content type, legacy body selector, non-object body), all unauthenticated, all required to return 403. It states the ordering as a property instead of inferring it from one input. **The same mutation now fails 5 tests instead of 0.**
- `_INGEST_MAX_JSON_DEPTH` had zero coverage for the same reason — the 60000-deep payload trips `RecursionError` before reaching `_json_depth_within`, so the constant could be raised to ten million unnoticed. Added a case CPython can parse but the guard must reject, plus a just-inside-the-limit bundle that must still be accepted. Mutation-verified.

### Production config confirmed

QA flagged that B1 is only closed on prod if that deployment sets `INTERNAL_TOKEN_MINT_SECRET` or `APP_ENV=production` — not knowable from the repo. Checked the Railway service directly: **`INTERNAL_TOKEN_MINT_SECRET` is set** (names only, no values read). The fail-closed branch genuinely engages; this is not relying on the `resolve_app_env()` fallback.

### Gates

`1925 passed, 6 skipped, 2 xfailed`; ruff clean; conformance 28/28 Grade A.

### Deliberately not fixed here

- **#293** — invalid-id and forbidden-type outcomes collapse to `skipped` with no log on the Fasten/SHC paths. Same silence that would have hidden the ceiling regression.
- **#294** — the transport-failure wrap in `careagents/healthclaw.py` has no regression pin.

Neither blocks this PR.

PR #316, opened 2026-08-03
docs: working protocol for the architecture refactor — pins before moves, no break-fix

aks129:
Records the founder decision to proceed with the refactor now, and the protocol that makes "careful and intentional, not rushed break-fix" operational rather than aspirational.

## The decision

The CTO/Product recommendation to defer all structural work past Aug 18 is **overruled on priority**. Their *design* objections stand and are being resolved before any implementation — the CTO BLOCKED two parts of the original kernel design (`Profile.RAW_INTERNAL` as an enum member meaning "no property", and an errorhandler that could swallow an unexpected `StepUpDenied` into a clean 401). A corrected spec lands before code.

## The protocol (the part that matters)

- **Pins before moves.** #313 (the write-guard matrix) must merge **first** — it is the instrument the migration is measured against.
- **The kernel lands as pure addition**, adopted by nothing. Zero call sites change; it cannot break what does not call it.
- **One guard, one blueprint, one PR** thereafter.
- **A behavior change is a failure, not a merge conflict.** No fixing forward by editing the test to match the new behavior — that is the precise move this protocol forbids.
- Never mix a move with a fix; Grade A at every step; every step independently revertable; migrated code is deleted, not orphaned.

## Scope narrowed for impact

**In:** the access kernel only — the piece that structurally prevents the defect class actually costing us.
**Still deferred, with reasons stated:** the `routes.py` split (pure motion, and 17 test patch-sites couple to the module path), the `tools.ts` rewrite (two data-only fixes carved out), the ingest engine, the CareAgents boundary.

## Definition of done

Not "the kernel exists" — the program is done when each recurring defect class is structurally unavailable, measured by a guard-matrix row flipping from `xfail` to `pass` **without the test being edited**.

### DIFF
--- a/docs/2026-08-03-refactor-working-protocol.md
+++ b/docs/2026-08-03-refactor-working-protocol.md
@@ -1,0 +1,105 @@
+# Refactor working protocol — 2026-08-03
+
+**Founder decision (Eugene, 2026-08-03):** the architecture refactor proceeds
+now. The CTO and Product recommendation to defer all structural work past
+Aug 18 is **overruled on priority**. Their *design* objections stand and are
+being resolved before any code is written (see
+`docs/2026-08-03-access-kernel-spec.md`).
+
+The constraint, in the founder's words:
+
+> Do it carefully and intentionally, not rushed and break-fix — that defeats
+> the whole purpose. The goal: stop the bleeding and the constant issues,
+> align on quality, simple implementation, and scope for vision impact.
+
+This document is the protocol that constraint implies. It governs every PR in
+the refactor program.
+
+## Why this refactor exists (the bleeding)
+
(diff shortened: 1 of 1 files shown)


aks129:
**Rebased onto #317** to green the checks.

`python-tests` was failing here for a reason that had nothing to do with this PR: `main` is currently red (the write-guard matrix went stale when #308/#309/#315 merged), and a doc-only branch cut from `main` inherits those 10 failures.

Side effect a reviewer should know about: `tests/test_write_guard_matrix.py` now appears in this PR's file list. That is #317's change showing through the stack, **not** a test edit belonging to this PR. It disappears once #317 merges. The doc file is still the only content here.

Issue #304, opened 2026-08-03
Ops reconciliation sweeps are not tenant-filtered — a tenant's step-up token drives other tenants' actions


aks129:
Found by the architecture audit (PR #303, finding S-1). **Verified against current `main`.**

## What happens

`_auth_error_or_none()` ([r6/ops/routes.py:50-63](r6/ops/routes.py#L50-L63)) validates a step-up token bound to tenant X. The sweep that follows then queries across **all** tenants ([r6/ops/routes.py:216-228](r6/ops/routes.py#L216-L228)):

```python
ProposedAction.query.filter_by(status='executing').all()
ProposedAction.query.filter(ProposedAction.status == 'unknown', ...).all()
ProposedAction.query.filter(ProposedAction.status == 'awaiting_confirmation',
                            ProposedAction.expires_at <= now).all()
```

None filter on `tenant_id`. `_reap_lapsed_approval` then calls `notify_tenant(action.tenant_id, ...)` ([r6/ops/routes.py:195-198](r6/ops/routes.py#L195-L198)).

## Impact

Any tenant holding a valid step-up token can transition **another** tenant's pending approvals to `expired` and cause a Telegram notification to be sent to that other tenant's patient. On a private pilot tenant this surfaces as an approval the patient did not lose, plus an unexpected message — during a supervised session.

This is a cross-tenant authorization break: the token is scoped to one tenant, the effect is global.

## Suggested fix

Scope every sweep query by the authenticated tenant, **or** move `/r6/ops/*` to internal-secret auth (as `/internal/*` uses) so it is an operator surface rather than a tenant surface. The audit's §12 Q4 asks which; the CTO design pass is deciding.

## Test to add

A sweep authenticated as tenant A must not transition, or notify about, any row belonging to tenant B. `tests/ops/test_reaper.py` covers transitions but not tenant scoping.

aks129:
## Confirmed LIVE and UNAUTHENTICATED in production (2026-08-03)

Non-destructive probe against `app.healthclaw.io` (the sweep itself was **not** executed — that would mutate real tenants' data):

- `POST /r6/fhir/internal/step-up-token {"tenant_id":"desktop-demo"}` → **HTTP 200, token issued with no credential**. `desktop-demo` is a public tenant, so `_internal_mint_authorized` short-circuits its secret check.
- Same call for a non-public tenant → **HTTP 403 `{"error":"forbidden"}`** (control).
[...]

Issue #296, opened 2026-08-03
Four tool descriptions instruct the model to call a withheld tool and pass a rejected argument (write path)


aks129:
Found while verifying the real-records path for a design partner. Split out of #295, which fixes the docs; this is the code.

## The defect

Four tool **descriptions** — the text the model reads to decide how to call a tool — instruct a mechanism that does not exist on any hosted deployment:

```
tools.ts:848  fhir_get_token — "Returns the token string — pass it as
              _stepUpToken in subsequent write tool calls."
tools.ts:873  (result text)  — "Pass it as _stepUpToken in fhir_propose_write,
              fhir_commit_write, action_commit, shl_generate, or curatr_apply_fix."
tools.ts:983  action_commit  — "Requires step-up authorization (call
              fhir_get_token first; pass as _stepUpToken)."
tools.ts:1017 shl_generate   — "step-up required — pass _stepUpToken"
```

Two things are wrong with that instruction:

1. **`fhir_get_token` is withheld.** It is in `PRIVILEGED_TOOL_NAMES` (`tools.ts:128`) and never appears in a network transport `tools/list`. Verified against production: 27 tools, no `fhir_get_token`.
2. **`_stepUpToken` is not an accepted argument.** Verified against production — every write tool schema is closed and none declares it:

```
action_commit     -> action_id           | additionalProperties: None
fhir_commit_write -> operation, resource | additionalProperties: None
```

The handler reads the header instead (`tools.ts:1156`):

```js
const stepUpToken = headers?.["x-step-up-token"];
```

## Why it matters more than a stale string

This is the **write and action path** — the most safety-critical surface we have. An agent following the description will call a tool that is not there, then attempt an argument that is silently dropped. Best case it reports it cannot proceed and the patient is stuck; worse, a model that "helpfully" works around the missing tool is improvising on the rail whose whole purpose is that it cannot be improvised on.

There is also an auto-mint path (`tools.ts:158-195`) that populates `X-Step-Up-Token` in the forwarded headers, so in practice the server may already supply it. If so, the description is not just wrong about the mechanism, it is asking the model to do work that is already done.

## Suggested fix

Rewrite the four descriptions to state the real mechanism: step-up arrives in the `X-Step-Up-Token` header, supplied by the operator or the patient connect flow, not by a tool call. Confirm whether the auto-mint path makes it transparent to the agent and, if so, say that instead of naming a header the agent cannot set either.

## Guard

[...]

Issue #214, opened 2026-08-01
security: X-Human-Confirmed is still the entire human gate for direct clinical FHIR writes


aks129:
CLAUDE.md states the spoofable `X-Human-Confirmed` header 'is gone — approval is a separate endpoint.' That is true **on the action rail only**.

`enforce_human_in_loop` (`r6/health_compliance.py:91-131`) still checks the client-supplied header on POST/PUT of clinical resource types, and the write handlers have no out-of-band confirmation. Any agent holding a write step-up token self-confirms by adding one header.

Docs that currently contradict the code: `CLAUDE.md`, `.github/REVIEW_STANDARDS.md` item 4, `docs/development.md`, `SECURITY.md`. The stale REVIEW_STANDARDS text is what the AI PR reviewer enforces on every PR.

**Fix (two parts):**
1. Immediate: correct the four documents so the stated invariant matches reality.
2. Then: route clinical creates/updates through the propose→out-of-band-approve rail (or the audience-bound token from #211) and retire the header.

From the 2026-08-01 guardrail + dev-system audits.

PR #318, opened 2026-08-03
docs: access-kernel spec — corrected design + un-rushable migration order

aks129:
The CTO's implementable spec for the access kernel, resolving the two design **BLOCK**s from the review pass. Design doc only — no production code. Read this before slice 1 is built.

## Corrections to the original plan (§3 of #303)

| Original | Corrected | Why |
|---|---|---|
| `Profile.RAW_INTERNAL` | **Deleted.** `unredacted_response(endpoint=, reason=)` gated by a pinned `_UNREDACTED_EXITS` allowlist; `outcome_response()` separate | An enum member meaning "no property" launders existing defects into documented decisions. Growing the allowlist is now a two-file diff a reviewer sees. |
| `StepUpDenied` | `StepUpDenied(checked=)`, set on exactly one line inside `require_grant`; unchecked raises propagate as 500 | Otherwise a `StepUpDenied` from an unexpected place renders a clean 401 and a bug looks like the gate working. |
| `require_write_grant()` | `require_grant(*, scope: Scope, ...)` with `TENANT_BOUND` / `WRITE` | `r6/read_auth.py:92` deliberately passes `require_scope=None`; a write-only grant breaks the patient connect flow. |
| `tenant_from_request()` "never defaults" | `sources=` as an **ORDERED tuple** (correcting the CTO's own earlier `frozenset`) | A set forces the kernel to invent a precedence, and the invented one differs from `r6/command_center/routes.py:76-82` — that would silently change which tenant the dashboard renders. All 13 multi-source sites itemized. |
| Extend `before_request` hooks app-wide | **Handlers app-wide; hooks unchanged** | `r6/sdc/delivery.py:8-11` is deliberately off `r6_blueprint` — "the signature is the credential." |

Status-code normalization is **not** performed: `absent_status`/`rejected_status` carry today's dialect per site, and normalization becomes one deliberate PR after all sites migrate. That remains Eugene's call.

## Migration order

Slice 0 (matrix green — **PR #317**) → 0.5 (matrix into the Postgres lane) → 1 (kernel adopted by nothing) → 2 (handlers registered, inert) → 3 (`r6/smbp/routes.py:70-76`, the first real migration: one call site, majority dialect, and *not* on `r6_blueprint`, so it proves handlers reach the blueprints the hooks never covered) → 4–14.

The per-PR contract, of which the load-bearing rule is: **the diff to `tests/` in a migration PR is empty.** If a test must change, it is a separate preceding PR against the old code. That is exactly the rule #308/#309/#315 skipped, which is why `main` went red.

## Honest scope — what the kernel does NOT fix

Closes 2 of 6 recurring defect classes structurally (tenant-defaulting strategies; blueprints accepting any string as a tenant id) and makes 1 detectable (status dialects). It does **not** close:

- **5 unaudited 404s** — `audit()` is a function you must call; the answer is a testing-mode read-audit assertion.
- **Clinical write with no HITL (`/r6/smbp/reading`)** — **#214 stays a gap after this ships.**
- **The #304/#311 "auth scoped one way, action scoped another" class** — the real control is making `TENANT_FILTER` a parametrized matrix assertion (2 of 22 rows have one today).

## Quietest hazard found

If `r6/access.py` does `from r6.stepup import validate_step_up_token`, the 33 tests that `monkeypatch.setattr('r6.stepup.*')` keep passing while patching an object the kernel no longer consults. Module-attribute resolution is a hard rule in §1.0 with that reason attached.

## Parallel track, no kernel dependency

The `tools.ts` manifest generator + tier gate should run **now, in parallel**: worst defect density in the repo, S-6 breaks every OpenAI/Gemini tool call today, #296 is live on the write path, and it is a different tree.

### DIFF
--- a/docs/2026-08-03-access-kernel-spec.md
+++ b/docs/2026-08-03-access-kernel-spec.md
@@ -1,0 +1,788 @@
+# Access kernel — interface spec and migration order
+
+Status: **design spec. No production code written.** Supersedes §3 of
+[2026-08-02-architecture-audit-and-refactor-plan.md](2026-08-02-architecture-audit-and-refactor-plan.md)
+where the two disagree.
+
+Context: the founder overruled the "defer past Aug 18" sequencing. The
+refactor proceeds now, on one condition he stated himself — carefully and
+intentionally, not break-fix. This document answers *how*, and resolves the
+two design objections that blocked the kernel as originally specified.
+
+Every file:line below was verified against `origin/main` at `516cf5b`.
+
+---
+
+## 0. Blocking precondition: the safety net is red
+
+`tests/test_write_guard_matrix.py` is the characterization baseline the whole
+plan rests on. It does not pass on `main`:
+
+```text
+$ uv run python -m pytest tests/test_write_guard_matrix.py -q
+10 failed, 162 passed, 6 skipped, 4 xfailed
+```
+
+CI has been red on `main` since the matrix merged (`cd3f4dc`, run
+30814856634, and every push since). The cause is the drift shape from the
+retro, not a bad test:
+
+| Failure | Cause |
+|---|---|
+| 8 × `[ops-reap]` | #308 changed `/r6/ops/*` from tenant step-up to `internal_secret_authorized()` (`r6/ops/routes.py:49-56`). The matrix row still declares `TENANT_HEADER, TENANT_FORMAT, STEP_UP` (`tests/test_write_guard_matrix.py:531-541`). |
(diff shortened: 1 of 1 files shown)


aks129:
**Rebased onto #317** to green the checks.

`python-tests` was failing here for a reason that had nothing to do with this PR: `main` is currently red (the write-guard matrix went stale when #308/#309/#315 merged), and a doc-only branch cut from `main` inherits those 10 failures.

Side effect a reviewer should know about: `tests/test_write_guard_matrix.py` now appears in this PR's file list. That is #317's change showing through the stack, **not** a test edit belonging to this PR. It disappears once #317 merges. The doc file is still the only content here.

PR #323, opened 2026-08-03
fix(fasten): the consent dialog was unreachable on a short viewport (iPad)

aks129:
A user reported the Fasten dialog would not let them check the policy-agreement box on an iPad. The checkbox and the Continue button were below the fold with no way to scroll to them — so a patient could start the connect flow and be unable to finish it. This is the dead-end class #224 exists to remove, on the flow the webinar journey and the clinical pilot both depend on.

Same defect shape on both surfaces: **a container clipping a child taller than itself.**

## `templates/fasten_connect.html` — the page CareAgents sends patients to

```css
/* before */
.fasten-modal { height: min(640px, 90vh); overflow: hidden; }
.fasten-modal iframe { flex: 1; }
```

Three compounding problems:
1. A **fixed height cannot shrink** on a short viewport, and `overflow: hidden` removes the scrollbar that would have rescued it.
2. iOS measures `vh` against the viewport with browser chrome **expanded**, so `90vh` routinely exceeds what is actually visible.
3. **iOS and iPadOS expand an iframe to its content height and will not scroll it internally**, whatever the iframe's own overflow says. So no amount of iframe CSS fixes this — the scroll container has to be a wrapper element.

Now: a 640px preference with `max-height: 100%` resolved against a backdrop that is already the visible viewport minus safe-area padding; the iframe sits inside a scrolling `.fasten-modal-body`; safe-area padding clears the notch and home indicator. `min-height: 0` on the wrapper is load-bearing — a flex child defaults to `min-height: auto` and refuses to shrink, which re-creates the clip even with `overflow-y` set.

Also: the close control is now a **44px tap target** (design.md) with the glyph still at 22px, plus `aria-label`; the iframe gained a `title`.

## `templates/r6_dashboard.html` — the surface in the screenshot

`#stitch-container` carried `overflow: hidden` for a rounded corner. `<fasten-stitch-element>` renders its consent dialog **inside** that element, so the corner radius clipped the dialog. A rounded corner is not worth a dead end in the connect flow. Moved out of an inline style into a real rule in `r6-dashboard.css`, and short viewports get a `max-height` plus internal scroll for the vendor dialog.

## Guard

`tests/test_fasten_dialog_reachable.py`, verified in **both** directions: **5 pass on the fix, 5 fail on the old code.**

Honest limit, stated in the module docstring rather than implied: these assert the **source**, not a rendered layout, because the widget is a third-party bundle we cannot render in CI. They catch the regression shape we actually hit — not every way a dialog can be cut off. Worth a real device check on the iPad that reported it before calling it closed.

Full suite **2246 passed, 12 skipped, 6 xfailed**; ruff clean.

### DIFF
--- a/static/css/r6-dashboard.css
+++ b/static/css/r6-dashboard.css
@@ -467,3 +467,35 @@
 
 
 .footer-legal { font-size: 0.78rem; }
+
+/* ---------------------------------------------------------------------------
+   Fasten Connect (Stitch) widget container.
+
+   The container used to carry `overflow: hidden` (for the rounded corner).
+   <fasten-stitch-element> renders its own consent dialog INSIDE this element,
+   so that one declaration clipped the dialog's bottom — the policy-agreement
+   checkbox and the Continue button — with no way to scroll to them. Reported
+   from an iPad, where the viewport is short enough for the dialog to exceed
+   the container.
+
+   A rounded corner is not worth a dead end in the connect flow, so the clip
+   is gone. The widget's own dialog positions itself against the viewport.
+   --------------------------------------------------------------------------- */
+.stitch-container {
+  margin-bottom: 1.25rem;
+  border: 1px solid var(--r6-info);
+  border-radius: 8px;
+  overflow: visible;
+}
+
+/* The vendor bundle sizes its dialog for a desktop window. Give it the short-
+   viewport behaviour the rest of the product has: fit the visible area, and
+   scroll inside itself rather than off the bottom of the screen. iPad in
+   landscape and any phone in landscape both land here. */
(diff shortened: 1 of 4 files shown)


aks129:
**Rebased onto #324** (the `ip-address` SSRF advisory patch) to green `dependency-audit`.

That failure was unrelated to this change — a newly published npm advisory on a transitive production dependency (`@modelcontextprotocol/sdk` → `express-rate-limit` → `ip-address`). It began failing every new CI run regardless of content; older PRs show a stale pass because their runs predate the advisory.

Merge #324 first and this goes green on its own.

Issue #341, opened 2026-08-04
Production logs retain ~11 minutes: the careagents worker poll is 100% of the log volume


aks129:
## What I measured

Pulling the last 5000 lines of the `HealthClawGuardrails` production log:

```
$ railway logs --service HealthClawGuardrails -n 5000 | grep -c '"path": "/command-center/api/runs/claim"'
5000
```

**5000 of 5000 lines** are the same request. They span `00:47:08Z` → `00:58:45Z` — 11 minutes 37 seconds, i.e. **~7.2 requests/second, ~620k/day**, every one of them `204 No Content` (nothing to claim).

The HTTP log stream (`railway logs --http`) is the same: 2000 lines, zero non-poll requests.

## Why this is a defect and not just noise

This is the intended poll rate, not a misconfiguration — `careagents/config.py:109-110` defaults to `CARE_RUN_WORKERS=4` and `CARE_RUN_POLL_SECONDS=0.5`, and `run_worker_pool` (`careagents/worker.py:412-422`) sleeps `run_poll_seconds` only when a slot claims nothing. 4 slots × 2/s = 8/s, which matches the measured 7.2/s.

The problem is that **every one of those empty polls writes a request log line**, so the production log retains roughly 11 minutes of history. Anything that is not the poll loop is evicted before anyone can look at it.

Concretely: I went to check whether a real patient's Fasten/MEDENT EHI export had arrived. There was no way to see it. Zero `fasten`, `ingest`, `export`, `webhook`, or `reaper` lines survive in the retained window — not because nothing happened, but because the buffer is one worker's idle chatter. The webhook receipt, the ingest thread's progress, and the boot reaper's sweep all log, and none of it is observable.

This also silently defeats several things we already built:
- the boot reaper logs `%d zombie job(s) found` at WARNING (`r6/fasten/reaper.py:68`) — unreadable
- `_handle_export_success` logs unknown-connection and reprocessing decisions (`r6/fasten/routes.py:139,152`) — unreadable
- any 500 on any real user request — unreadable within minutes

## Suggested fix

Do not log a successful empty claim. Options, roughly in order of preference:

1. **Skip the request log for `204` on `/command-center/api/runs/claim`** — the narrowest change; the poll is the only caller and a 204 carries no information.
2. Log empty claims at `DEBUG` and set the service to `INFO`.
[...]

PR #345, opened 2026-08-04
fix(careagents): a provider 429 is not "something went wrong on our side"

aks129:
## The report

A patient asked *"give me a timeline of my cholesterol results"*. The UI showed eight tool chips — `Interpreting your labs`, `Checking preventive care gaps`, and **five** `Searching your records` — and then:

> ⚠️ Something went wrong on our side.

## What actually happened

Pulled the run from production. `agent_runs` for that tenant:

```
('completed', None,       2, 2026-08-04 01:33:03)
('failed',    'LLMError', 1, 2026-08-04 02:11:00)
```

and the worker traceback:

```
careagents.llm.LLMError: model call failed (HTTP 429)
```

**Nothing was wrong on our side.** The provider rate limited one request, the first attempt was the only attempt, and a transient throttle was shown to a patient as a defect in their health records.

## Two fixes

**1. Retry transient failures.** `_openai_complete` raised on the first non-200 with no retry at all. It now retries 408/425/429/5xx up to `MAX_ATTEMPTS` with exponential backoff, honouring `Retry-After`.

Three deliberate refusals, each with a test:

- **4xx client errors are not retried.** Retrying a malformed or unauthorised call just spends the run's deadline before failing identically.
- **A `Retry-After` longer than we can hold the lease gives up immediately** rather than sleeping past the deadline and failing anyway. Failing now, honestly, lets the person re-ask; sleeping first just wastes their time to reach the same place.
- **The Anthropic path gets classification only, no retry loop.** The SDK already retries 429/5xx internally, so wrapping it would multiply the delay while the run holds its lease. What was missing there was never the retry — it was the classification.

`requests.RequestException` is now caught too. A connection reset or read timeout previously escaped as a raw requests exception, so the worker recorded an error class that named the HTTP library rather than the failure.

**2. Say something true.** `LLMRateLimited` (a subclass of `LLMError`, so every existing handler still catches it) selects:

> I'm getting more requests than I can answer right now. Nothing is wrong with your records — try asking again in a moment.

The generic text fails twice over: it invites a bug report for a defect that doesn't exist, and it leaves a patient wondering whether their health records are broken.

## Tests

18 new in `tests/test_llm_retry.py`, each naming the mutation that kills it. Full suite 2357 passed, 12 skipped, 6 xfailed. `uv run ruff check .` clean.

## What this does NOT fix

**The five `Searching your records` calls are the real story.** The model looped because every search came back as unlabeled records, so it kept trying other resource types looking for something readable. That inflates the request until it trips the rate limit. #344 (terminology lookup) is the actual fix for the loop; this PR only stops the loop's failure mode from being a lie.

Also seen while diagnosing, filed separately: the worker's claim poll is getting **24 HTTP 502s in 43 log lines** from HealthClaw, which is consistent with the app being saturated by its own 7.2 req/s poll storm (#341).

### DIFF
--- a/careagents/llm.py
+++ b/careagents/llm.py
@@ -19,6 +19,7 @@ Tools use a neutral shape: {"name", "description", "parameters": JSONSchema}.
 from __future__ import annotations
 
 import json
+import time as _time_mod
 from dataclasses import dataclass, field
 
 import requests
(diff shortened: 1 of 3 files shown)


aks129:
Pushed the D5 follow-up from the review. Consolidating the four string copies surfaced something worse than drift: **`run_turn` yielded `str(exc)` as patient-facing text**, so the streamed chat and SMS surfaces showed the literal `model call failed (HTTP 429)`. Now: `agent.py` owns the text and `failure_text()`; worker/app/sync-path all route through it; a repo-grep test pins single ownership. 2365 passed, ruff clean.

PR #409, opened 2026-08-05
fix(careagents): an outage is not "you have no records" (#403)

### LINKED ISSUE #403: A HealthClaw outage reaches the patient as 'you have no records' and a spinner that never ends
Characterized in PR #402 as strict xfails. **Sixth instance of this week's pattern**, and the one on the journey a brand-new patient sees on their phone.

## Two functions, both on the connect screen

**`careagents/healthclaw.py:311` `record_count()`** swallows `HealthClawError` per resource type and returns **0**. `careagents/app.py:526` baselines a refresh on that number; `app.py:554` reports growth against it.

During an engine incident, the patient is told **as fact** that their record count is zero.

**`careagents/healthclaw.py:297` `tenant_has_records()`** returns **False** on any `HealthClawError`, which `careagents/app.py:544` renders as `{"status": "pending"}`.

During an incident, the connect screen says **"still fetching your records" forever**, and nothing anywhere says the store is down.

The first tells a patient something false about their health record. The second tells them nothing at all, indefinitely, at the moment they are deciding whether this product works.

## Three more on the same seam

- **16 of 23 methods do not wrap `requests.RequestException`.** A refused connection escapes raw and reaches Flask as an unhandled 500. #267 fixed `ingest_bundle` rather than the seam, so `search`, `read`, `interpret_labs` and `care_gaps` are all still exposed.
- **Success-path `r.json()` is unguarded everywhere**, including the durable-run methods that *do* wrap the request. In `ingest_bundle` an unparseable 200 becomes `{}` and reports a successful ingest of nothing.
- **A wrongly-shaped 200 is returned as data** — `search()` hands a JSON string to the agent as a Bundle.

## The fix already exists in the file

`agent_worker_health()` handles this correctly. Every finding here is "apply that to the rest of the seam."

The shape is the one applied five times this week (#379, #381, #389, #390, #400): three states, not two. `0` and `False` are answers; an outage is not an answer. A caller must be able to tell "we looked and there are none" from "we could not look."

## Why this is worth doing before the type-level fix

The CTO's proposal to make unavailable-vs-absent a **type** was deferred past Aug 18. That is the right call for the terminology layer, but this seam is different: it is the boundary every CareAgents call crosses, the failure is patient-visible on the first screen, and the corrected pattern is already sitting in the same file.

## Verification

Pinned as `xfail(strict=True)` in `tests/test_healthclaw_transport.py` with the live consequence in each reason, so a fix flips them rather than silently satisfying them:

```
uv run python -m pytest tests/test_healthclaw_transport.py -q -k "outage or pending" --runxfail
→ 2 failed  (both DID NOT RAISE HealthClawError)
```

Note for whoever fixes it: a first attempt that widened `fetch_appointment_brief`'s `except` did **not** flip its test, because that method never raises — it returns the wrong shape. The test refuses a wrong fix.

Related: #294, #374, PR #402.

aks129:
Closes #403.

## Stacked PR — read the base

This is based on `test/scorecard-discovery-and-transport-coverage`, **not**
`main`, and targets that branch. The strict-xfail pins this fix flips were
added in #402 and are not on `main` yet; branching from `main` would have
produced a PR that could not show the pins turning green. Merge #402 first.

## What was wrong

`careagents/healthclaw.py` is the seam every CareAgents call crosses, and it
answered "we could not look" with the same value as "we looked and there are
none".

| | before | after |
| --- | --- | --- |
| `record_count()` on an engine outage | `0` — the patient is told, as fact, that they have no records | raises `HealthClawError` |
| `tenant_has_records()` on an outage | `False` → `{"status": "pending"}` → "still fetching your records", forever | raises; the endpoint answers `503 records_unavailable` |
| refused connection, 16 of 23 methods | raw `requests` exception → unhandled Flask 500 | `HealthClawError`, on all 23 |
| unparseable 200 in `ingest_bundle` | `{}` → reported as a completed upload of zero records | raises |
| non-object 200 from `search()` | handed to the agent as a FHIR Bundle | raises |
| non-object 200 from `fetch_appointment_brief()` | a bare string handed to the brief renderer | `None`, per its own docstring |

## The shape

No new policy — `agent_worker_health()` already did this correctly. Its two
halves are now `_send` (transport failure becomes the one type CareAgents
catches) and `_json_object` (a success body is decoded **and** type-checked,
or it is a failure), and every other method goes through them. Same three
states as `_indeterminate()` in `r6/labs/interpret.py`: `0` and `False` are
answers, an outage is not.

Error messages on the already-correct durable-run paths are byte-identical, so
those methods change no observable behaviour.

## The caller side

A third state that the caller ignores fixes nothing, so:

- `poll_connection` (`careagents/app.py`) answers `503 {"status":
  "unavailable", "error": "records_unavailable", "message": …}` instead of
  `pending`.
- `careagents/static/home.js` renders that message on the connection card —
  in both the pending poller and the post-refresh watcher — and keeps
  polling, so the card clears itself when the engine returns. The patient
  sees "We couldn't reach your records right now", never a count and never an
  endless spinner.
- `/api/trust` keeps answering `unavailable` rather than 500ing, now that
  `conformance_badge()` can raise.

The handlers already in place at `app.py:441`, `:526` and `:554` were dead
code — `record_count` could not raise. They are live now.

## Pins flipped in this PR

Per [docs/agent-task-guide.md](docs/agent-task-guide.md) section 6, a fix
does not land while a test still describes the route as broken. 56 rows went
`XPASS(strict)` on the fix and are now plain assertions;
`test_a_raw_requests_exception_really_does_escape_the_unwrapped_reads` is
deleted, as its own docstring instructed — it asserted the defect.

`tests/test_healthclaw_transport.py` has **no `xfail` marker left**. The
three remaining xfails in the suite are #304/#305 in
`tests/test_write_guard_matrix.py`, untouched here.

## Mutation evidence

Each mutation applied with `PYTHONDONTWRITEBYTECODE=1`, run, seen red,
reverted:

| mutation | result |
| --- | --- |
| restore `except HealthClawError: continue` in `record_count` | 1 failed |
| restore `except HealthClawError: return False` in `tenant_has_records` | 1 failed |
| drop the `except requests.RequestException` arm from `_send` | 23 failed |
| drop the decode guard from `_json_object` | 14 failed |
| drop the `isinstance(body, dict)` arm from `_json_object` | 6 failed |
| return `body or {}` on `ingest_bundle`'s 200 path | 1 failed |
| return `r.json()` from `fetch_appointment_brief` | 1 failed |
| drop `poll_connection`'s `except HealthClawError` | 1 failed |
| drop `/api/trust`'s `except HealthClawError` | 1 failed |

Note for the record: widening `fetch_appointment_brief`'s `except` does
**not** flip its test. That method never raised — it returned the wrong
shape, and the type check is what fixes it.

## Verification

```
2582 passed, 12 skipped, 3 xfailed, 2 warnings in 130.43s

uv run ruff check .
All checks passed!
```

Conformance stays Grade A (`tests/test_guardrail_conformance.py`, in the run
above). The pin file's live-engine tests exercise the changed client against
a real HealthClaw over a real socket, not a fake.

## Deliberately not touched

- `recent_messages`, `log_message`, `claim_inbound_message` and
  `fetch_appointment_brief` still degrade rather than raise. That is their
  documented contract and the tests pin it.
- `review_submit` and `telegram_bind` can still return an unhandled 500 when
  HealthClaw is unreachable. That was true before this change (the raw
  `requests` exception 500'd too); typing it does not make it worse, and
  route-level handling there is separate work.
- `_agent_owns_action` maps a transport failure to "that form isn't yours".
  Pre-existing, and worth its own issue.


aks129:
Reviewed. **The seam fix is right and the mutation evidence is the strongest this repo has had** — 23 tests red on dropping `_send`'s `except`, 14 on the decode guard, 6 on the `isinstance` arm. `_send` + `_json_object` is the correct generalisation of `agent_worker_health()`, and keeping the durable-run error strings byte-identical is what makes 56 xfail flips reviewable.

Two verifications, both independent of the report:

- No `xfail` marker remains in `tests/test_healthclaw_transport.py`. The single remaining occurrence of the word is prose. The rule from `docs/agent-task-guide.md` §6 is satisfied.
- `MedicationStatement`-style reasoning checked the other way: the issue's warning about `fetch_appointment_brief` held. Widening the `except` does not flip that test, because the method never raised — it returned the wrong shape.

**Holding merge for one thing**, filed as #410 and sent back to this branch.

`_agent_owns_action` was reported as pre-existing. It is not, quite. `origin/main:careagents/healthclaw.py:227-233` shows `action_status` did **not** wrap `requests.RequestException`:

- **Before:** refused connection → raw exception → Flask **500**.
- **After:** typed `HealthClawError` → caught at `app.py:897` → `None` → **"That form isn't yours."** 404.

A 500 is ugly but honest. "That form isn't yours" is a confident false statement about ownership, on the human-approval path — the guardrail this product sells. Four more sites are in the same class (`:828`, `:849`, `:1084` report an outage as "unknown run" or "not found"; `:716` flips an availability flag to `False`), and this PR widens how many transport conditions reach them.

So the seam fix is correct and relocates the defect to the callers. That is the shape the fix exists to remove, and it should not ship as a follow-up issue. Table and the proposed rule — `HealthClawError.status`: an engine 404 is a real not-found, a transport failure or 5xx is not — are in #410.

The house style is already in this file (`:298`, `:527`, `:807`, `:1061` → 503; `:878` → 502) and this PR adds another instance of it at `:544-558`. Matching it at the remaining five closes the loop.

PR #453, opened 2026-08-08
design: Swiss editorial for healthclaw.io, and one owner for the CSP

aks129:
Redesigns both public surfaces from first principles, and fixes the contradiction the redesign kept tripping over.

## Why this direction

The brief was to match what wins on Awwwards. Looking there honestly says two things.

Its **healthcare** category is dated — the most-cited entry, [Possible Health](https://www.awwwards.com/sites/possible-health), is an Honorable Mention from **May 2014**: WordPress, parallax, 7.4/10. Its **typography** category is current, and the most decorated recent winner, [MONOLOG](https://www.awwwards.com/sites/monolog) (Typography Honors + Developer Award + SOTD, July 2026), is two colours with typography as the primary element, built on GSAP and Three.js.

What makes that win is the two-colour discipline and the typographic hierarchy. The Three.js is the part we cannot ship — in-app webviews, `prefers-reduced-motion`, and a person checking whether they are due for a screening — and it is not the part doing the work.

So: **International Typographic Style**. Strict grid, hierarchy from size and weight alone, hairline rules instead of shadows, one grotesque plus one monospace, tabular figures. Zero JavaScript, so it survives every constraint intact.

## The actual root cause of the drift

healthclaw.io **had no stylesheet**. `templates/index.html` carried ~200 inline `style=` attributes and inline `onmouseover` handlers, so every section re-invented its own card and no visual fact had an owner. Swapping hex values would have drifted back within a month.

`static/css/healthclaw.css` is the owner now; `index.html` carries content and structure only.

`skills.html` had gone further and declared a **whole second palette** (`--cyan`, `--bg-card`, `--border`, `--gray`) shadowing the site's own, plus a `JetBrains Mono` that `base.html` has never loaded — so every identifier there silently fell back to the platform monospace.

## One owner for the CSP

| source | claimed |
|---|---|
| `design.md` | `default-src 'self'`, no CDNs |
| `app.py` | four CDN hosts allowed |
| `check_table_stakes.py` | two Google hosts exempted |

Same fact, three owners, three answers. `app.py`'s own comment explains how: the strict policy had *silently broken styling and fonts on every deploy*, and whoever fixed the deploy did not own the two documents describing it.

Fixed by **making the claim true rather than relaxing it**. Bootstrap, Font Awesome and all four webface families are vendored under `static/`. The policy is now genuinely `default-src 'self'`; `design.md` describes it; the checker **parses it out of `app.py`** instead of restating it, and fails closed when it cannot parse.

This is also a privacy fix: every page a patient opened — including a lab result — announced itself to `fonts.googleapis.com` with their IP, the referring URL and their user agent.

## Surfaces

**healthclaw.io** — paper `#FAF9F5`, ink `#14140F`, one ultramarine `#1B2FBF`. The accent sits outside the semantic set on purpose: green/amber/red already *mean* pass/caution/fail here, so an accent in those hues would make a decorative chip read as a status claim. Archivo + Fragment Mono, vendored. No gradients, no glow, no shadows — the `--cyan-glow` and `.hero-glow` it used to wear were the exact tell.

**base.html** — custom chrome replacing the stock Bootstrap navbar. The ground is a Jinja block that **defaults to the existing dark console**, so the operational dashboards render exactly as before; `faq`/`wiki`/`skills`/`privacy`/`terms` opt in to paper. `privacy.html` and `terms.html` carried byte-identical `<style>` blocks that had already drifted apart by one rule; both use a shared `.legal` component now.

**CareAgents** keeps its warm editorial identity — it was already the strong one. What changed is compliance with rules `design.md` already stated:

- focus rings restored — two `:focus { outline: none }` on the **patient app**, with a 1.5px border-colour change as the only substitute
- `prefers-reduced-motion` honoured — nine transitions, zero handling
- tabular figures, and error text raised off 14px

## A bug this found, and the guard that missed it

Vendoring moved `fontawesome.min.css` a directory deeper, so its `url(../webfonts/…)` resolved to `static/css/webfonts/` and **every icon 404'd**. Invisible to the Flask test client and to CI, which render HTML and never fetch a subresource; a missing font degrades to a placeholder glyph rather than an error. Only a real browser asked for the file.

The vendoring script's own check had **passed** — it verified the files were in a directory handed to it rather than resolving the URL the way a browser does. It confirmed a fact nobody was asking about. Both that check and a new `tests/test_vendored_assets_resolve.py` now resolve relative to the stylesheet, and both were mutation-tested against the exact defect.

## Verification

- **2781** pytest passed · **42** Playwright e2e passed · ruff clean · table stakes clean · conformance **Grade A** (35 passed)
- Every page loaded in Chromium: **zero** CSP violations, **zero** third-party requests, **zero** broken assets
- Mutation-tested: the CSP checker in both directions and on a parse failure; the asset-resolution guard against the shipped 404

Vendored bytes total 1.1MB. Licences permit redistribution — Archivo, Fragment Mono, Fraunces and Public Sans are OFL; Bootstrap is MIT; Font Awesome Free is CC BY 4.0 / SIL OFL / MIT.

## Pins that moved, per §6

Both say why in place:

- `test_subscribe` asserted `/api/subscribe` appears literally in the page. The handler moved to `static/js/landing.js`, so the assertion was pinning where code lived rather than that the form works; it now follows the code to the file.
- `test_google_fonts_is_allowed_because_the_csp_allows_it` asserted Google Fonts passes. The policy no longer allows it. Replaced with a pair that reads the live policy, plus a fail-closed case.

## Not done, deliberately

- The dashboards keep Font Awesome and their dark palette — out of scope, and they run the demo 11 days before the webinar.
- **CareAgents has no CSP of its own.** Worth its own change rather than a design PR, but worth knowing now.


aks129:
### Two follow-up commits since review — both from looking at the deployed preview

**`d07d065` — the lint gate caught three findings my local run had reported clean.**

The reason matters more than the findings. I ran `check_table_stakes.py --base origin/main` with the work still uncommitted, so `origin/main...HEAD` was an empty range, the script examined nothing, and printed `table stakes: clean`. Findings and "no input" printed the same word.

It now counts what it examined: `clean (13 file(s) checked)` versus `NOTHING TO CHECK — no added lines vs <base> ... this is not a pass.` That is the generator this repo's own review doc is about, sitting in the tool that enforces it.

The three findings: `bootstrap.bundle.min.js` flagged for not honouring `prefers-reduced-motion` (we vendor it and cannot fix it — vendored paths are exempt now), and `skills.html` / `wiki.html` declaring hover transitions with no local guard.

**`bb9a527` — the preview showed the primary CTA in ultramarine on near-black, and the whole masthead blue.**

One cause: the base link rule was `.surface-paper a`, specificity (0,1,1), which beat every single-class component under it. `.btn--primary` and `.hc-nav__link` both set a colour and both lost. It is `.surface-paper :where(a)` now — `:where()` contributes zero specificity, so a base element rule sits *under* the components instead of over them.

Two more found while looking: no `scroll-padding-top`, so in-page jumps landed under the sticky masthead; and `.code-block__copy` at 32px against a documented 44px floor with no small-control exemption.

**Every existing guard on this design system reads source and none of them could have caught any of that.** Each rule was correct in isolation; only the computed style was wrong. So `e2e/tests/design-invariants.spec.ts` measures `getComputedStyle` and does the WCAG arithmetic — CTA and body contrast against whatever background actually paints, nav colour, scroll-padding versus the measured masthead height, horizontal overflow at 390/768/1440, and every control height at 390px.

The tap-target check found the 32px button on its first run.

**Now:** 48 e2e passed, 2783 pytest, ruff clean, all 8 required checks green on a branch up to date with `main`. Mutation: restoring `.surface-paper a` turns the contrast and nav assertions red.

Preview: https://healthclaw-git-design-typographic-redesign-aks129s-projects.vercel.app

Issue #422, opened 2026-08-05
_resolve_subject counts soft-deleted Patient rows, so a tombstone still reads as ambiguity


aks129:
Found while fixing #415. Small, but it makes a whole class of cleanup ineffective and it is quietly load-bearing.

## What happens

`_resolve_subject` in `r6/caregaps/routes.py` counts Patient rows for a tenant without filtering `is_deleted`. A soft-deleted Patient therefore still counts toward the ambiguity check that decides whether `$care-gaps` can pick a subject.

Consequences:

1. **A tenant that has correctly deleted its extra patients still reads as ambiguous.** The operator does the right thing, the symptom does not move, and the obvious next step is to reach for a hard delete against production — which is a much worse operation to be nudged into.
2. **`delete-my-records` does not fully take effect for this path.** A patient who removed a record still has it counted when the system decides whose preventive care to evaluate.

## Why it matters beyond tidiness

The ambiguity guard is a real guardrail: it is what stops `$care-gaps` guessing which of several patients to evaluate, which is #389's defect with a coin flip added. A guard whose input includes rows the system considers deleted is deciding on a set nobody can see or reason about, and the only way to discover that is to delete something and watch nothing change.

## Fix

Filter `is_deleted` in the count. Then check the same question everywhere else a resource count feeds a decision rather than a display — this is unlikely to be the only place, and a count that silently includes tombstones is the same shape wherever it appears.

## Proof required

Soft-delete a Patient in a two-patient tenant; `$care-gaps` with no subject resolves to the survivor rather than answering `ambiguous-patient`. MUTATION: drop the `is_deleted` filter and the test reddens.

Worth a sweep in the same PR: `grep` for Patient/resource counts that feed a branch, and state which ones were checked.

Related: #415, #389, #393, #217 (purge leaves orphans).

PR #457, opened 2026-08-10
fix(demo): the tenant stops duplicating itself, and the bot answers what it advertises

aks129:
Three findings from Dr. Magan's dry run before the launch recording. All three read on camera as "the product is broken."

## 1. The demo tenant duplicates itself on every deploy

> "/conditions shows about a dozen duplicate Type 2 diabetes mellitus entries and the labs repeat too, so it seems the tenant may have been seeded more than once."

It had — about a dozen times. `railway.toml` runs `seed-demo --tenant-id desktop-demo` as a pre-deploy command on **every deploy**, and its own comment calls that seeding "idempotent". It was not: six of the seven built-in resources carried no fixed `id`, so each took a fresh UUID and was inserted again.

Production right now:

| resource | in the tenant | in the seed set |
|---|---:|---:|
| Patient | **19** | 1 |
| Condition | **12** | 1 |
| Observation | **40** | 3 |
| MedicationRequest | **11** | 1 |

Every deploy added a whole new synthetic patient with their own diabetes, labs and metformin order.

The failure was silent in a specific way worth naming: the one resource that *did* carry a stable id (`healthclaw-intake`) collided on the primary key every single deploy and was swallowed as a `logger.warning`. The component that could have shouted was the one being handled quietly; the six that duplicated did it somewhere only a human reading the UI would ever look. A physician advisor found it from the outside.

**Fix:** every seeded resource carries a stable id, and the seed skips what is already present rather than inserting and catching the collision.

## 2. A typed sentence got no reply at all

> "The conversational BP flow seems not to answer. I sent 'my blood pressure this morning was 150 over 94' and got no reply, so I suspect it wants different phrasing or something isn't wired on this tenant."

Not phrasing. `main()` registered exactly one `MessageHandler`, on `filters.COMMAND`. Nothing was listening for text, so the message reached no handler and produced silence — indistinguishable from a hung backend, which is exactly why she could not tell which it was.

Plain text now routes to `on_text`. A recognised reading is classified **server-side** by `r6/smbp/triage.py`, the shared 2025 AHA/ACC logic the clinician report and flags already use. `openclaw/Dockerfile` ships only `bot.py`, so those thresholds are not importable in that container — and copying them in is precisely the drift that module exists to prevent.

Two behaviours pinned because getting them wrong is dangerous rather than merely broken:

- **Unreachable classifier means say so, never guess.** An invented "you are fine" is the one output worse than an error.
- **Red-flag symptom means 911 regardless of the number.** `classify()` evaluates symptoms first and independently; a handler that checked the number first would reassure a symptomatic patient at 118/76.

The parser refuses `see you on 10/11` and `I rate it 9 over 10`. Putting a triage band on a date is worse than not answering.

## 3. /start advertised two commands that did not exist

`/summary` and `/curatr_fix` were both on the menu with no handler, so each answered *"Unknown command. Try /start for the command list."* — the list that had just offered them. `/start` is the first thing the recording shows.

`/curatr_fix` now has its handler (the function existed, reachable only as `/curatr fix`). `/summary` is removed, because nothing implements it. `cmd_start` already carried a comment saying this text must promise only controls that exist; the rule was being applied to the risk banner and not to the command list three lines below it.

## Two things the suite caught in this change

- **`test_ratchets.py` flagged my own existence check as soft-delete-blind.** A tombstoned demo patient would have read as "already seeded" and left the tenant empty — #422's shape in a second place. Now filters `is_deleted=False`.
- **`test_seed_demo.py` pinned the opposite behaviour** ("the Patient keeps a generated UUID"). Moved in this PR per the repo rule, with the reason recorded: the pin was right about the mechanism and was what kept the consequence invisible.

## Verification

- 2844 passed, 13 skipped, 1 xfailed
- `uv run ruff check .` clean
- Guards mutation-tested: drop the fixed ids, drop the TEXT handler, compute the band locally, swallow the failure — each goes red

## Not fixed here

Filed separately: `/curatr` evaluates a single Observation so it cannot see the duplicates; the `/start` banner's "your own records" on a shared public tenant; the validator reporting "validation passed" after checking two fields.

**Production still holds the 19 patients** — this stops the bleeding, it does not clean up. That needs a separate, authorized purge.

### DIFF
--- a/openclaw/bot.py
+++ b/openclaw/bot.py
@@ -27,6 +27,7 @@ STEP_UP_SECRET       HMAC secret for step-up tokens.
 
 import json
 import logging
+import re
 import os
 import time
 from collections import deque
@@ -352,6 +353,13 @@ async def cmd_start(update: Update, context: ContextTypes.DEFAULT_TYPE) -> None:
     # that does nothing leaves the user worse off than saying nothing, because
     # they choose to continue on the strength of it. If summary-only ships,
     # add it back here and to the privacy policy's mitigations list together.
+    #
+    # The SAME rule governs the command list below, and it was broken there:
+    # /summary and /curatr_fix were both listed with no handler registered,
+    # so each answered 'Unknown command. Try /start for the command list.' —
+    # the list that had just offered them. /curatr_fix now has its handler;
+    # /summary is gone, because nothing implements it.
+    # tests/test_bot_answers_what_it_advertises.py holds the two in sync.
     risk_line = (
         '⚠️ Heads up: chat apps aren’t encrypted medical channels. '
         'You’re accessing your own records here; by continuing you accept '
@@ -367,7 +375,6 @@ async def cmd_start(update: Update, context: ContextTypes.DEFAULT_TYPE) -> None:
         '/hbo\\_connect — authorize Health Bank One (OAuth)\n'
         '/hbo\\_pull — pull + ingest HBO verified records\n'
         '/dashboard — open the command center (signed 24h link)\n'
-        '/summary — high-level review of your record\n'
         '/conditions — list Conditions\n'
         '/labs — recent lab results\n'
         '/curatr — run Curatr data-quality evaluation\n'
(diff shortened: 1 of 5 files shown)


aks129:
## Correction to this PR's description

I overstated the cause here and want it on the record before it gets quoted.

**What I claimed:** the demo tenant duplicates "on every deploy", and production's 19 Patients were ~12 deploys' worth of re-seeding.

**What the evidence actually shows**, checked against production after this merged:

- The newest `agent_id=seed` AuditEvents in `desktop-demo` are dated **2026-07-08**. The pre-deploy seed has not run since.
- After this PR deployed (active deployment `4efbbf4`, which includes it), **none** of the new stable-id resources exist:
  `Patient/demo-patient-rivera`, `Condition/demo-condition-dm2`, `Observation/demo-obs-a1c`, `MedicationRequest/demo-medreq-metformin` all 404. The pre-existing `Questionnaire/healthclaw-intake` still 200s.
- Breaking the 19 Patients down by name shape:

  | count | patient | source |
  |---:|---|---|
  | 12 | Rivera, Maria Elena | the demo seed, run repeatedly |
  | 6 | Zzyzxbarton, Quintavious | the **conformance harness** probe patient |
  | 1 | other | `demo-loop-pt-8b870d38` |

**So:**

1. The non-idempotent seed is real and the fix is correct — 12 duplicate Riveras is exactly what it produces, and matches the advisor's "about a dozen".
2. But it is **not currently running per-deploy**, which is why the fix's stable-id resources have not appeared in production. Something changed about the pre-deploy step after 2026-07-08 and I have not established what.
3. **A second source I missed entirely:** running the conformance harness against `desktop-demo` creates a synthetic Patient per run. Six of them are sitting in the demo tenant. `scripts/guardrail_conformance.py` line 11 documents `--tenant desktop-demo` as *the usage example*, so this is the tool behaving as documented and polluting the demo tenant as a side effect.

Item 3 deserves its own issue: the harness should not default-document a tenant anyone demos from.

Item 2 means this PR stops a bleed that is currently paused rather than active. The fix is still right — it makes re-seeding safe whenever it resumes — but I should not have described it as ongoing without checking.

No production data has been changed. The cleanup to one patient is still pending explicit authorization.

Issue #520, opened 2026-08-17
actions: a human who declines and a human who never looked produce the same record


aks129:
Surfaced while writing `docs/specs/2026-08-16-human-gate.md` (the set-4 architecture review). Verified in code before filing.

## What

`r6/actions/models.py:34`:

```python
_TRANSITIONS = {
    'proposed': {'awaiting_confirmation', 'expired'},
    'awaiting_confirmation': {'executing', 'expired'},
    ...
}
```

There is **no `declined` state**. From `awaiting_confirmation` an action can only become `executing` or `expired`. So a person who reads the proposal and says **no** lands in exactly the same record as a person who **never opened it**.

## Why it matters for this set specifically

Set 4's whole claim is *"nothing executes without an out-of-band human step."* The evidence for that claim is the record of what the human did. If the record cannot distinguish a refusal from a timeout, then:

- **The audit trail is weaker than the gate.** We can prove nobody approved. We cannot prove someone objected — which is the more important fact, and the one a patient or a clinician would care about.
- **A safety signal is lost.** Repeated declines on a rail are the strongest possible signal that the proposals are wrong. Today they are indistinguishable from an inattentive user, so nothing can act on them.
- **The patient's own action is unrecorded.** Someone who deliberately said "no, do not send this" leaves no trace that they did.

## Reachability

Low urgency, and stated honestly: `#215` says no Tier-2 approve surface exists yet, so today there is no UI where a human *could* click decline. This is a gap in the state machine ahead of the surface that would exercise it — the same "fix the reader before the writer" shape as #422 and #509, and the same argument for doing it now: the day the approve surface ships, decline is a button that silently records nothing.

## Fix, roughly

- Add `declined` as a terminal state, reachable only from `awaiting_confirmation`.
- Record who declined and when, on the same footing as an approval.
- Do **not** make it inferrable — a decline is an explicit human act, exactly like an approval, and must never be synthesized from a timeout.

MCP's elicitation model already carries this distinction as three actions (`accept` / `decline` / `cancel`), where decline and cancel are deliberately different. Ours collapses all three into two.

Related: #215 (the approve surface), #216 (reserved-but-unenforced controls), #214.

Issue #216, opened 2026-08-01
actions: allowlist + daily cap error codes are reserved but enforced nowhere


aks129:
`CONTACT_NOT_ALLOWLISTED` and `DAILY_CAP_REACHED` exist in `r6/actions/errors.py` and are referenced **nowhere else in the codebase**.

Spec v3's acceptance bar for calls/SMS is 'allowlisted referenced contact' and 'capped'. Today an agent can propose a call to any phone number, unlimited times — the human tap is the only defense.

**Fix at propose time:** payload phone must match a contact the account holder saved; count same-day phone-call `ActionEvent`s per tenant against a cap.

Must land before any real outbound dial. From the 2026-08-01 actions audit.

Issue #215, opened 2026-08-01
actions: no Tier-2 approve surface exists — nothing could execute even with vendor keys


aks129:
The phone-call and SMS rails are code-complete but have **no approve surface**. The only callers of `/r6/actions/<id>/confirm` in the entire repo are the CareAgents form-fill review relay (`careagents/app.py:478`) and `scripts/demo_smoke.py`.

The spec's 'authenticated dashboard/Telegram handler' for the confirm card was never built. `r6/telegram_push.py` is push-only; no dashboard template has an approve button. The commit response message says 'approve in your HealthClaw dashboard' — pointing at a UI that does not exist.

**Consequence:** even if Bland onboarding cleared tomorrow, no real user could approve a call.

**This is the piece entirely in our control** and it is required for *every* future Tier-2 action, regardless of vendor. ~1 day, mirroring the existing review relay: an approve page at `/approve/<agent_id>/<action_id>` (login_required, `_agent_owns_action` guard) showing recipient, number, and the verbatim script, whose server-side handler calls `hc.confirm_action`.

Should be built with the audience-bound credential from #211 so the human gate is true by construction.

From the 2026-08-01 actions audit.

aks129:
The engine half is in #742: `GET /r6/actions/<id>/review` renders an approve page for every kind (recipient, number, verbatim script; the change list for a curatr fix) from the sealed payload, and its POST records the review-page confirmation. The CareAgents relay already proxies it for any kind, so a signed-in person can approve a phone call, text or record correction from `/review/<agent>/<action>` today. Still open after #742: nothing tells the person that URL for an action proposed over MCP or Telegram (needs the CareAgents agent id), and CareAgents has no pending-approvals list. Those are the remaining half of this issue.

aks129:
Closed: both halves shipped. #742 gives every action kind an approve page rendered from its sealed payload (recipient, number, verbatim script; the change list for a record correction), relayed by CareAgents. #748 adds 'Waiting for you', a signed-in list of pending requests with server-side ownership, which is how an action proposed over MCP reaches its owner. #751 makes the commit push and next_step name that surface. The real chain (propose → find → review → approve → execute → reopen) is tested end to end against the engine.

PR #521, opened 2026-08-17
docs(spec): define the human gate and the Tier-2 approve surface (#215, #216, #214)

aks129:
**Step 3 of `docs/2026-08-16-delivery-process.md` — architecture review. Design only, no production code.**

Set 4's PRD says it plainly: the gates are real and pinned (428/401/428/201), and **#215 says nothing could execute even with vendor keys.** The rail guards a door into a room that has not been built. This is the design for the room, and for what "out-of-band" has to mean before anything can be tested against it.

## The design, in three sentences

1. **An approval is out-of-band if and only if no combination of credentials the agent legitimately holds can produce it** — six checkable properties (separate request, separate principal, unmintable by the caller, bound-and-spent, shown verbatim, recorded independently). `X-Human-Confirmed` fails in one clause, because the header is a string the caller writes about itself.
2. **The Tier-2 approve surface** is a `login_required` approve/decline page on CareAgents rather than the engine, with **no credential in the URL**, an ownership binding from session → account → agent → action, the verbatim payload rendered rather than a model-authored summary, and the single-use confirm credential minted server-side only.
3. **Proof** is an agent-credential closure test enumerated from `app.url_map` — asserting the tested set *equals* the registered set — plus a two-context Playwright recording, run synthetic end to end against the `webhook-poster` rail.

## How it gets proven, which is the part that was missing

Two independent browser contexts. Context A holds the agent's full credential set and no browser storage; context B holds its own CareAgents login. `contextA.storageState()` is asserted to contain no cookie for the human's origin — that turns *"somewhere the agent cannot reach"* from narration into an assertion.

Refusals are asserted **before the human acts**; asserting them afterwards proves nothing. The positive is asserted first at every step (the action exists and is `awaiting_confirmation`) before any refusal. After the human taps Approve, the sink must have received **exactly one** POST whose body is **byte-equal to what the human saw on the page** — that is "shown verbatim" made testable, and it catches the failure where the human approves text A and bytes B are sent.

Synthetic tenant, `webhook-poster` rail: a real registered executor making a real outbound POST, so the execution is genuine while the side effect is not real-world. No pharmacy, no carrier, no dial.

**The phone and SMS legs stay unproven.** The recording demonstrates the gate, not Bland or Twilio. Anyone citing it as evidence that a pharmacy call works is over-reading it.

## The MCP finding — prior research had moved, and I verified it

Our 2026-08-05 research note said URL-mode elicitation via **error `-32042`** was the standardized closure path for #214. The shape is right; **the mechanism named is wrong.** Checked against the 2026-07-28 spec:

- URL-mode elicitation survives and is still correct for an out-of-band human step.
- It is now delivered inside an **`InputRequiredResult`** via **Multi Round-Trip Requests**, not an error code. `elicitationId` and `notifications/elicitation/complete` are gone; `-32042` is in the "MUST NOT emit" table.
- The client learns the outcome by **retrying the original request** with an opaque `requestState`. `action: "accept"` means the user consented to *open the URL*, not that the interaction completed — [confirmed verbatim in the spec](https://modelcontextprotocol.io/specification/2026-07-28/client/elicitation).

**The useful half: our rail is already MRTR-shaped.** Commit's 202 `awaiting_confirmation` is the `input_required` analogue, the approve URL is the elicitation, and the `GET /r6/actions/<id>` polling we already instruct is the retry. **No engine change** — the wrapper is set 6's build.

And the spec independently mandates two constraints this design reached on its own: a server **MUST NOT** provide a pre-authenticated URL, and **MUST** ensure the user who completes the flow is the user who started it. We can cite the spec rather than our own judgement.

## #216 — a split verdict, not a hedge

- **`DAILY_CAP_REACHED`: build it.** Count **claims** (transitions into `executing`), not proposals — counting proposals lets an agent exhaust a patient's cap without a single human tap, converting a safety control into a DoS lever.
- **`CONTACT_NOT_ALLOWLISTED`: recommend removal until the contact store is funded.** There is no human-attested contact store. An allowlist derived from FHIR records is not human-attested; one derived from prior actions is circular — the first call to a wrong number allowlists it. A reserved control that does nothing is the retro's defect shape.

## Filed, not fixed

**#520** — `_TRANSITIONS['awaiting_confirmation']` is `{'executing', 'expired'}`, so **a human who declines and a human who never looked produce the same record.** Verified in code. Set 4's claim rests on the record of what the human did; today that record cannot tell a refusal from a timeout, and repeated declines — the strongest signal that proposals are wrong — are invisible.

## Honest scope

**EVIDENCE PARTIAL, deliberately.** This is step 3, not step 5: none of the four artifacts exists yet, and the document specifies what will produce them. Calling it complete would over-claim in exactly the way its own risk register warns about.

One requirement is recorded as **unverified**: payload immutability after commit, which "shown verbatim" depends on. `transition_action` refuses `status` in `**fields` but does not restrict `payload_json`, and auditing every write path was outside this task's scope.

`3153 passed, 13 skipped, 1 xfailed`. `uv run ruff check .` clean. No production code changed.

### DIFF
--- a/docs/specs/2026-08-16-human-gate.md
+++ b/docs/specs/2026-08-16-human-gate.md
@@ -1,0 +1,778 @@
+# Spec — the human gate
+
+> Step 3 (architecture review) of `docs/2026-08-16-delivery-process.md`, for
+> feature set 4 (action rail). Owner: `.claude/agents/owner-action-rail.md`.
+> PRD: `docs/prd/04-action-rail.md` — this document is the §6 item recorded
+> there as *"Missing, and a SOW item: a written specification of the human
+> gate itself."*
+>
+> **Design only. No production code, no deploy, no merge.** Everything below
+> is a contract to be built against, plus the test that decides whether it
+> was built.
+
+## 0. What this document settles
+
+Three things had no written definition, so there was no test for any of them:
+
+1. **What makes a human step count.** `X-Human-Confirmed` and the approval
+   endpoint are both called "human confirmation" in our docs. Nobody had
+   written down what separates them in principle, so "is this gate real"
+   was a matter of taste. §2 makes it one sentence and one test.
+2. **What the Tier-2 approve surface is** (#215). The gates are real and
+   pinned; nothing behind them can be approved by a human, because no
+   surface exists for a human to approve on. §4.
+3. **Whether the reserved allowlist and daily-cap codes get enforcement or
+   get deleted** (#216). §6 gives a split verdict, not a hedge.
+
+It also corrects a piece of prior research that has gone stale (§5), and
+names the closure path for #214 (§7).
+
+---
+
+## 1. The four architecture-review questions
+
+### Q1. Does this serve the vision, or is it adjacent work that feels productive?
+
+It is the vision's hardest case, and `docs/2026-08-16-hard-truths.md` already
+says so:
+
+> Reading under constraint is the easy half; acting under constraint is the
+> claim nobody else is making.
+
+The product thesis is an enforcement layer that lets an agent be useful on
+real health records **without being trusted**. Every read guardrail answers
+"what may this agent see". The human gate is the only control that answers
+"what may this agent *do*", and it is the one a partner will ask about first,
+because the failure mode is a phone call placed to a real pharmacy rather
+than a record shown to the wrong person.
+
+The specific work here is not adjacent because of §6 of the hard truths: the
+gates are demonstrated and the room behind them is not built. Every demo of
+the gate implies a thing that does not exist. Building the approve surface is
+what makes the existing demo honest, and it is, per #215, *the piece entirely
+in our control* — it needs no vendor, no onboarding, and no credential we do
+not already hold.
+
+### Q2. What is the honest failure mode, and who notices it first?
+
+The failure mode is not the gate being removed. It is **the gate quietly
+becoming agent-reachable while still looking like a gate**. Four concrete
+shapes, in the order they are likely:
+
+| Shape | How it happens | Symptom |
+|---|---|---|
+| Capability URL | the approve link is built to carry a token so it can be pasted into a chat message | anyone holding the link approves; the agent generated the link |
+| Mint-secret reach | `INTERNAL_TOKEN_MINT_SECRET` becomes readable by a process that also runs model output, or is returned in a response body | the agent mints its own approval credential |
+| Session sharing | the approve surface is added to a service that already holds an agent-facing session or API credential | one credential set does both halves |
+| Summary drift | the approve page renders a model-written summary rather than the payload that executes | the human approves text A; bytes B are sent |
+
+**Who notices first, today: nobody.** There is no assertion anywhere that the
+gate is unreachable from agent credentials. `tests/test_write_guard_matrix.py`
+pins that `/confirm` refuses a *missing* step-up token (401) and that
+`/approval-token` refuses a missing internal secret (403). Neither asks the
+question this document is about, which is whether the *full* set of
+credentials an agent legitimately holds can reach execution by any route.
+
+That gap is the entire argument for §8's closure test. The first three shapes
+above are all caught by it; the fourth is caught by the byte-equality
+assertion in §8.3 step 9.
+
+### Q3. What does it make harder later?
+
+Stated plainly, because these are real costs and two of them will be
+proposed as "improvements" later:
+
+- **Headless and server-to-server deployments need their own surface.**
+  Binding approval to an authenticated human session means an integration
+  with no browser cannot approve. That is the intended consequence, not an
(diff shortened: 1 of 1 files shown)


aks129:

**Verdict: APPROVE WITH CONDITIONS ×4** (CTO, QA, Product, Dev). Nobody blocked.

Floor and ceiling both pass. This is the strongest-written document of the four, and its §4.2 approve-card contract — full destination not a label, verbatim payload not a summary, neither button a default, *"no element of the card is authored by a model"* — is the first place in this repo the human step is treated as product surface rather than plumbing.

### Two statements in the document are false, and both were proven

**H6 is presented as satisfied. It is falsifiable today.** `POST /r6/actions/<id>/review` requires only `X-Tenant-Id` + `_require_step_up(tenant)` — `Scope.WRITE`, which is inside `C_agent` — and then calls `issue_confirmation(action_id, approved_via='review-page')` at `r6/actions/review.py:445`. An agent holding only its own credentials can mint a durable record asserting a human approved via the review page. **Not an execution bypass**: `/confirm` still demands `require_grant(audience=ACTION_APPROVAL_AUDIENCE, consume_nonce=True)` (`r6/actions/routes.py:509`), unmintable without the mint secret. What fails is the evidentiary layer this set's whole claim rests on.

**R2 is marked "unverified". It is verified absent** — reproduced against the real models: `transition_action(aid, ['awaiting_confirmation'], 'executing', payload_json=<new bytes>)` returns `True`, the payload swaps, and the single `ActionEvent` records only the status change. `r6/actions/state.py:36` — `updates = dict(fields)`. Filed as #528. Note `review.py:449` mutates `payload_json` *after* the confirmation is minted at `:445`.

### §8.2 is the section the document rests on, and step 3 is unbuildable

Step 1 is the best engineering here: enumerate from `app.url_map`, assert set equality. Step 3 then says assert a refusal from *every* enumerated route — but `propose` returns 201 by design, `GET /r6/actions/<id>` returns 200, and both `/review` methods legitimately answer 200 on a generic write step-up. Followed literally the test is permanently red; narrowed to the three routes step 3 names, it becomes the guard-narrower-than-its-property defect step 1 exists to prevent (hard-truths §5, cited by this very section).

Also: the claim is worded *"moves an action out of `awaiting_confirmation`"* — `/review` doesn't move it, so the test passes while H6 fails.

### Conditions

1. **Correct H6 in §2.2** — name `review.py:445` as a live counter-example.
2. **Upgrade R2 to verified absent**, cite #528.
3. **Re-word §8.2's claim** to *"produces any artifact H1–H6 name"*, and replace step 3's blanket refusal with a **per-route disposition table** (route → method → expected status → expected state effect) covering every endpoint `app.url_map` enumerates, including the ones that must answer 200. Recenter the property on step 4. One table resolves the review, propose and callback cases together.
4. **Name the engine decline route** — method, status codes, and specifically whether it consumes an `action-approval` nonce or takes the internal secret. Those are materially different security designs and no reasonable default exists. Today `careagents/healthclaw.py` has no decline method and `r6/actions/routes.py` registers no decline route, so §4.4 is a state diagram with no way to reach it.
5. **Resolve the `approved_via` triangle** — §4.1 says "reuses that method unchanged" while §4.4 adds `'approve-page'`; `careagents/healthclaw.py:378` hardcodes `"review-page"`. Built as §4.1 instructs, the consent record names the wrong surface.
6. **Nothing says how a human learns an approval is waiting** on our own surface. §5 solves it for MCP clients via elicitation; CareAgents has no path, and the commit response's *"approve in your HealthClaw dashboard"* names a dashboard that does not exist. Specify it or exclude it in §10.
7. §7 step 1's document list misses `docs/beta-tester-guide.md`, which is published, user-facing, and claims *"Only you can approve it, through a separate step the AI's own tools cannot reach"* — false for phone/SMS/webhook kinds, where nobody can approve.
8. Extend §4.1's invariant to the payload: the verbatim card body must never land in a CareAgents log line or row. CareAgents storing no PHI is a standing non-negotiable.
9. Card contract has no phone constraints (§4.2 `:294` gives expiry as wall-clock with no timezone; `:292` says the payload "scrolls"). Run the §8.5 sign-off on a phone.
10. Low: §3 cites `routes.py:318` — use the full path; in this repo `routes.py` reads as the god module. §3 also propagates the stale "3,151-test suite".

### Verified by the chair, independently
`review.py:445`/`:449` and the `_require_step_up` write scope; `state.py:36` passing arbitrary fields into the guarded UPDATE; `routes.py:509`'s grant requirement bounding the severity; the write-gate matrix pinned at 428/401/428/201; the MCP MRTR correction against the published 2026-07-28 spec.

**Disposition: does not merge unedited.** Conditions 1–2 are truth-edits (five small changes); 3–6 are design gaps that can be *named as open items* in the spec — which is the 60–80% posture, not a failure of it.

Issue #290, opened 2026-08-03
Locked MCP endpoint is unreachable from hosted connectors: no OAuth metadata, and the 401 gives no way forward


aks129:
Split out of the fix for the keyless-quickstart drift (branch `fix/keyless-quickstart-endpoint`). That PR routes every keyless surface to the demo deployment, which unblocks trials. This issue is the thing it does **not** fix.

## What is broken

`https://mcp-server-production-5112.up.railway.app/mcp` requires a deployment-scoped bearer token. Hosted chat connectors — claude.ai, ChatGPT, Perplexity — cannot attach a static header. So the locked endpoint is not merely inconvenient there, it is unusable, and **there is currently no supported way to reach a real tenant from a hosted connector.**

The failure is also actively misleading. Observed live:

```
$ curl -i -X POST https://mcp-server-production-5112.up.railway.app/mcp ...
HTTP/2 401
www-authenticate: Bearer
{"error":"Unauthorized"}

$ curl -o /dev/null -w "%{http_code}" .../.well-known/oauth-protected-resource
404
```

Per RFC 9728 / the MCP auth spec, a `401` is a clients cue to begin OAuth. It looks for `resource_metadata` in the `WWW-Authenticate` challenge (absent), then falls back to `/.well-known/oauth-protected-resource`, `/.well-known/oauth-authorization-server`, `/.well-known/openid-configuration` — all `404`. Dynamic client registration therefore fails, and the client asks the user to supply an OAuth Client ID by hand. **No such Client ID exists**, so the user is stuck in a loop with no correct action available.

This is what a design partner hit. Their words: *"It keeps saying Couldnt register with HealthClaws sign-in service and suggesting I add an OAuth Client ID."*

Note they also reported the discovery 404s on `app.healthclaw.io`. That host is a red herring — it is the Flask app, whose SMART discovery correctly lives under `/r6/fhir/.well-known/`. The connector never talks to it.

## Options

1. **Cheapest, not a fix:** return `WWW-Authenticate: Bearer realm="...", error="invalid_token"` plus a human-readable pointer, and serve `/.well-known/oauth-protected-resource` describing the endpoint as operator-issued-token-only. Clients still cannot connect, but they stop prompting for a Client ID that does not exist, and the error tells the truth.
[...]

Issue #523, opened 2026-08-17
A conformant MCP client cannot get from our 401 to a token: the discovery chain is broken in three places


aks129:
Measured during the set-6 architecture review (`docs/specs/2026-08-16-mcp-authorization.md`). Three separate defects, one user-visible symptom: **a hosted connector that wants to authenticate has no path from our refusal to a credential.** This is the mechanism behind #290.

## 1. The authorization-server metadata is not where clients look

Three locations a conformant client MUST try were probed and returned **404 / 404 / 400**. On top of that:

- the document's `issuer` value does not match the URL it is served from — a conformant client rejects a metadata document whose `issuer` disagrees with its own location
- every endpoint URL in it is `http://`. **OAuth 2.1 requires https**, so a conformant client rejects the document on sight even if it finds it.

Any one of these three is fatal on its own. All three are present.

## 2. The 401 carries no CORS header, so a browser-context client cannot read it

The auth middleware is installed **before** the CORS middleware, so the refusal never gets `Access-Control-Allow-Origin`. A browser-based client does not merely fail to authenticate — it cannot see the challenge at all, and gets an opaque network error instead of a WWW-Authenticate it could act on.

Middleware ordering, one line.

## 3. There is no `resource_metadata` on the challenge

Even with 1 and 2 fixed, the 401 does not tell a client where to go. RFC 9728 exists for exactly this: `WWW-Authenticate` carries `resource_metadata`, and the client follows it. Today the refusal is correct and terminal.

## Why this is filed and not fixed

It is the set-6 architecture review's finding, and the review is step 3 of `docs/2026-08-16-delivery-process.md` — design before build. The spec on branch `spec/mcp-authorization` designs all three fixes together, with the deploy gate stated: **the metadata must resolve, at a location clients try, with a matching issuer and https endpoints, before anything else matters.**

Fixing them piecemeal would produce three PRs that each look complete and together still leave a client unable to connect, which is how this state arose.

## The scope boundary, stated

Complete and deployed, this reaches the **synthetic demo tenant only.** Real-tenant access is blocked on a consent surface, which is a named follow-on. Nobody should read "MCP authorization works" as "a partner can reach a patient's records."

Related: #290, #155 (deploy drift is unobservable), #522 (the dangling `mcp.healthclaw.io` record, which is also the natural canonical resource URI).

aks129:
## Severity elevation on the `http://` finding — from the review board on #524

[...]

Issue #522, opened 2026-08-17
security: mcp.healthclaw.io is a dangling record pointing at Vercel


aks129:
Found while measuring the OAuth discovery chain for the set-6 architecture review. **Verified independently over DoH** (not `dig` — this LAN forges port-53 answers).

## What

```
$ curl 'https://dns.google/resolve?name=mcp.healthclaw.io&type=A'
216.150.16.193
216.150.16.1                     ← Vercel

$ curl -sS -w '%{http_code}' https://mcp.healthclaw.io/
404
The deployment could not be found on Vercel.
DEPLOYMENT_NOT_FOUND
```

A live DNS record on a health product's domain, pointing at Vercel, claimed by no deployment.

## Why it is worth an issue rather than a shrug

**1. It is the natural canonical resource URI for the MCP server.** The set-6 authorization spec has to name a canonical `resource` value that lands in the token `aud` and in every partner's connector config. Changing it later invalidates both. `mcp.healthclaw.io` is the obvious candidate and it currently 404s — so the decision and this record are the same decision.

**2. A dangling record aimed at a hosting platform is the precondition for subdomain takeover.** Whether another account could actually claim it depends on how `healthclaw.io` is verified in your Vercel account — that is the thing to check, and it is a five-minute check. If the apex is verified and owned there, the exposure is low; if it is not, someone else adding `mcp.healthclaw.io` to their own Vercel project would be serving content on our domain, in a namespace that appears in health-product documentation.

**3. It is in the docs.** Anyone following a link to it today gets a Vercel error page, which reads as "this product is broken" rather than "this subdomain was never finished."

## What to do

- **Decide the canonical MCP resource URI** — this is a prerequisite for the authorization spec, not a follow-on.
- **Then either point the record at the real MCP deployment or remove it.** Leaving it is the only option with no upside.
- **Check the Vercel domain verification state for `healthclaw.io`** while you are there.

Related: #290 (the locked endpoint is unreachable from hosted connectors), and the open question of what Vercel is *for* — `api/index.py` is a second Flask entry point whose purpose is unsettled.

PR #524, opened 2026-08-17
docs(spec): make the locked MCP endpoint reachable without unlocking it (#290)

aks129:
**Step 3 of `docs/2026-08-16-delivery-process.md` — architecture review. Design only, no production code.**

Set 6's headline: the production MCP endpoint is **correctly token-locked and effectively unusable by the clients it exists for.** #290. A partner who tries to connect after Tuesday's webinar cannot, and the 401 gives them nowhere to go.

## The design in three sentences

1. The MCP server serves an **RFC 9728 protected-resource-metadata** document at `/.well-known/oauth-protected-resource/mcp` and at the root, and the 401 gains `resource_metadata` and `scope` so the refusal **points somewhere instead of dead-ending**.
2. The existing issuer becomes discoverable from its own issuer URL (https via `OAUTH_ISSUER`, a root metadata route, the RFC 8707 `resource` recorded as `aud`, RFC 7662 introspection), and the MCP server accepts audience-validated tokens **behind a default-off flag** — strictly additive to `MCP_AUTH_TOKEN`, which stays required at boot and stays accepted.
3. Complete and deployed, this reaches the **synthetic demo tenant only**; real-tenant access is blocked on a consent surface, which is a named follow-on spec.

That third sentence is the scope boundary, and it is deliberate. Nobody should read "MCP authorization works" as "a partner can reach a patient's records."

## The lock is not touched

`MCP_AUTH_TOKEN` stays required at boot and stays accepted. Audience-validated tokens are an *addition* behind a default-off flag, with a migration path that never leaves production unlocked, not even briefly. The founder's ruling — *"the locked is non-negotiable for real product"* — is a constraint of the design, not a thing the design negotiates with.

## How it gets proven

A `walkthrough.sh`-pattern script plus a Playwright recording produced **by** the asserting run, against the synthetic demo tenant — so no PHI is reachable **by construction**, not by precaution.

Two assertions carry the weight:

- **A4** — the authorization server's metadata resolves at a location clients actually try, with a matching `issuer` and https endpoints. **This fails today**, and it is the deploy gate.
- **R3** — a token minted at the same issuer for the *FHIR* audience, replayed at the MCP endpoint, must **401**. This is the only assertion that distinguishes "we wrote an audience check" from "the audience check works", and the design notes that its failure would otherwise be noticed by nobody.

## Three defects found while measuring — filed, not fixed

Filed as **#523**: the AS metadata is served where no conformant client looks, its `issuer` does not match its location, and every endpoint URL is `http://` (OAuth 2.1 requires https — a conformant client rejects it on sight). Separately, the auth middleware is installed **before** CORS, so the 401 carries no `Access-Control-Allow-Origin` and a browser-context client cannot read the challenge at all.

Filed as **#522**: `mcp.healthclaw.io` resolves to Vercel and returns `DEPLOYMENT_NOT_FOUND`. I re-verified this over DoH myself. It is a dangling record on a health product's domain — and it is also the natural canonical resource URI, so the decision and the record are the same decision.

Fixing these piecemeal would produce three PRs that each look complete and together still leave a client unable to connect. That is how the current state arose.

## Prior research corrected

Our 2026-08-05 note called 2026-07-28 an RC. **It is Current.** The auth mechanism is unchanged between revisions, but DCR is now deprecated (still spec-legal, still what we use) and RFC 9207 issuer identification is new. Checked rather than trusted — the same instruction produced a correction in the set-4 spec too.

## Open, and the reviewer's call

- **The canonical resource URI** — owner decision, blocks phase 1. It lands in the token `aud` and in partner configs; changing it later invalidates both. See #522.
- **`READ_AUTH_ENABLED` in the running production env** — verified in source, **not observed**. It is the single conditional between a public discovery chain and arbitrary tenant binding, so it is a phase-2 pre-deploy check.
- **What credential the MCP server presents to the introspection endpoint** — deferred to the implementation spec.
- **Whether #290 closes at phase 1 or phase 3.**

## Honest scope

**No hosted client's actual behaviour was measured** — not claude.ai, not ChatGPT, not Perplexity. The document asserts only what the spec requires *of* clients and says so explicitly; measuring real client behaviour is the end-user sign-off run. No authenticated call was made to production; this step exercised the refusal half only.

`3153 passed, 13 skipped, 1 xfailed`. `uv run ruff check .` clean. No production code, nothing deployed.

### DIFF
--- a/docs/specs/2026-08-16-mcp-authorization.md
+++ b/docs/specs/2026-08-16-mcp-authorization.md
@@ -1,0 +1,815 @@
+# MCP authorization — making the locked endpoint reachable without unlocking it
+
+**Status: proposed. Design only. No code in this branch.**
+
+| | |
+|---|---|
+| Closes | [#290](https://github.com/aks129/HealthClawGuardrails/issues/290) |
+| Feature set | 6 — Surfaces (`docs/prd/06-surfaces.md` §6 names this as the missing spec) |
+| Pipeline step | 3, architecture review (`docs/2026-08-16-delivery-process.md`) |
+| Author | owner-surfaces |
+| Decision required from | the founder, on §3.1 (the canonical resource URI) and §7 phase 3 (deploy authorization) |
+
+Everything asserted about live behaviour in this document was executed on
+2026-08-16/17 against the running deployments. The raw responses are quoted
+inline. Nothing here was inferred from reading code alone, and where a claim
+is an inference it says so.
+
+---
+
+## 1. The problem, measured
+
+The production MCP endpoint is correctly locked and unusable by the clients it
+exists for. Both halves are measured.
+
+**The lock works.** Unauthenticated `POST /mcp`, executed:
+
+```
+HTTP/2 401
+www-authenticate: Bearer
+{"error":"Unauthorized"}
+```
+
+**There is no way forward.** Every document a conformant client looks for
+after that 401, executed against the same host:
+
+```
+/.well-known/oauth-protected-resource       -> 404
+/.well-known/oauth-protected-resource/mcp   -> 404
+/.well-known/oauth-authorization-server     -> 404
+/.well-known/oauth-authorization-server/mcp -> 404
+/.well-known/openid-configuration           -> 404
+```
+
+The `WWW-Authenticate` header carries `Bearer` and nothing else — no
+`resource_metadata`, no `scope`. A client that follows the specification has
+exhausted its options at that point and falls back to asking the user for an
+OAuth Client ID that does not exist. That is the loop the design partner
+reported in #290.
+
+The failure is worse than a refusal because it is a refusal that lies about
+its own recoverability. A `401` is defined as the beginning of an
+authorization flow. Ours is the end of one.
+
+### 1.1 What is already correct and stays
+
+- `MCP_AUTH_TOKEN` is required at boot when `NODE_ENV=production` and the demo
+  flag is unset (`services/agent-orchestrator/src/index.ts`,
+  `assertMCPAuthConfigured`). The server fail-closes without it. That once
+  crash-looped production, which is the right failure.
+- The credential comparison is constant-time (`crypto.timingSafeEqual`).
+- The demo endpoint is a separate deployment, hard-pinned to a synthetic
+  tenant, and is not touched by this design.
+- Expired sessions return `404` so clients re-initialize (#490).
+
+**Production MCP stays token-locked.** This design adds a second credential
+type. It removes nothing. At no point in §7 does an unauthenticated call
+succeed.
+
+---
+
+## 2. Prior research, verified rather than trusted
+
+The note carried in `docs/prd/06-surfaces.md` §6 is dated 2026-08-05. Checked
+against the specifications on 2026-08-16:
+
+| Claim in the note | Verdict |
+|---|---|
+| RFC 9728 `/.well-known/oauth-protected-resource` + audience-validated tokens is the smallest fix keeping the existing issuer | **Holds.** MCP servers **MUST** implement RFC 9728; clients **MUST** use it for authorization-server discovery. |
+| "MCP spec 2025-11-25 is stable; the 2026-07-28 RC…" | **Stale.** 2026-07-28 is now **Current**, not a release candidate. 2025-11-25 is a past revision. |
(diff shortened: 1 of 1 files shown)


aks129:

**Verdict: APPROVE WITH CONDITIONS ×4** (CTO, QA, Product, Dev). Nobody blocked. **Nothing inside the document is false** — every defect it names was confirmed against the running production systems, not just source.

### The lock holds — traced, not assumed

The CTO traced all three phases. Phase 1's two well-known routes need **no middleware exemption and no reordering**, because `isMCPTransportPath` (`services/agent-orchestrator/src/index.ts:66-76`) protects only `/mcp`, `/sse`, `/messages`. That is the usual way this class of change opens a hole and it is structurally absent. Phase 2 is Flask-only and cannot admit anyone at the MCP server — a non-static bearer is refused at `index.ts:109` before `extractHeaders` runs. Phase 3 is behind a default-off flag; rollback leaves the static path untouched. **The founder's non-negotiable is satisfied at every phase.**

### Confirmed live by QA

- `https://app.healthclaw.io/.well-known/oauth-authorization-server` → **404**; the three RFC 8414 path-component locations → 404 / 404 / 400, exactly as §3.3 claims
- The document lives at the fourth location, returns 200 with `"issuer":"http://app.healthclaw.io"` and **every endpoint `http://`**, served over HTTPS. Mechanism: no `ProxyFix` anywhere, so `request.host_url` ignores `X-Forwarded-Proto`
- `POST /mcp` with `Origin: https://claude.ai` → `HTTP/2 401`, `www-authenticate: Bearer`, **no `access-control-allow-origin`** — a browser client cannot read the challenge

### Conditions

1. **Elevate the `http://` severity in #523.** The document frames it as "a conformant client rejects it on sight." A *non*-conformant one **downgrades**: it sends an OAuth authorize redirect over plaintext because our HTTPS-served metadata told it to.
2. **Add `APP_ENV` to the phase-2 pre-deploy check.** §10 names `READ_AUTH_ENABLED` as "verified in source, not observed" — but its boot guard (`r6/runtime_config.py:105`) sits *behind* `if app_env != "production": return app_env` at `:102`. If `APP_ENV` is unset or misspelled the guard never fires and read-auth can be off with the app healthy. The set-2 evidence pack records a runbook running `APP_ENV=development`. Observe both.
3. **Name `ProxyFix` (or `OAUTH_ISSUER`) as the phase-1 item that fixes the scheme.** `request.host_url` is the root cause of both the scheme and the issuer mismatch; §3.3's three changes should say which one addresses it.
4. **Re-scope the §7 invariant to name both origins**, or re-title it "the MCP endpoint is never unlocked." As written the heading claims a blast radius the invariant does not cover — phase 2 makes a pre-existing open-DCR + auto-approve chain (`r6/oauth.py:279-291`, whose own comment says it "AUTO-APPROVES with no consent screen") discoverable and https. It does not open it; it advertises it.
5. **Bind the quickstart edit into the phase-1 PR.** §7 names the copy risk and assigns it to `docs/quickstarts/claude.md` — where `:83-88` currently tells readers the production endpoint "asks for a Client ID you don't have… re-add it with the demo URL." Phase 1 makes that troubleshooting text wrong the hour it ships.
6. **Record Tuesday's answer.** Nothing says whether any phase lands before 18 August. The honest position: #290 does not close before the webinar, `mcp.healthclaw.io` still returns `DEPLOYMENT_NOT_FOUND` (#522), and the URL given to a room of integrators is the demo endpoint with the §6.1 boundary said out loud.
7. **Do not close #290 without its successor filed.** Phase 1 closing it is right *if* the consent spec is filed the same day with §6.1 as its problem statement.
8. **Phase 3 is not buildable as written** — the credential the MCP server presents to introspection is unspecified (DCR client? `X-Internal-Secret`? new shared secret?), each with a different rotation story across two separately-deployed services. Phases 1–2 may proceed.
9. Minor: `<RESOURCE_ROOT>` is used in §3.4 and never defined; §3.3's "all in `r6/oauth.py`" understates change 2, which needs app-level registration (`r6/routes.py:79,82`); §9.4 states the CORS requirement but not the mechanism, and the existing middleware denies by default.

### Chair's note
`mcp.healthclaw.io` re-verified over DoH — Vercel IPs, `DEPLOYMENT_NOT_FOUND`. It is also the natural canonical resource URI, so the dangling record and the URI decision are one decision.

**Disposition: clean to merge.** Every condition is a follow-on, now durable here.

Issue #525, opened 2026-08-17
$conformance grades other people's servers against our own parameter list — no external implementation can score A


aks129:
Found while writing the guardrail specification (#234). **Verified in code before filing** — this is the item that decides whether #234 produces a standard or a document.

## What

`r6/conformance/probes.py:939` — the corrective-rejection check for an unknown search parameter:

```python
if (parameter_lower not in declared
        and rejection is not None
        and declared == _LOCAL_SUPPORTED_PARAMETER_EVIDENCE):
    return True
```

**Set equality**, against:

```python
_LOCAL_SUPPORTED_PARAMETER_EVIDENCE = {
    "patient", "code", "status", "_lastupdated",
    "_count", "_sort", "_summary", "context-id",
}
```

`context-id` is a HealthClaw parameter. No other FHIR server has it.

## The consequence

A third-party server that refuses an unknown search parameter **perfectly correctively** — names the bad parameter, returns an OperationOutcome, declares its own supported set — grades **C** on this check, because its set is not byte-identical to ours. C caps error fidelity, and error fidelity caps the deployment at **B**.

So: **no external implementation can score A.** The grade is not measuring conformance to a guardrail contract. It is measuring resemblance to our implementation.

## Why this matters more than its size

Three things we are actively doing rest on this endpoint meaning something to someone else:

1. **The Aidbox example**, submitted upstream, invites a reader to run `$conformance` against their own server.
2. **The tester program** (`docs/2026-08-16-tester-program.md`) proposes "a conformance badge for their own server" as the main carrot for developer-testers. They cannot earn the top grade.
3. **#234 itself** — a versioned guardrail specification other implementers build to. A spec whose reference test can only be passed by the reference implementation is not a spec.

It is also the pattern named in `docs/2026-08-16-hard-truths.md` §4: the most polished artifact making the worst claim. `$conformance` is the flagship, and it grades everyone else's server against our parameter list.

## The decision, which is why this is filed rather than fixed

What *should* the requirement be? Options, and this is a CTO-gate call:

- **Declare a set, any set.** The property is "the refusal tells the caller which parameters ARE supported" — the specific parameters are the server's business. Check that a supported-set is declared and non-empty, not that it equals ours.
- **Declare a set that includes the parameter families the spec requires.** Stricter, still implementable by others.
- **Keep set equality and rename the grade** so it does not claim to be a conformance measure.

The first looks right and I have not made the call.

[...]

Issue #498, opened 2026-08-14
Error fidelity degrades in upstream proxy mode: unknown search parameters are forwarded, not refused


aks129:
Found while running the Aidbox example (#496) end to end against a live Aidbox for the first time.

## What happens

With `FHIR_UPSTREAM_URL` set, the guardrail conformance harness scores **B (6/7)**. The failing property is **Error Fidelity**, and every one of its failures has the same shape: a request that the guardrail should refuse itself is instead forwarded upstream, and the caller gets the upstream's answer.

```
[FAIL] Error Fidelity — F (local-fhir-only)
  ✗ strict unknown parameter is rejected — grade C; status 404
  ✗ lenient unknown parameter carries an outcome warning — grade C; status 404
  ✗ lenient warning is audited truthfully — grade F
  ✗ unsupported modifier is rejected — grade C; statuses 502,502,502
  ✓ strict rejection is audited as a failure — grade A
  ✓ unsupported modifier rejections are audited as failures — grade A
```

In local mode the same deployment scores A. The difference is entirely the proxy path.

## Why it matters more than the grade

Error fidelity is graded on *what the caller is told when something goes wrong*. A 404 or a 502 from Aidbox tells an agent nothing about which parameter it got wrong, so the agent cannot correct itself — which is the whole point of the property. In proxy mode part of that answer is currently the upstream server's rather than ours, and the guardrail layer is where the contract was promised.

The 502s on the modifier probe are worth a second look on their own: three of them in a row suggests the request is failing upstream rather than being answered.

## Where

The strict/lenient parameter handling that produces the OperationOutcome lives in the local search path. The proxy path in `r6/fhir_proxy.py` forwards the query without applying it.

## Scope note

Not fixed in #496. That PR corrects the example and makes the walkthrough assert honestly — every property except error fidelity must pass, so this gap is stated rather than graded away, and the same assertion passes at 7/7 unchanged once this is closed.

## Related

A neighbouring defect in the same area **is** fixed in #496: the harness built clinical Observations referencing `Patient/conformance-subject`, which nothing creates. SQLite validates no references so it passed locally; Aidbox refused it, and the report blamed the human-in-the-loop gate — "the gate blocks confirmed writes too, so its 428s prove nothing" — when the gate was fine and the reference was dangling. Probes now create a subject first, which took the live grade from D (4/7) to B (6/7).

Issue #401, opened 2026-08-05
Conformance cannot grade read auth, and that is the gap a skeptical adopter finds first


aks129:
Split out of #213 while making five of its seven properties two-sided (PR #400). Filed because it is a **missing property**, not a one-sided one, and it needs design rather than another probe.

## What is ungraded

`READ_AUTH_ENABLED` is required in production — pinned by `tests/test_runtime_security_config.py`, and stated in `README.md` and `SECURITY.md`. It **defaults off**, and it is off in the conformance test fixture, where `conformance-selftest` is not in `PUBLIC_TENANTS`.

So the conformance harness has no read-auth property at all. A deployment that serves every tenant's records to an unauthenticated caller scores exactly what a correctly-gated one scores.

## Why it cannot just be added

Adding a probe would fail `test_our_deployment_records_full_conformance` and drop our own CI grade — not because our deployment is wrong, but because the harness cannot tell "read auth is off because this is a public demo tenant" from "read auth is off because someone disabled it."

That is the actual design problem: **the harness needs a way to know what a deployment claims before it can grade whether the claim holds.** Every other property is unconditional; this one is conditional on a posture the deployment declares.

## Why it matters more than the count suggests

PR #400 proved that five of seven properties were previously satisfiable by a deployment doing nothing — including one where forged step-up tokens authorized clinical writes, which still scored Grade A. Read auth is the same class of hole, still open: the grade is the product claim, and a partner evaluating us reads it as covering "can a stranger read patient records."

It is also the first thing a skeptical adopter checks, and the first thing they will notice we do not measure.

## Shape of a fix

Something like a declared posture in the conformance request or the deployment's capability statement — "this deployment claims read auth" — with the probe grading the claim, and a deployment that claims nothing scoring an explicit *unclaimed* rather than a silent pass. That keeps the public demo honest (it claims no read auth, and is graded accordingly) without letting a production deployment omit the claim to dodge the check.

Do **not** fix this by adding a probe that passes when read auth is off. That reproduces exactly the defect #213 was about.

## Also still ungraded, from the same review

The action rail's propose/commit/confirm separation; step-up negatives beyond forged-token (expired, cross-tenant, read-scope-on-write, replayed nonce — all four verified correct by hand, none graded); redaction on `$lastn` and `SubscriptionTopic/$list`; rate limiting; mint-gate fail-closed; purge.

Related: #213, #214, PR #400.

aks129:
[...]

Issue #382, opened 2026-08-05
r6/brief/ reads resources unredacted, and is on neither #282 probe list


aks129:
Found while auditing the redaction boundary after #365 and #369. This site is on **neither** `tests/test_redaction_coverage_inventory.py` nor `tests/test_redaction_probes_multistep.py`, because it was not on #282's original list of eight.

## What happens

`r6/brief/routes.py:28-37` `_resources_for` returns `r.resource` straight from `R6Resource` with no `apply_redaction`. There is no `apply_redaction` call anywhere in `r6/brief/`.

`r6/brief/engine.py:41-50` `_code_text` then reads `code.text` first, then `coding[].display` — the two fields CLAUDE.md names explicitly because real feeds put patient names in them — into the `AppointmentBrief` extension the patient and their clinic receive.

## The trap in fixing it

`_code_text` falls back to the literal string `"Unknown"` (`engine.py:50`).

Add `apply_redaction` here without changing anything else and **every brief field becomes "Unknown"** — because redaction strips the upstream text and the brief never asks the terminology table for a server-derived label. That converts a #282-shaped leak into a #376-shaped hole: a document full of "Unknown" handed to a clinician, with no indication whether that means "no data" or "we stripped it".

So this is not a one-line fix. It needs the redact-then-relabel pair that `r6/routes.py` and (since #365) `r6/labs/routes.py` both use — strip the feed's text, then put back a label keyed by code from `r6/terminology.py`.

## Order of work

1. Add a marker probe for the brief path to the multistep inventory first, so the leak is measured rather than assumed, and so the fix has something that goes green.
2. Then apply redact-then-relabel, and assert both halves: no upstream marker survives, AND a recognised code still produces a readable label. Asserting only the first is what let #376 hide — see the positive-check pair added to `probe_phi_redaction`.

## Severity

The brief is generated for a real appointment and delivered outside the redacted FHIR read path. Treat as the same class as #365, not as a cleanup.

Related: #282, #365, #369, #376.

aks129:
**Measured. FAIL — twelve fields leak, and the "Unknown" question is answered.**

Nineteen probes on `test/brief-redaction-probe`, at the HTTP boundary, deserialized with the consumer's own parser rather than a re-implementation. Eleven production mutations applied, run and reverted, including a no-op control (M11) that stayed green — so the harness is not painting everything red.

## Leaking (12)

[...]

PR #526, opened 2026-08-17
docs(spec): a versioned guardrail specification, with the gaps named (#234)

aks129:
**Step 3 of `docs/2026-08-16-delivery-process.md` — architecture review. Design only, no production code, no probe weakened, no pin touched.**

Set 1's PRD §6 said it: *"Missing, and a SOW item: a versioned, published guardrail specification with a threat model. Today the contract is seven properties in a test file."* This is `0.1.0-draft` of that specification.

## Normative requirement vs. implementation detail, per property

The seven properties were the starting point, not the answer. Some of what the test file asserts is *our implementation*, not a requirement anyone else could meet:

| Property | Normative | Implementation detail the tests pin |
|---|---|---|
| P1 redaction | identifiers absent, resource still returned, upstream free text stripped, labels re-derived from codes | the **transformation** — initial+period, last-four, birthDate year. Explicitly non-normative, so improving it is not a breaking change |
| P2 audit | every access emits a retrievable record; read distinguishable from write; detail PHI-free | `AuditEvent` at `/AuditEvent`, `action=R`, the search shape |
| P3 step-up | the gate discriminates issued from forged; refusal 401 **or** 403 | exactly-401 — **a conforming 403 fails our probe** |
| P4 human gate | out-of-band confirmation by a second party — **aspirational, not met** | that a client-set header flips 428→201. A discrimination floor, not the property |
| P5 tenant isolation | cross-tenant read returns nothing; owner still served | 404-vs-403; existence non-disclosure is a SHOULD the probe does not check |
| P6 disclaimer | a machine-readable notice attached to the payload | substring test — a labelling requirement holding 1/7 of the grade |
| P7 error fidelity | refusals corrective, echo no supplied text, audited as failures, ignored params declared | the A-grade ladder, and the supported-parameter set — **see #525** |

P4 is the largest divergence and is marked as aspirational twice.

## The three sharpest entries in "what these guardrails do NOT stop"

1. **An agent with a write token approves its own clinical writes** (#214). `X-Human-Confirmed` is caller-set, and the 428 *names the header it wants* — an agent that reads the error message satisfies the gate on its next request.
2. **Grade A covers `/r6/fhir`; data leaves through other routes** (#282, #382). Labs, care gaps, quality, actions, form-fill, SDC and SMBP never call `apply_redaction` — and `r6/brief/` reads `CodeableConcept.text` and `coding[].display` into a document a patient and their clinic receive.
3. **A deployment serving records to anonymous callers scores identically** (#401). `READ_AUTH_ENABLED` defaults off and no property grades it.

Writing that section with the same care as the rest is the point. Our honest "Grade B, and here is why" has been worth more to partners than any green tick.

## Filed from this review: #525

`_outcome_names_parameter_and_supported_set` requires **set equality** with our own eight search parameters — `context-id` included, which no other FHIR server has. A third-party server refusing a bad parameter perfectly correctively grades **C**, which caps error fidelity, which caps the deployment at **B**.

**No external implementation can score A.** The grade measures resemblance to our implementation, not conformance to a contract — and the Aidbox example invites readers to run it against their own server, while the tester program offers a conformance badge as the main carrot for developer-testers.

That is the decision this spec surfaces and does not make: what *should* the requirement be. It is a CTO-gate call and #525 lays out three options.

## Also in this branch: the language guard now opens `docs/specs/`

`tests/test_deidentification_language.py` globs `docs/blog/` and `docs/recipes/`. This branch creates `docs/specs/` — **the most public surface we have**, written for other implementers to build to — and the guard could not see it.

Same shape as the gap it was widened to close earlier today: the needle became the claim rather than four phrases, and the *scan* stayed as wide as its file list. Proven both ways: an unhedged "HIPAA Safe Harbor de-identification" sentence appended to the spec goes **red** with the glob and **green** without it.

## Honest scope

**Set 1 remains EVIDENCE PARTIAL.** A spec document is not a proxy run. Proxy mode is still unmeasured, and the recording and two sign-offs are still outstanding.

Stated in the document: proxy mode was never run by its author and Grade B is cited from #498 throughout, labelled as not measured here. The spec has had **no review** — no CTO pass, no QA pass — and carries an explicit clause saying any part may change without a version bump at `0.1.0-draft`. One requirement (audit detail carries no PHI beyond the one SSN token) rests on the standing invariant rather than a test citation.

Three further gaps are recorded in the document rather than filed, deliberately: no probe attempts to modify or delete an AuditEvent yet "Immutable Audit Trail" prints on every passing scorecard; 6/7 is Grade B whether the miss is tenant isolation or a disclaimer.

`3153 passed, 13 skipped, 1 xfailed`. `uv run ruff check .` clean.

### DIFF
--- a/docs/specs/guardrail-spec-0.1.0-draft.md
+++ b/docs/specs/guardrail-spec-0.1.0-draft.md
@@ -1,0 +1,1094 @@
+# HealthClaw Guardrail Specification
+
+**Version 0.1.0-draft** · 2026-08-16 · Issue [#234](https://github.com/aks129/HealthClawGuardrails/issues/234)
+
+**Status: draft. Nothing in this document has been agreed with an external
+party, and no deployment has yet been certified against it.**
+
+This is the normative contract for an enforcement layer that sits between an AI
+agent and a FHIR record server: reads come back redacted, every access is
+audited, writes need a credential the agent cannot mint, clinical writes need a
+second party, and one tenant cannot see another's data.
+
+It exists because a third party adopting HealthClaw today has no document to
+conform to beyond reading our source. The de facto contract is seven properties
+in a test file (`tests/test_guardrail_conformance.py`) and the probes behind
+them (`r6/conformance/probes.py`). This document promotes that contract to a
+stated one, and — more importantly — states where the test file is narrower than
+the property it is named after.
+
+## The rule this document was written under
+
+> No claim here may be stronger than what `tests/test_guardrail_conformance.py`
+> actually verifies. Where the requirement exceeds the suite, it is marked
+> **aspirational** and the gap is named.
+
+Every property section therefore has two halves that are allowed to disagree in
+public: the **normative requirement** an implementer builds to, and **what
+0.1.0's suite actually checks**. Where they diverge, the divergence is written
+down. A specification that hid that would join a list this project keeps
+internally of its own polished artifacts making claims the code did not
+support — a flagship conformance report that graded F against a real server, a
+security policy that claimed a legal de-identification standard, a QA script
+reporting "7/8 checks passed" against a server that was not running. In every
+case the polish is what suppressed inspection.
+
+---
+
+## 1. Conformance model
+
+### 1.1 Terminology
+
+The key words MUST, MUST NOT, SHOULD, SHOULD NOT, and MAY are to be interpreted
+as described in RFC 2119 and RFC 8174, when and only when they appear in
+capitals.
+
+- **Implementation** — the guardrail layer under test.
+- **Agent** — a non-human caller holding a credential, acting on a user's
+  behalf. The primary subject of this specification.
+- **Tenant** — the isolation boundary. Every request carries exactly one.
+- **Upstream** — a FHIR server the implementation proxies to, when configured.
+- **Graded surface** — the request paths this suite probes. In HealthClaw that
+  is the FHIR route family under `/r6/fhir`. See §6.2: it is not the whole read
+  surface, and that matters more than anything else in this document.
+
+### 1.2 The system under test
+
+An implementation conforms at a **deployment**, not at a commit. Configuration
+is part of what is graded: a deployment with read authentication disabled and a
+deployment with it enabled are different systems under test, and this version of
+the suite cannot tell them apart (§4, [#401](https://github.com/aks129/HealthClawGuardrails/issues/401)).
+
+The suite probes over HTTP with synthetic data only. It **writes**, and it does
+not clean up. An implementation being graded MUST designate a self-test tenant
+for it; pointing the suite at a demo or production tenant leaves probe records
+behind, which has happened here and was found on camera the day before a launch
+recording ([#463](https://github.com/aks129/HealthClawGuardrails/issues/463)).
+
+### 1.3 Profiles
+
+A property MAY be measured over more than one profile. Three are defined:
+
+| Profile | What it exercises | Default |
+|---|---|---|
+| `local` | the implementation answering from its own store | run |
+| `proxy` | the implementation in front of a configured upstream | `not_run` |
+| `mcp` | the implementation's MCP tool surface | `not_run` |
+
+A property graded over one profile says nothing about the others. A report MUST
+declare which profiles ran (§3, R1). HealthClaw's own `error_fidelity` grade
+carries `coverage=local-fhir-only` with `proxy: not_run` for exactly this
+reason.
+
+### 1.4 Grading
+
+The overall grade is a pure fraction of properties passed:
+
+```
+7/7 -> A    6/7 -> B    5/7 -> C    4/7 -> D    <=3/7 -> F
+```
+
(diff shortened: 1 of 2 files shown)


aks129:

**Verdict: APPROVE (QA, unconditional) + APPROVE WITH CONDITIONS ×3.** The strongest artifact of the four, and the only one that drew an unconditional approval.

### It survived a hostile audit

QA walked all seven property rows against `r6/conformance/probes.py` and could not falsify one:

- P3's *"a conforming 403 fails our probe"* is exact — `probes.py:743-744` and `755-756` both assert `status == 401`
- P5's divergence row is exact — `probes.py:812-815` passes on `status != 200 or returned_id != pid`, so 403 passes and no status is pinned
- Appendix A's counts reconcile: P1 **17** (3+5+2+5+2), P2 3, P3 3, P4 2, P5 2, P6 2, P7 6 = **35**, which is the "35 checks" figure PRD 01 quotes

Dev independently tried to break the traceability table and failed: P5 and P6 look like undercounts at 2 until you notice the third `Check(` sits on a failure early-return branch, so a passing run emits exactly 2.

Every external citation resolves. The non-docs diff against the merge base is **exactly the one glob addition** — no probe weakened, no pin raised, nothing removed. Appendix A cites its source commit, which is the precision the topology lacked.

### #525 has a second mechanism, and #525 as filed does not name it

Beyond `declared == _LOCAL_SUPPORTED_PARAMETER_EVIDENCE` at `probes.py:941`, there is a short-circuit at `:913` calling `_outcome_has_unsafe_last_updated_suggestion`. Its `strip_exact_supported_set` helper (`:884-891`) removes the declared-parameters sentence before scanning for `_lastupdated` **only when the set exactly equals ours**. `_lastupdated` is in our set, so the strip exists purely to let *our* declaration through.

Any third-party server that truthfully lists `_lastUpdated` among its supported parameters — nearly every FHIR server — keeps the sentence, trips the unsafe-suggestion heuristic on its own legitimate declaration, and is forced to C by a **second, independent path**. A fix that relaxes set equality alone will look complete and change nothing.

### CTO ruling on #525 — option 1, with a mandatory second half

The normative property is *"the refusal tells the caller which search parameters are supported."* Which parameters those are is the implementing server's business. Option 3 (keep set equality, rename the grade) preserves a broken measure and spends credibility relabelling it. Option 2 needs a US Core-anchored family list `0.1.0-draft` does not have.

Requirement becomes: **a corrective rejection names the offending parameter, and declares a non-empty well-formed supported-parameter set that does not contain it.** No comparison to our set anywhere.

**Mandatory second half:** `strip_exact_supported_set` must be generalized in the *same* PR to strip any well-formed declaration before the `_lastupdated` scan, keeping the flag for the genuine unsafe case (`_lastUpdated` offered as a *substitute*, with no declaration preamble). Both mechanisms move together, two-way mutation on each.

### Conditions

1. **Record the `probes.py:913`/`:884` mechanism in #525** before anyone fixes it. (Chair has posted it there.)
2. **The widened glob is not pinned by the test that exists to pin scan width.** `test_the_scan_cannot_be_narrowed_without_saying_so` asserts `len(published) >= 5`; `docs/blog/*.md` (1) + `docs/recipes/*.md` (4) = exactly 5. Delete `docs/specs/*.md` tomorrow and the suite stays green. **That is this spec's own G-C shape reproduced inside the fix for G-C.** Raise the floor or assert the specs directory by name.
3. `docs/specs/*.md` is flat — a spec in a subdirectory is unguarded. `**/*.md` while the file is open, for all three globs.
4. The glob was written from the file in front of it, not from the property: #519 landed 14 public documents in `docs/prd/`, `docs/evidence/` and four top-level `docs/2026-08-16-*.md`, none scanned. Widening to `docs/**/*.md` needs a run, not a blind change — `hard-truths.md:58` deliberately quotes the bad Safe Harbor claim and would need an explicit exemption.
5. §8.1 `:947-948` — *"the scorecard is theirs to publish"* — is corrected by §8.2 four lines later, but §8.1 is where an integrator stops. Put the forward reference in place so the offer and the cap are never separated.
6. Carry §7.2's G-B ("Immutable Audit Trail" is not tested for immutability) to the owner as a naming-or-probe decision. It prints on every passing scorecard.
7. Nothing links to `docs/specs/` — no index, no README entry, no pointer from the scorecard. A specification nobody can find is a document, not a standard.
8. The PR **body's** compressed table is looser than the spec in two rows (P1 reads as though the tests pin the transformation; P5's "404-vs-403" reads as a probe pin). Appendix B repeats neither error — the document is right, the summary is not.

### Chair's verification
`_outcome_names_parameter_and_supported_set`'s set equality and the second mechanism at `:913`/`:884`, both read in code. The widened glob tested against a tree containing #521's and #524's spec files: green, and a two-way mutation (unhedged Safe Harbor sentence into #521's spec) goes red — so it scans them rather than passing vacuously. Merge order is free in every direction.

**Disposition: clean to merge**, and worth merging *first* so the other two specs are not sitting unguarded on the most public surface the repo has.

Issue #535, opened 2026-08-31
A mistyped tenant returns 401 'requires authentication', sending you to hunt for a credential instead of a typo


aks129:
A tenant id that does not exist is answered as an authentication problem:

```
$ curl 'https://app.healthclaw.io/r6/fhir/Patient/$care-gaps?patient=demo-marisol' \
       -H 'X-Tenant-Id: desktop-dmo'
HTTP 401
{"issue":[{"code":"security","diagnostics":"Read access to tenant 'desktop-dmo' requires authentication",...}]}
```

`desktop-dmo` is a typo for `desktop-demo`. There is no such tenant. The message says the tenant requires authentication, which is true of the general rule and false about this request — nothing could authenticate to a tenant that does not exist.

In the care-gaps browser app this surfaces as **"care-gaps returned HTTP 401 — this tenant requires a step-up token"**, which sends the reader looking for a credential.

## Why this is the shape we keep filing

It is the family in `docs/2026-08-02-retro.md` and the CareAgents fixes: *"a reason that blames the record"*, *"could not look it up is not 'the source sent free text'"*. The system reports the reason it checked rather than the reason that applies.

Found by a tester who mistyped the tenant in the browser app while walking the demo. On camera, that message costs the take.

## The tension worth deciding, not assuming

Distinguishing "no such tenant" from "exists but unauthorised" is **existence disclosure**, and the guardrail spec's P5 treats non-disclosure as a SHOULD. So the fix is not simply "say it does not exist" — that would leak which tenant ids are real to an unauthenticated caller.

Options:
1. Keep the response identical (correct for the isolation property) and improve only the **client-side** copy in the care-gaps app: "401 — check the tenant id is spelled correctly, or this tenant needs a credential."
2. Distinguish server-side only for tenants marked public, where existence is already disclosed.

Option 1 is probably right and is a one-line change. Recording the reasoning so the next person does not "fix" it into a disclosure.

aks129:
Fixed by #541 (`6484e0b`), option 1: the server answer stays identical for a tenant that does not exist and one that is not yours — by design, so a probe cannot learn which — and the care-gaps app's copy names both causes: *check the tenant id is spelled correctly, or this tenant needs a credential.* The reasoning is in the comment beside the copy (`templates/mcp_apps/care_gaps.html`) so the next person does not fix it into a disclosure. Closing; the issue outlived its fix.

PR #541, opened 2026-09-02
fix(mcp-apps): care-gaps app rendered "0 DUE" over an unevaluated patient (#538, #535)

### LINKED ISSUE #538: The care-gaps MCP App drops the API's own honesty note and renders '0 DUE' over an unevaluated patient
The `$care-gaps` endpoint returns an explicit note saying nothing was evaluated. The MCP App that renders it throws the note away and shows a large **0 DUE** tile.

## Reproduce

```sh
curl -s -X POST 'https://app.healthclaw.io/r6/fhir/Patient/$care-gaps' \
  -H 'X-Tenant-Id: desktop-demo' -H 'Content-Type: application/json' -d '{}'
```

The payload carries:

> `unevaluated_note`: *"More than one patient record is connected here… Nothing was examined, and that is not a finding that you have no screenings outstanding."*

Now open <https://app.healthclaw.io/r6/fhir/mcp-apps/care-gaps/>, type `desktop-demo`, click Load. The page renders **0 Due / 0 Up to date / 1 Not applicable / 6 Needs data**, six cards reading *"indeterminate — date of harm unknown — cannot determine eligibility"*, and **none of the note**.

## Why this is the shape we keep filing

This is precisely *"a crashed screening review is not 'no screenings due'"* (#381) and *"a reason that blames the record, and a partial list sold as whole"* (#417) — both fixed in the engine, and both re-appearing here because **the UI does not render the honesty field the engine produces.**

The engine was made careful and the surface discards the care. A patient reading this page is told zero screenings are due when the truth is that nothing could be evaluated.

Found while a tester walked the app as a non-technical user preparing a demo recording — it was proposed as the browser-friendly way to show the product, and it turned out to demonstrate the opposite.

## Two smaller defects in the same page

1. **A typo reads as an auth failure.** Entering `desktop-dmo` renders *"care-gaps returned HTTP 401 — this tenant requires a step-up token."* Step-up is the **write** gate; this is a read. The API says `Read access to tenant 'desktop-dmo' requires authentication`. See #535 — the client copy is the cheap half of that fix.
2. **Enter does not submit** the tenant field; only the Load button works.

aks129:
Candidate for feature set 6 (surfaces — MCP Apps). Closes #538; client half of #535.

## What was wrong

Verified live 2026-08-31: a tenant holding more than one Patient gets `unevaluated: "ambiguous-patient"` from `$care-gaps` and a note saying nothing was examined. The MCP App page threw the note away and drew a large **0 Due** tile over six cards reading "date of birth unknown" — every one an artefact of the call, not a fact about the person.

Cause: the note was rendered only inside a box gated on `consumer.lines`, which is empty on exactly the path that has a note to show.

## What changed

- The route injects the engine's own caller-reason family (`caller_reasons()` in `r6/caregaps/report.py`) so the page has one source of truth and never keeps a copy of the list.
- On a **caller** reason (`no-patient`, `ambiguous-patient`, `check-incomplete`): tiles show `—` (labelled for assistive tech), cards give way to "No screenings were evaluated.", the box is headed **Not evaluated** with the engine's note.
- On an **engine** reason (e.g. `sex-unavailable`): counts and cards stay; the caveat is drawn above the list it qualifies.
- 401 copy no longer names the write gate. It says to check the tenant id spelling or that the tenant needs a credential (#535, client half).
- Enter in the tenant box submits, like the sibling pages.

## Evidence

- `tests/test_mcp_app_care_gaps.py` — 8 source-shape guards, comments stripped before matching (the `test_mcp_app_lab_trends.py` pattern). Mutation-verified: restoring the `lines ?` ternary goes red in both the Python guard and the Playwright spec.
- `e2e/tests/care-gaps-app.spec.ts` — asserts the fixture at the API first (so a fixture change reads as a fixture change), then watches the render paint against the seeded demo tenant; second test intercepts a production-shaped 401 for a misspelled tenant and checks the copy.
- Local: 39 passed (care-gaps app + lab-trends + report tests), 2/2 Playwright on `E2E_PORT=5099`, `uv run ruff check .` clean.

## What was NOT run / remaining 20–40%

- Not exercised through a real MCP host (`_meta.ui.resourceUri` embedding). The Playwright spec loads the page directly.
- Not run against a tenant with real records. Synthetic demo tenant only.
- Route-layer half is **not** in this PR and is filed separately: `$care-gaps` still calls the evaluator with `patient=None` when `not_evaluated` is set, so `detail` carries false `not_applicable` rows and the audit detail records `evaluated=7`. The page now hides those rows on the caller path, but the payload still carries them.

### DIFF
--- a/e2e/tests/care-gaps-app.spec.ts
+++ b/e2e/tests/care-gaps-app.spec.ts
@@ -1,0 +1,102 @@
+/**
+ * Care Gaps MCP App — the page must not say "0 Due" over a patient it never
+ * looked at (#538), plus the client half of #535.
+ *
+ * Verified live 2026-08-31: a tenant with more than one Patient gets
+ * `unevaluated: "ambiguous-patient"` from $care-gaps and a note that nothing
+ * was examined. The page dropped the note and drew a "0 Due" tile over six
+ * cards saying "date of birth unknown" — artefacts of the call, not facts
+ * about the person. The Python guards (tests/test_mcp_app_care_gaps.py) pin
+ * the source's shape; this is the only place the render is watched paint.
+ *
+ * What it proves:
+ *   1. the seeded demo tenant really does answer ambiguous-patient — asserted
+ *      against the API first, so a fixture change reads as a fixture change
+ *      and not as a page regression
+ *   2. the page shows the engine's own note, no "0 Due", and no per-rule cards
+ *   3. Enter in the tenant box submits, and a 401 is explained as a spelling
+ *      or credential problem, not as a missing step-up token
+ *
+ * Requires the demo history: `flask --app main seed-demo-history`, which the
+ * e2e webServer runs at boot.
+ */
+import { test, expect } from '@playwright/test';
+
+const TENANT = 'desktop-demo';
+const headers = { 'X-Tenant-Id': TENANT };
+
+function param(parameters: any, name: string) {
+  const p = (parameters.parameter || []).find((x: any) => x.name === name);
+  return p ? JSON.parse(p.valueString) : null;
+}
+
+test.describe('care-gaps MCP App', () => {
+  test('an unevaluated tenant shows the engine note, not "0 Due"', async ({ page, request }) => {
+    // --- 1. The fixture, checked at the API --------------------------------
+    const res = await request.post('/r6/fhir/Patient/$care-gaps', {
+      headers: { ...headers, 'Content-Type': 'application/json' }, data: {},
+    });
+    expect(res.ok()).toBeTruthy();
+    const consumer = param(await res.json(), 'consumerSummary');
+    expect(consumer.unevaluated, 'fixture: desktop-demo must hold more than one Patient')
+      .toBe('ambiguous-patient');
+    expect(consumer.unevaluated_note).toContain('Nothing was examined');
+
+    // --- 2. The page -------------------------------------------------------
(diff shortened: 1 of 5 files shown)


aks129:
**Dedicated QA review (2026-09-02): FAIL → fixed in 9a941ed.**

1. HIGH — `r6/routes.py` grew 3924 → 3931 against the 3927 pin (`tests/test_ratchets.py::test_the_god_module_only_shrinks`); CI `python-tests` was red and Playwright never ran. The "39 passed" in the original body did not include the ratchet. Fixed: the 5-line docstring became a one-line comment; routes.py is exactly 3927.
2. LOW — the refactored `tile()` interpolated the count bare (`${n || 0}`); driving the page with a hostile `summary.due` executed in headless Chromium. Source is engine ints, but the line was touched and the escape test is named for it. Fixed: `esc(n || 0)`, pinned in `test_every_interpolated_string_is_still_escaped`.

Verified by the reviewer: comment stripper sound; mutation A (restore the `lines ?` ternary) → 1 failed 7 passed; `tojson` escapes `<`/`'`; no circular import; no hand copy of the reason list in the template; PHI: reasons are engine keys only.

After the fixes: full suite `3165 passed, 13 skipped, 1 xfailed`; `uv run ruff check .` clean; Playwright `2 passed` on 5099.

`dependency-audit` is red here for three transitive npm advisories published after main's last audit — PR #544 fixes that; it is not this PR.

Issue #312, opened 2026-08-03
Unauthenticated RecursionError on FHIR write paths — body parsed before the auth gate


aks129:
Found by the write-guard matrix. Reproduced in `tests/test_write_guard_matrix.py::test_deep_nesting_is_refused_only_on_the_path_that_was_patched`. This is #267's crash lever, still live on the main write paths.

## What happens

`create_resource` parses the body **before** its step-up gate — [r6/routes.py:454](r6/routes.py#L454) `request.get_json(silent=True)` runs before the token check at [:464-469](r6/routes.py#L464-L469). `silent=True` suppresses JSON decode errors but **not `RecursionError`**, which `json.loads` raises on a deeply nested payload (~1000 deep, Python's recursion limit).

The guard (`_json_depth_within` + `except RecursionError`) exists **only** on `/internal/ingest-bundle` ([r6/routes.py:2363-2382](r6/routes.py#L2363-L2382)). The following write paths parse before their gate with no such guard:
- `POST /r6/fhir/<type>` (create)
- `PUT /r6/fhir/<type>/<id>` (update)
- `POST /r6/fhir/Bundle/$ingest-context`
- `POST /r6/actions/propose`
- `POST /r6/smbp/enroll`

## Impact

A deeply nested JSON body — reachable with **only a tenant header**, no step-up token — raises an unhandled `RecursionError` and 500s/crashes the worker. Cheap unauthenticated DoS on the primary FHIR write surface. create and update also disagree with each other on parse/gate ordering.

## Fix

Apply the ingest-bundle pattern (depth check + `except RecursionError` → 400) to these paths, ideally by gating auth before the parse, or by a shared body-reader that bounds depth. Candidate Phase-0 PR.

PR #545, opened 2026-09-02
candidate: pin the kernel's uniform step-up token strip (#334)

### LINKED ISSUE #334: Decide once: the access kernel strips the step-up token, widening 6 sites during slices 4-8
Surfaced by slice 3 (#333). **Decide here so slices 4–8 don't repeat it six times.**

## The change

`require_grant` strips the token (`r6/access.py:286`). Five of the six remaining migration sites read the header **without** stripping today, so migrating them will accept a valid token sent with surrounding whitespace where they currently return 401.

Confirmed empirically on the migrated SMBP route: a padded valid token passes the gate; garbage still 401.

**Do not strip today:**
`r6/routes.py:475` (create) · `:640` (update) · `:3032` (curatr apply-fix) · `:3427` (share-bundle) · `r6/actions/review.py:63` · `r6/agent_runs/routes.py:52`

**Already strips:** `r6/read_auth.py:86`.

## Why this is a widening and not a hole

The HMAC must still verify for the tenant. A whitespace-padded valid token *is* the valid token — stripping admits nobody who wasn't already admitted. It is authority-neutral.

It is still a behaviour change on a security gate, and the migration protocol says a behaviour change is a failure rather than something to resolve in passing. Hence this issue rather than a silent repeat.

## Recommendation: keep the strip

1. **Not stripping is worse.** `read_auth.py` strips today. If the kernel didn't, slice 8 would *narrow* that site and start rejecting tokens that currently work — a regression shipped as a refactor.
2. **One rule beats seven.** The codebase currently has two behaviours for the same header. Uniform stripping ends that. A per-site `strip=` flag would reintroduce exactly the per-site variation this refactor exists to remove — the shape behind the four step-up status dialects.
3. **The failure mode it fixes is real.** A token copy-pasted with a trailing space currently yields an opaque 401 on five write paths.

## The alternative, stated fairly

Make the kernel not strip, and hand-roll a stripping read at `read_auth`'s call site to preserve it. That keeps every current behaviour exactly, at the cost of one special case in the kernel's only tolerant caller — and it means the kernel's token read is not the single token read after all.

## What I need

A ruling: **strip uniformly** (recommended) or **preserve per-site behaviour**. If the former, slices 4–8 proceed and each PR states the widening in its body. If the latter, `r6/access.py` needs the change before slice 4.

Related: #333 (slice 3), `docs/2026-08-03-access-kernel-spec.md` §2.5, `docs/2026-08-03-refactor-working-protocol.md`.

aks129:
## What

Records the #334 ruling in the three places a reader would look, with no behaviour change:

- `tests/test_access_kernel.py` — two pin tests. `test_a_padded_token_is_admitted_and_padded_garbage_is_still_refused`: a valid token padded with `" \t"` on both sides yields a Grant; the same padding around garbage raises `StepUpDenied` with `checked=True`. `test_the_tenant_header_is_not_stripped_even_though_the_token_is`: the same padding on `X-Tenant-Id` is rejected as `malformed`.
- `r6/access.py` — one comment line at the strip in `_step_up_token`, citing #334.
- `docs/2026-08-03-access-kernel-spec.md` §2.5 — a "Ruled, #334" paragraph after the 2026-08-16 update, before §2.5b.

Verified the first test goes red under its named mutation (drop the `.strip()` at `r6/access.py:353`) and green with it.

## Why

Council D5, unanimous: "Strip uniformly, ASCII whitespace only. The kernel already does and six merged slices depend on it, so the ruling is retroactive and there is no behaviour change. Close #334 with one pin test, a one-line comment at the strip citing #334, and a §2.5 note. The tenant header stays unstripped: it is an identifier compared against a stored value, not an authority-neutral token."

The strip is authority-neutral: the HMAC still has to verify for the tenant, so a padded valid token is the valid token and stripping admits nobody who was not already admitted.

**One thing to note.** The shipped strip is `str.strip()` with no argument, which is Unicode-wide (it also removes e.g. U+00A0, U+2003), whereas the ruling names ASCII whitespace only. This PR pins what ships today, because six slices depend on it; narrowing to `.strip(" \t\r\n\f\v")` is a separate decision and is deliberately not made here. The spec note says so.

## Ran

```
uv run ruff check .
All checks passed!

3159 passed, 13 skipped, 1 xfailed, 2 warnings in 108.23s (0:01:48)
```

Ratchets untouched: `r6/routes.py` not modified; nothing new imports `r6.routes`; no new raw tenant reads, direct step-up calls, or soft-delete-blind files.

## What was NOT run

- No live server. The change is a test, a comment and a doc paragraph; there is no runtime path to exercise beyond the unit tests.
- The Postgres CI lane runs on this PR, not locally.

Closes #334

### DIFF
--- a/docs/2026-08-03-access-kernel-spec.md
+++ b/docs/2026-08-03-access-kernel-spec.md
@@ -620,6 +620,20 @@ characterization test so they can become gates, because they should not become
 gates. §1.2b adds `has_grant`, which is the same decision returned rather than
 raised, and they adopt *that*. Slices 15–18 below.
 
+**Ruled, #334 (council D5, unanimous).** The kernel strips the step-up token
+uniformly, at every source (`_step_up_token`, `r6/access.py:353-369`). Five of
+the six pre-migration sites did not strip and `r6/read_auth.py` did; one rule
+replaces both. It is authority-neutral — the HMAC still has to verify for the
+tenant, so a padded valid token is the valid token and stripping admits nobody
+who was not already admitted. Slices 3–11 landed on this behaviour before the
+ruling was recorded, so it is retroactive with no behaviour change; the pin is
+`test_a_padded_token_is_admitted_and_padded_garbage_is_still_refused`. Two
+edges deliberately outside the ruling: the strip is `str.strip()` with no
+argument, which is Unicode-wide rather than the ASCII-only the ruling names —
+narrowing it is a separate decision, not a migration PR; and the **tenant
+header stays unstripped** (§1.1, `_validated`), because it is an identifier
+compared against a stored value, not an authority-neutral token.
+
 ### 2.5b Slices 15–18 — `has_grant`, the four predicates
 
 Landed first as a pure addition adopted by nothing (§1.2b). Then, one site per
--- a/r6/access.py
+++ b/r6/access.py
@@ -349,6 +349,7 @@ def _step_up_token(*, also_bearer: bool, also_body_field: str | None) -> str:
     is opt-in for the same reason TenantSource is: a reader must be able to
     see which inputs an endpoint trusts without reading this helper.
     """
+    # Stripped uniformly at every source: a padded valid token is valid (#334).
     token = (request.headers.get('X-Step-Up-Token') or '').strip()
     if token:
         return token
--- a/tests/test_access_kernel.py
+++ b/tests/test_access_kernel.py
@@ -323,6 +323,47 @@ def test_a_read_scoped_token_does_buy_a_tenant_bound_grant(app, tenant_id):
     assert grant.scope is Scope.TENANT_BOUND
 
 
+def test_a_padded_token_is_admitted_and_padded_garbage_is_still_refused(
+        app, tenant_id):
+    """#334 (council D5): the kernel strips the step-up token uniformly.
+
+    A whitespace-padded valid token IS the valid token — the HMAC still has
+    to verify for this tenant, so stripping admits nobody who was not already
+    admitted. Six merged slices depend on this, so the ruling is retroactive
+    and this test pins the shipped behaviour rather than changing it. Note
+    the strip is `str.strip()` with no argument, which is Unicode-wide;
+    narrowing it to ASCII whitespace is a separate decision.
+
+    MUTATION: drop the .strip() in _step_up_token -> the first half red.
+    """
+    tenant = Tenant(id=tenant_id, source=TenantSource.HEADER)
+    padded = ' \t' + generate_step_up_token(tenant_id) + ' \t'
+    with app.test_request_context(headers=_headers(padded)):
+        grant = require_grant(scope=Scope.WRITE, tenant=tenant)
+    assert grant.tenant_id == tenant_id
+
+    with app.test_request_context(headers=_headers(' \tnot-a-token \t')):
+        with pytest.raises(StepUpDenied) as exc:
+            require_grant(scope=Scope.WRITE, tenant=tenant)
+    assert exc.value.checked is True
+
+
+def test_the_tenant_header_is_not_stripped_even_though_the_token_is(app,
+                                                                    tenant_id):
+    """#334, the other half: the strip is for the token only.
+
+    The tenant id is an identifier compared against a stored value, not an
+    authority-neutral token, so the padding that the token gate forgives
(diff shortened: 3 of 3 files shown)


aks129:
## QA review — **PASS**

Reviewed at `22620d6` in an isolated worktree. `autoMergeRequest: null` (checked, not armed).

### Spec compliance (council D5)

| D5 asks for | In the diff | Verified |
|---|---|---|
| one pin test: padded valid token admitted, padded garbage refused | `test_a_padded_token_is_admitted_and_padded_garbage_is_still_refused` | mutation red, see matrix |
| one-line comment at the strip citing #334 | `r6/access.py:352` | `git diff origin/main -- r6/` is exactly `1 insertion(+)`, a comment |
| §2.5 note in the kernel spec | `docs/2026-08-03-access-kernel-spec.md` after the 2026-08-16 update | line refs `353-369` re-checked against the file: accurate |
| tenant header stays unstripped | `test_the_tenant_header_is_not_stripped_even_though_the_token_is` | mutation red |
| no behaviour change | zero non-comment production change | yes |

**Spec note honesty.** The note does *not* claim ASCII-only stripping; it says the strip is `str.strip()` (Unicode-wide) and that narrowing is a separate decision. I confirmed the Unicode claim empirically rather than take it from the body: a valid token padded with U+00A0, U+2003 or U+3000 on both sides is **admitted** by `require_grant`. So the PR pins what ships. The disagreement is between the ruling text and the code, not between the PR and either.

### Mutation matrix (each applied by hand, run, reverted; `git diff --stat origin/main -- r6/` back to `1 insertion(+)` after each)

| Mutation | Command | Result |
|---|---|---|
| drop `.strip()` at `r6/access.py:353` | `uv run python -m pytest tests/test_access_kernel.py -q -p no:cacheprovider -k padded` | `1 failed` — `r6.access.StepUpDenied: Invalid token signature` on the *padded valid* token, i.e. red for the stated reason |
| `.fullmatch(value.strip())` in `_validated`, `r6/access.py:200` | `... -k "not_stripped or malformed_tenant"` | `2 failed, 4 passed` — the new test **and** the pre-existing `' leading'` row both go red |

### Findings

**LOW — the second test partly duplicates an existing pin.** `tests/test_access_kernel.py:93-108` already carries `' leading'` in `test_a_malformed_tenant_is_rejected_whatever_the_source`, with a docstring saying the same thing. D5 asked for "one pin test". The new test adds the trailing-and-tab variant and a #334 cross-reference, which is a fair reason to keep it, but it is more than the ruling asked for. Owner's call; no change needed for merge.

**LOW — the ruling's premise is wrong, and this PR is where that gets recorded.** D5 says "ASCII whitespace only. The kernel already does". The kernel does not; it strips Unicode whitespace (three probes above). The PR handles this the right way (pin what ships, disclose the gap in the spec), but PR #543's D5 text should be corrected or the discrepancy accepted there in writing, so the two documents do not disagree once both merge.

**Out of scope, pre-existing, worth its own issue (MEDIUM).** While probing the strip I found that a non-ASCII byte in the *signature* half of a step-up token raises `TypeError: comparing strings with non-ASCII characters is not supported` from `hmac.compare_digest` at `r6/stepup.py:224`, and nothing catches it: `POST /r6/fhir/Observation` with `X-Step-Up-Token: <valid>\xe9` (or `<valid>​`) returns **500**, where ASCII garbage returns 401. Fails closed (no bypass), but it is an unauthenticated crash lever on the auth gate — the #312 shape — and reachable from every `require_grant` site plus the bearer and body-field sources. `r6/stepup.py` is untouched by this PR and the behaviour is identical on `origin/main`, so this is not a finding against #545. Repro (Flask test client, `TESTING=False`): valid → 201, `not.a-token` → 401, `tok + '\xe9'` → 500, `tok + '​'` → 500.

### Gates (on this branch)

```
uv run ruff check .
All checks passed!

3159 passed, 13 skipped, 1 xfailed, 2 warnings in 108.12s (0:01:48)
```

### Not covered

- No live HTTP server. Note that a real HTTP parser strips ASCII OWS from header values before the app sees them (RFC 7230 §3.2.4), so the header-source half of this pin is a kernel property, not an observable wire behaviour; the bearer and body-field sources are where the strip is observable on the wire, and they are not separately pinned here (the ruling did not ask for that).
- Postgres lane: CI only.

PR #546, opened 2026-09-02
candidate: refuse non-string resource ids on ingest, pin the four id cases (#286)

### LINKED ISSUE #286: ingest: a blank resource id is now refused rather than auto-assigned — decide which is correct

aks129:
## What

`r6/fasten/ingester.py` `_ingest_one`: one type guard before the existing `str()` coercion. An `int` id still coerces to a string; any other non-string id (`bool`, `float`, `list`, `dict`/object) is refused as `invalid_id`. `bool` is a subclass of `int`, so it is checked explicitly before the int branch. The docstring's `invalid_id` line says so.

`tests/test_ingest_resilience.py`: four pins in the file's existing `_ingest_one` style, each driving the real ingester entry point and checking the row table afterwards.

| id | outcome | status |
|---|---|---|
| `""` | `invalid_id`, nothing stored | already the case, now pinned |
| absent | `ok`, a well-formed UUID, row stored | already the case, now pinned |
| `123` | `ok`, stored as `"123"` | already the case, now pinned |
| `True`, `False`, `1.5`, `{"a": 1}`, `[1]` | `invalid_id`, nothing stored | **new** for `True`/`False`/`1.5`; list/dict never passed the pattern |

The fourth test was red before the guard (`True` → `"True"` passed `_RESOURCE_ID_PATTERN`) and is green after. `invalid_id` stays in `REFUSED_OUTCOMES` (asserted in the blank-id test), so each refusal pages rather than counting as a skip.

## Why

Council D8: "Refuse a blank id as `invalid_id` (already the case); an absent id gets a UUID (already the case); an integer id keeps coercing to a string (pre-existing, live on the Fasten connector); any other non-string id (bool, float, object, list) is `invalid_id`, since `str()` of a bool or float passes the id pattern today. Pin all four."

The blank-id refusal is the right one: the pre-#267 behaviour fabricated a UUID for `""`, which turned every re-ingest of the same resource into a duplicate row instead of an upsert on `(tenant, type, id)`.

## Ran

```
uv run ruff check .
All checks passed!

3161 passed, 13 skipped, 1 xfailed, 2 warnings in 105.00s (0:01:45)
```

Ratchets untouched: `r6/routes.py` not modified; nothing new imports `r6.routes`; no new raw tenant reads, direct step-up calls, or soft-delete-blind files.

## What was NOT run

- No live Fasten or SHC import. The guard sits inside `_ingest_one`, which every path shares, and the tests drive that function directly on SQLite; the Postgres CI lane runs on this PR, not locally.
- The `0` int case is not separately pinned; it rides the int-coercion pin (`123` → `"123"`) and stores as `"0"`, as the issue table says.

Closes #286

### DIFF
--- a/r6/fasten/ingester.py
+++ b/r6/fasten/ingester.py
@@ -375,7 +375,8 @@ def _ingest_one(resource: dict, tenant_id: str,
 
     Returns ('ok', resource_id) on success, ('skipped', None) for unsupported
     types, ('forbidden', None) for a type excluded by `allowed_types`, or
-    ('invalid_id', None) for a caller-supplied id outside the FHIR id shape.
+    ('invalid_id', None) for a caller-supplied id outside the FHIR id shape
+    or not a string (an int is coerced; a bool, float, list or object is not).
     Raises on unexpected DB errors.
 
     `agent_id` and `detail` parameterize the audit event so a caller other
@@ -400,6 +401,12 @@ def _ingest_one(resource: dict, tenant_id: str,
 
     resource_id = resource.get('id')
     if resource_id is not None:
+        # An int keeps coercing to a string — pre-existing, and live on the
+        # Fasten connector. Any other non-string (bool, float, list, object)
+        # is refused BEFORE str(), because str(True) and str(1.5) both pass
+        # the id pattern. bool is a subclass of int, hence checked first (#286).
+        if isinstance(resource_id, bool) or not isinstance(resource_id, (str, int)):
+            return 'invalid_id', None
         resource_id = str(resource_id)
         if not _RESOURCE_ID_PATTERN.fullmatch(resource_id):
             return 'invalid_id', None
--- a/tests/test_ingest_resilience.py
+++ b/tests/test_ingest_resilience.py
@@ -244,6 +244,36 @@ def test_any_other_non_string_resource_id_is_refused(client, tenant_id):
         tenant_id=tenant_id, resource_type="Observation").count() == 0
 
 
+def test_an_explicit_json_null_id_mints_a_uuid_rather_than_being_refused(
+        client, tenant_id):
+    """The fifth falsy id, which D8's four cases do not name: `"id": null`.
+
+    `resource.get('id')` cannot tell an absent key from a present `null`,
+    so a null id takes the absent branch and is MINTED, while `""` — the
+    other empty id — is refused. That divergence is the behaviour today and
+    this pins it as-is; it is not an endorsement. If null should instead be
+    refused (the argument: a UUID for an id the feed did send, but sent
+    empty, turns the next re-ingest into a duplicate row — the same
+    append-not-upsert shape #286 exists to stop), that is a product call,
+    and this test is what makes changing it deliberate rather than silent.
+
+    MUTATION: `if resource_id is not None:` -> `if resource_id:` makes the
+    blank-id test red while this one stays green, which is precisely the
+    asymmetry being recorded.
+    """
+    from r6.fasten.ingester import _ingest_one
+
+    result, rid = _ingest_one(
+        {"resourceType": "Observation", "id": None, "status": "final"},
+        tenant_id)
+
+    assert result == "ok"
+    assert uuid.UUID(rid)  # minted, not "" and not the string "None"
+    assert R6Resource.query.filter_by(
+        tenant_id=tenant_id, resource_type="Observation", id=rid
+    ).first() is not None
+
+
 # ---------------------------------------------------------------------------
 # #293 / #306: a record we REFUSED is not a record we chose to skip, and the
 # SHC path never got the rollback the Fasten path was patched for on
--- a/tests/test_ingest_resilience.py
+++ b/tests/test_ingest_resilience.py
@@ -170,6 +170,81 @@ def test_resource_id_pattern_is_a_charset_control_not_a_length_control():
 
 
 # ---------------------------------------------------------------------------
+# #286 (council D8): what `_ingest_one` does with an id that is not a
+# well-formed string. Four cases, pinned in the shape the ruling gave them.
+# `invalid_id` is in REFUSED_OUTCOMES, so each refusal pages rather than
+# counting as a skip; keep it there.
+# ---------------------------------------------------------------------------
+def test_a_blank_resource_id_is_refused_as_invalid_id(client, tenant_id):
+    """`""` is not an absent id. It used to earn a fabricated UUID, which
+    turned every re-ingest of the same resource into a duplicate row
+    instead of an upsert on (tenant, type, id). Refused, and refused as a
+    refusal (#286).
+    """
+    from r6.fasten.ingester import REFUSED_OUTCOMES, _ingest_one
+
+    result, rid = _ingest_one(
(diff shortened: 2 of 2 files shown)


aks129:
## QA review — **PASS**

Reviewed at `81dd9a6` in an isolated worktree. `autoMergeRequest: null` (checked, not armed).

### Spec compliance (council D8)

| D8 asks for | Diff | Verified how |
|---|---|---|
| blank id → `invalid_id` (already the case) | pinned | mutation D below |
| absent id → UUID (already the case) | pinned | `uuid.UUID(rid)` asserted, row stored |
| int id keeps coercing to a string | pinned, untouched | mutation C below |
| any other non-string (bool, float, object, list) → `invalid_id` | `r6/fasten/ingester.py:408` | mutations A/B below |
| pin all four | four tests | all four go red under a mutation |

Guard placement checked directly, not from the body: it sits inside `if resource_id is not None:` and **before** `str()`, so `None` still reaches the UUID branch (`ingester.py:403-414`). `bool` is tested before the `(str, int)` membership, which is what makes `True` refusable given `bool` ⊂ `int`. `invalid_id` was already in `REFUSED_OUTCOMES` (`ingester.py:61`) and the blank-id test asserts it, so a refusal still increments `refused` and calls `log_refusal`, not `skipped`.

**No accounting change.** All three `_ingest_one` callers (`r6/fasten/ingester.py:233`, `r6/shc/routes.py:267`, `r6/routes.py:2526`) already branch on `invalid_id`/`REFUSED_OUTCOMES`; `git diff origin/main -- r6/` touches no counter line. **No new PHI surface**: a refusal returns before `add_audit_event`, and `log_refusal` still logs only tenant, `resourceType` and reason — never the offending id, which is the point of that function.

### Mutation matrix (each applied by hand, run, reverted; `git diff --stat origin/main -- r6/` back to `8 insertions(+), 1 deletion(-)` after each)

Command for all four: `uv run python -m pytest tests/test_ingest_resilience.py -q -p no:cacheprovider -k resource_id`

| # | Mutation | Result |
|---|---|---|
| A | delete both guard lines (`ingester.py:408-409`) | `1 failed` — `assert ('ok', 'True') == ('invalid_id', None)` |
| B | drop only the `isinstance(resource_id, bool)` clause | `1 failed` — same assertion; the bool-first ordering is load-bearing and pinned |
| C | tighten to `if not isinstance(resource_id, str):` | `1 failed` — the `123` → `"123"` pin catches the narrowing |
| D | regress `if resource_id is not None:` → `if resource_id:` (pre-#267 shape) | `2 failed` — `assert ('ok', '05359…') == ('invalid_id', None)`, i.e. the blank id earns a fabricated UUID again |

Baseline unmutated: `7 passed, 13 deselected`.

### Exercised beyond the unit tests

Drove the **real route** `POST /r6/fhir/internal/ingest-bundle` (Flask client, `INTERNAL_TOKEN_MINT_SECRET` set) with one Bundle carrying five ids — `123`, `True`, `1.5`, `""`, absent:

```
200 {"entries": 5, "ingested": 2, "skipped": 3, "failed": 0,
     "errors": [{"index":1,"code":"invalid_resource_id",...},
                {"index":2,...},{"index":3,...}]}
GET /r6/fhir/Observation/123 -> 200, json.id = '123'
```

So the refusal reaches the caller as a per-entry `invalid_resource_id` with no id echoed in the message, and the int coercion is readable end to end.

### Findings — all LOW, none blocking

1. **LOW, as-asked: `0` and `-1` behave sanely.** Probed directly: `0` → `('ok', '0')`, `-1` → `('ok', '-1')`. Both match `_RESOURCE_ID_PATTERN` (`-` and digits are in the charset) and both are valid FHIR ids under `[A-Za-z0-9\-\.]{1,64}`. `10**300` → `invalid_id` (301 chars, over the 255 cap) and `1e3` → `invalid_id` (float, caught by the new guard — note that *before* this PR `str(1000.0)` = `"1000.0"` passed the pattern, so the guard genuinely closes that one). Stating, not fixing, per the brief.
2. **LOW — an explicit JSON `null` id mints a UUID and is not pinned.** `{"resourceType":"Observation","id":null}` → `('ok', <uuid>)`, because `resource.get('id')` is `None` and takes the absent branch. Defensible (JSON `null` is "no id"), pre-existing, and outside D8's four cases — but it is the one falsy id the four pins do not cover, and `""` and `null` now diverge. Worth a line in the docstring or a fifth pin if the owner wants the set closed.
3. **LOW, pre-existing — the coerced id is not written back into the payload.** For `id: 7` the row is `row.id='7'` while `resource_json` still holds `"id": 7` (an int, invalid FHIR). The single-resource read path re-emits the row id (the `GET` above returned the string `'123'`), so this is internal today; anything that serves `resource_json` verbatim would emit a JSON number where FHIR requires a string. Untouched by this PR — the ruling said keep the coercion — but it is the loose end behind the coercion and belongs in #286's close comment or its own issue.

### Gates (on this branch)

```
uv run ruff check .
All checks passed!

3161 passed, 13 skipped, 1 xfailed, 2 warnings in 113.50s (0:01:53)

uv run python -m pytest tests/test_ratchets.py -q -p no:cacheprovider
13 passed in 1.43s
```

### Not covered

- No live Fasten or SHC job. I drove `_ingest_one` directly and through the internal ingest-bundle route; the Fasten (`ingester.py:233`) and SHC (`shc/routes.py:267`) loops were read, not executed, so the `refused`-vs-`skipped` counters are verified by inspection on those two paths and by execution only on the direct-upload path.
- Postgres lane: CI only. SQLite here.

Issue #509, opened 2026-08-16
Two soft-delete defects the #422 sweep found, in the opposite direction from #422


aks129:
Found by the sweep #422 asked for. Both are about `is_deleted`, and neither is the #422 shape — these are cases where the code and its own stated intent disagree.

## 1. `form_fill.py` says it fails loud on a deleted record, and does not

`r6/actions/rails/form_fill.py:59-66`:

```python
# (3) Load the reviewed QR, tenant-scoped. If it vanished (deleted or
# stale), the reviewed answers are gone: fail loud, never render a
# blank/guessed form.
row = R6Resource.query.filter_by(
    resource_type='QuestionnaireResponse', id=reviewed_qr_id,
    tenant_id=action.tenant_id).first()
if row is None:
    return ExecutionResult(status='failed', error=errors.STALE_SOURCE_DATA)
```

The comment names deletion as the case it handles. The query does not filter `is_deleted`, so a soft-deleted QuestionnaireResponse is loaded and **rendered into a form that gets submitted on a patient's behalf** — the exact "blank/guessed form" outcome the comment says it prevents.

This is the retro's defect shape with a comment on it: a control that reads as one thing and does another, and the comment is what makes it read that way. It is also on the action rail, which is the one place where "nothing executes without a human" has to be demonstrable.

## 2. `context_builder` re-ingest leaves a tombstoned row invisible

`r6/context_builder.py:86-97` upserts:

```python
existing = R6Resource.query.filter_by(
    id=resource_id, resource_type=resource_type, tenant_id=tenant_id).first()
if existing:
    existing.update_resource(resource_json)
```

It must see tombstones (otherwise it inserts a duplicate composite PK), so the missing filter is correct here. What is missing is the other half: it never clears the flag. `r6/fasten/ingester.py:422` does exactly that, on the same operation:

```python
existing.is_deleted = False
```

So re-ingesting a resource whose id carries a tombstone writes the new content into a row that stays invisible to every read path. The data is there, the ingest reports success, and nothing can read it.

This is the **opposite** failure from #422 — there, deleted rows were read; here, live rows are written into a grave — and it is the more confusing one to debug, because the write succeeds.

## Why one issue

[...]

Issue #306, opened 2026-08-03
SHC ingest: no per-entry rollback and logs the raw exception


aks129:
Found by the architecture audit (PR #303, finding S-4). **Verified against current `main`.**

## What happens

[r6/shc/routes.py:255-257](r6/shc/routes.py#L255-L257):

```python
except Exception as exc:
    failed += 1
    logger.warning('SHC ingest error (job=%s): %s', job_id, exc)
```

Two defects in three lines:

1. **No rollback.** A failed entry leaves the SQLAlchemy session poisoned, so every subsequent entry in the batch fails too. This is the exact defect the Fasten path was patched for on 2026-07-08 — that fix ([r6/fasten/ingester.py:140-153](r6/fasten/ingester.py#L140-L153)) carries a comment documenting the incident. The SHC path never got it.
2. **Logs the raw exception.** `%s` on the exception can embed record content from the payload being ingested. Every other ingest path deliberately logs `type(exc).__name__` instead — compare [ingester.py:152-153](r6/fasten/ingester.py#L152-L153) and [routes.py:2445-2448](r6/routes.py#L2445-L2448). `.github/REVIEW_STANDARDS.md` rule 1 requires this.

## Context: three strategies for one callee

All three callers of `_ingest_one` handle failure differently:

| Caller | Strategy |
|---|---|
| `internal/ingest-bundle` | SAVEPOINT per entry, continue on failure ✅ |
| Fasten `stream_ingest` | full rollback per bad line ✅ |
| **SHC `_ingest_bundle`** | **nothing** ❌ |

## Suggested fix

Adopt the SAVEPOINT-per-entry pattern and log `type(exc).__name__`. The audit proposes extracting a shared `ingest_entries()` so this cannot diverge a fourth time; Product has asked for the patch now and the extraction later.

aks129:
S-3 ($ingest-context flag-coupled gate) triaged **down from HIGH to MED** by the CTO design pass: `r6/runtime_config.py:73-74` raises at startup if `READ_AUTH_ENABLED` is off in production (called from `create_app`, main.py:310), so the app **cannot boot** in prod with this gate open. Verified live: prod returns 401 on an unauthenticated $ingest-context. Real exposure is dev/docker-compose/non-prod only. The genuine defect is design — it is the only write whose authorization is coupled to a *read*-auth flag — and that is what the eventual fix should decouple.

aks129:
[...]

PR #547, opened 2026-09-02
candidate: revive a tombstoned row on context-builder re-ingest (#509 defect 2)

aks129:
## What

`r6/context_builder.py`, in `ingest_bundle`'s upsert branch: one line, `existing.is_deleted = False`, with a two-line comment citing the issue. The select stays unfiltered on `is_deleted` — it has to find the tombstone or it would insert a duplicate composite PK.

`tests/test_context_builder.py`: `test_reingest_revives_a_tombstoned_row`, in the file's existing class and route-driven style. It seeds an Observation row at `is_deleted=True` with `status: preliminary`, POSTs the standard `sample_bundle` to `/r6/fhir/Bundle/$ingest-context`, and asserts the row is `is_deleted is False`, `version_id == 2`, and its JSON now says `status: final`. Red before the fix (`assert True is False` on `is_deleted`), green after.

## Why

Council D13, unanimous "today": "`existing.is_deleted = False` in `r6/context_builder.py` where the upsert lands on a tombstone, matching what the Fasten ingester already does (`r6/fasten/ingester.py:422`)."

This is the opposite failure from #422: not a deleted row being read, but a live write landing in a grave. The ingest reported success, the data was there, and no read path could see it — the more confusing of the two to debug.

**Can #509 close?** Yes, once this merges. Defect 1 (`form_fill.py` loading a soft-deleted QuestionnaireResponse) was fixed in `b265cfa`, which is an ancestor of `origin/main` (`r6/actions/rails/form_fill.py:63-72` carries the `is_deleted=False` filter and a comment naming the issue). The maintainer's comment on #509 kept it open on defect 2 alone; this PR is defect 2. I used `Refs` rather than `Closes` so the close is a human call after review.

## Ran

```
uv run ruff check .
All checks passed!

3158 passed, 13 skipped, 1 xfailed, 2 warnings in 105.33s (0:01:45)
```

Ratchets untouched: `r6/routes.py` not modified; nothing new imports `r6.routes`; no new raw tenant reads, direct step-up calls, or soft-delete-blind files (the test file imports `R6Resource` to seed and read a row directly, which is the same thing `tests/test_ingest_resilience.py` already does).

## What was NOT run

- No live server. The route path is exercised through the Flask test client on SQLite; the Postgres CI lane runs on this PR, not locally.
- Not reachable in production yet, per the issue: `is_deleted = True` is written on one line today (the demo walkthrough, on Permission rows) and no route accepts DELETE. This fixes the writer before the delete path ships.

Refs #509

### DIFF
--- a/r6/context_builder.py
+++ b/r6/context_builder.py
@@ -90,6 +90,9 @@ class ContextBuilder:
 
                 if existing:
                     existing.update_resource(resource_json)
+                    # A re-ingest of a tombstoned (tenant, type, id) revives
+                    # it, as r6/fasten/ingester.py does (#509 defect 2).
+                    existing.is_deleted = False
                     r6_resource = existing
                 else:
                     r6_resource = R6Resource(
--- a/tests/test_context_builder.py
+++ b/tests/test_context_builder.py
@@ -4,6 +4,9 @@ Tests for the Context Builder service.
 
 import json
 
+from models import db
+from r6.models import R6Resource
+
 
 class TestContextBuilder:
     """Test context builder functionality."""
@@ -77,3 +80,35 @@ class TestContextBuilder:
         data = resp.get_json()
         assert 'expires_at' in data
         assert data['expires_at'] is not None
+
+    def test_reingest_revives_a_tombstoned_row(self, client, sample_bundle,
+                                               sample_observation, tenant_id,
+                                               tenant_headers):
+        """#509 defect 2: the upsert lands on a soft-deleted row and must
+        clear the flag, as the Fasten ingester does. Before the fix it wrote
+        the new content into the tombstone and reported success — data that
+        was there, and that no read path could see.
+        """
+        stale = dict(sample_observation, status='preliminary')
+        row = R6Resource(
+            resource_type='Observation',
+            resource_json=json.dumps(stale, separators=(',', ':'), sort_keys=True),
+            resource_id=sample_observation['id'],
+            tenant_id=tenant_id,
+        )
+        row.is_deleted = True
+        db.session.add(row)
+        db.session.commit()
+
+        resp = client.post('/r6/fhir/Bundle/$ingest-context',
+                           data=json.dumps(sample_bundle),
+                           content_type='application/json',
+                           headers=tenant_headers)
+        assert resp.status_code == 201, resp.get_data(as_text=True)
+
+        row = R6Resource.query.filter_by(
+            tenant_id=tenant_id, resource_type='Observation',
+            id=sample_observation['id']).first()
+        assert row.is_deleted is False
+        assert row.version_id == 2
+        assert json.loads(row.resource_json)['status'] == 'final'
--- a/tests/test_context_builder.py
+++ b/tests/test_context_builder.py
@@ -112,3 +112,47 @@ class TestContextBuilder:
         assert row.is_deleted is False
         assert row.version_id == 2
         assert json.loads(row.resource_json)['status'] == 'final'
+
+    def test_a_revived_row_is_readable_again_through_the_read_path(
+            self, client, sample_bundle, sample_observation, tenant_id,
+            tenant_headers, step_up_token):
+        """#509 defect 2, stated as the symptom rather than the column.
+
+        The sibling test asserts `is_deleted is False`. That is the fix's
+        mechanism; THIS is the thing a patient notices — the record was
+        unreadable, and after the re-ingest it reads. Asserting the symptom
+        is what keeps the pin honest if the read path ever stops filtering
+        `is_deleted` (then the 404 below fails and someone has to think),
+        and it is the half that would have caught the original bug: the
+        ingest reported success while every read still returned 404.
+
+        MUTATION: drop `existing.is_deleted = False` in
+        r6/context_builder.py -> the final GET returns 404, red.
+        """
+        stale = dict(sample_observation, status='preliminary')
+        row = R6Resource(
+            resource_type='Observation',
+            resource_json=json.dumps(stale, separators=(',', ':'), sort_keys=True),
+            resource_id=sample_observation['id'],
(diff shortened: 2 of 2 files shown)


aks129:
## QA review — **PASS** (with one pre-existing MEDIUM on the same function, filed below, not a blocker)

Reviewed at `19d7663`. The branch was checked out by another worktree, so I reviewed the same commit detached. `autoMergeRequest: null` (checked, not armed).

### Spec compliance (council D13)

D13: "`existing.is_deleted = False` in `r6/context_builder.py` where the upsert lands on a tombstone, matching what the Fasten ingester already does."

- Production change is **one line** (`r6/context_builder.py:95`) plus a two-line comment. `git diff origin/main -- r6/` = `3 insertions(+)`, nothing else.
- The Fasten equivalent is real and identical: `r6/fasten/ingester.py:422`, in the same `if existing:` branch after `update_resource`. The two paths now agree.
- The select stays unfiltered on `is_deleted` (`context_builder.py:86-89`) — correct, and necessary: filtering would miss the tombstone and attempt an insert on an existing composite PK.
- No extras. No behaviour smuggled in.

### The test really exercises the tombstone

Not taken on the body's word:

1. **Mutation** — delete `context_builder.py:95`, keeping the comment:
   `uv run python -m pytest tests/test_context_builder.py -q -p no:cacheprovider -k tombstone` → `1 failed`, `E assert True is False`. Reverted; diff back to `3 insertions(+)`.
2. **`version_id == 2` is meaningful** — `R6Resource.__init__` sets `version_id = 1` and `update_resource` increments (`r6/models.py:109,126`), so the assertion distinguishes an update from a fresh insert.
3. **The tombstone was genuinely invisible to a read path** — the assertion the test does *not* make, so I made it. Seeded the same tombstoned Observation, then drove real requests:

```
GET /r6/fhir/Observation/test-obs-1  (tombstoned)  -> 404
POST /r6/fhir/Bundle/$ingest-context               -> 201
GET /r6/fhir/Observation/test-obs-1  (after)       -> 200, status = "final"
```

   Read paths do filter (`r6/routes.py:607,712,1019,1331,…`), so the fix restores readability, not merely a column value.
4. **No regression on the ordinary path** — two consecutive ingests of a never-deleted bundle still give `version_id == 2`, `is_deleted False`.

Both probes are scratch files, run and removed; the worktree is as I found it. Say the word and I will contribute #3 as a committed test — it is the assertion that would catch a future read-path change, and it is stronger than the column check.

### Audit, as asked

**No audit event distinguishes the revive.** `r6/context_builder.py` emits none; the route emits exactly one bundle-level event (`r6/routes.py:1270-1277`), `create` / `Bundle`, with a count-based detail — `"ingested N resources; skipped M: …"` — which is PHI-free and unchanged by this PR. So resurrecting a tombstoned record is indistinguishable in the audit trail from an ordinary upsert. The Fasten path is no better (per-resource `create`, same silence). Pre-existing and out of D13's scope; worth a decision before the delete path ships, because "who un-deleted this record" is a question the trail cannot currently answer.

### Finding — MEDIUM, pre-existing, same function, **not a blocker for this PR**

**`$ingest-context` performs no resource-id validation at all, and a blank id is fabricated — so a re-ingest appends instead of upserting.**

`r6/context_builder.py:85` does `resource.get('id', str(uuid.uuid4()))`, which returns `''` when the key is present and empty; `R6Resource.__init__` then does `resource_id or str(uuid.uuid4())` (`r6/models.py:104`), minting a fresh UUID. Reproduction — POST the same bundle three times with `"id": ""` on the Observation:

```
ingest #1: Observation rows = 1  ids = ['909aa1d0']
ingest #2: Observation rows = 2  ids = ['909aa1d0', 'dcc4d9a4']
ingest #3: Observation rows = 3  ids = ['035b6486', '909aa1d0', 'dcc4d9a4']
```

That is verbatim the consequence D8 names — *"fabricating an id for a blank one turns an upsert into an append and makes every re-ingest a duplicate row: the #509 defect shape"* — still live on the path #509 is about. Same probe, other ids: `"has space"` → stored, `"a"*300` → stored (the column is `String(255)`; SQLite ignores the cap, **Postgres would raise a driver error**, which is the #306 over-length-id shape that cost a 250-resource Fasten import on 2026-07-08).

Why it is not a blocker here: pre-existing, unrelated to the one-line revive, and D13 asked for exactly that line. Why it matters now: **#546 closes this hole on `_ingest_one` (Fasten/SHC/direct-upload) and leaves it open on this path**, so after that merge the two ingest routes disagree about what a valid id is — the "each call site decides for itself" root cause #509 itself names. Recommend a follow-up issue: route `$ingest-context` through the same id guard, or extract one shared validator.

### Can #509 close?

The body's claim checks out, verified rather than read across: `git merge-base --is-ancestor b265cfa origin/main` succeeds — that is the trap CLAUDE.md names, since a moved ref is not a merged commit — and `r6/actions/rails/form_fill.py:63,72` carries the `is_deleted=False` filter with the comment naming the issue. With this PR, both listed defects are fixed. Using `Refs` and leaving the close to a human is the right call.

### Gates (on this commit)

```
uv run ruff check .
All checks passed!

3158 passed, 13 skipped, 1 xfailed, 2 warnings in 112.12s (0:01:52)

uv run python -m pytest tests/test_ratchets.py tests/test_guardrail_conformance.py -q -p no:cacheprovider
50 passed in 2.61s
```

Guardrail conformance — the Grade A release gate — passes on this commit, and on #545 / #546 inside their full-suite runs.

### Not covered

- SQLite only; the Postgres lane is CI. This matters more than usual here: the over-length-id case in the finding above behaves *differently* on Postgres, and my probe could not show that locally.
- No live server; everything above went through the Flask test client against real routes.
- I did not exercise a delete→revive→delete cycle, because no route accepts DELETE today (per #509's reachability note); the tombstone was seeded directly, as the PR's own test does.

PR #550, opened 2026-09-02
candidate: proposed-action payload is immutable once confirmed (#528)

### LINKED ISSUE #528: The executable payload can be swapped after the human approved it
`transition_action` refuses `status` in `**fields` and passes everything else straight into the guarded UPDATE — including `payload_json`.

```python
# r6/actions/state.py:29-31
if 'status' in fields:
    raise IllegalTransition('status cannot be passed via fields')
...
updates = dict(fields)          # :36
updates['status'] = to_state
```

Reproduced against the real models by the QA reviewer on the board for #521:

```
transition_action(aid, ['awaiting_confirmation'], 'executing',
                  actor='agent', payload_json=<new bytes>)  -> True
payload: {"to": "+15550000000", "script": "ORIGINAL TEXT SHOWN TO HUMAN"}
      -> {"to": "+19995551234", "script": "SWAPPED AFTER APPROVAL"}
```

No exception. One `ActionEvent`, recording **only** the status change — `from_status`, `to_status`, `actor`, `detail`. Nothing in the ledger says the payload moved.

## Why it matters

This is the failure the human gate exists to prevent: **the human approves text A and bytes B execute.** The draft human-gate spec (#521) makes "shown verbatim" one of its six out-of-band properties and rests it on payload immutability after commit — which it records as *unverified*. It is not merely unverified. It is reproducibly absent.

## A live write path already does it

`r6/actions/review.py:449` sets `action.payload_json = json.dumps(payload)` **after** `issue_confirmation(action_id, ...)` at line 445 — the payload is mutated after the approval credential is minted.

Today's use is benign: it appends `reviewed_qr_id`. The absence of any guard is not, and the ordering is the wrong way round regardless of what the mutation contains.

## Scope, honestly

**Not an execution bypass.** `/confirm` still requires `require_grant(audience=ACTION_APPROVAL_AUDIENCE, operation=action_id, consume_nonce=True)` (`r6/actions/routes.py:509`), which is unmintable without the mint secret. What fails is the integrity of what executes, and the ledger's ability to show it.

Sibling of #520 (a decline and a timeout produce the same record). Together: the ledger can neither distinguish a refusal from a timeout, nor prove that what ran is what the human saw.

## Suggested shape

Make `payload_json` immutable once an action leaves `proposed`, with a pinning test. Where a post-commit annotation is genuinely needed (the `reviewed_qr_id` case), put it in a column that is not the executable payload.

Found by the review board on #521, not by the test suite.

aks129:
## What

`ProposedAction.payload_json` is sealed once an `ActionConfirmation` exists for the action, and the review route now finalises the payload **before** it mints the confirmation.

- `r6/actions/review.py` — `reviewed_qr_id` is written into the payload first, then `issue_confirmation()` runs. The previous order minted the signature and then changed what it signed.
- `r6/actions/models.py` — `@validates('payload_json')` raises `PayloadSealed` on any ORM assignment once a confirmation names the action (pending or committed). A new row has no id, so construction passes.
- `r6/actions/confirmations.py` — `has_confirmation(action_id)` backs the seal: scans the session's pending rows, then queries without autoflush so a validator never flushes unrelated state as a side effect.
- `r6/actions/state.py` — `transition_action()` refuses `payload_json` in `**fields`, unconditionally, the same way it refuses `status`.

## Why

Human-gate spec §9 R2 (`docs/specs/2026-08-16-human-gate.md`): a confirmation is the human's signature over the payload they saw; a payload that changes after it means the ledger cannot say what was approved. Council ruling on the D2 dissent: land this guard now, ahead of any approve surface. Nothing from #215 / #520 is built here.

Two points that are stricter than "after a confirmation exists", on purpose:

- The `transition_action` refusal is **unconditional**, not confirmation-keyed. The `/confirm` route claims (`awaiting_confirmation -> executing`) *before* the confirmation row is minted (that ordering is the single-winner mutex, per its docstring), so at the one moment the #528 repro swaps the payload, "a confirmation exists" is still false. A confirmation-keyed check there would leave the repro open. No caller passes `payload_json` through `**fields` today.
- Bulk `Query.update()` never reaches a validator, which is why the transition writer needs its own guard rather than relying on the model seal.

## Writers covered

`grep -rn "payload_json" r6/` — every `ProposedAction.payload_json` site (the `agent_runs` hits are a different model's column):

| Site | What it is | Status |
| --- | --- | --- |
| `r6/actions/models.py` `__init__` | constructor, new row | passes the seal (no id yet), pinned by `test_constructing_a_new_action_is_not_sealed` |
| `r6/actions/review.py` (was `:449`) | `action.payload_json = ...` after `issue_confirmation` | reordered; now under the seal (mutation: restoring the old order trips `PayloadSealed` in 3 tests) |
| `r6/actions/state.py` `transition_action(**fields)` | bulk guarded UPDATE, the #528 repro | refused with `IllegalTransition` |

The two other bulk updates on `ProposedAction` (`routes.py` `external_ref` and `provider_request_at`) use literal dicts that cannot carry `payload_json`; they are unchanged and not reachable by the seal.

## Tests

- `tests/actions/test_review_flow.py::test_post_payload_is_final_before_the_confirmation_is_issued` — snapshots `payload_json` at the instant `issue_confirmation` runs and asserts it equals the stored value after the request. Red on `main` (snapshot lacked `reviewed_qr_id`).
- `tests/actions/test_payload_immutability.py` (new, 6 tests) — sealed after a committed confirmation, after a pending one, after a consumed one; still writable with no confirmation; scoped to the action (another action's confirmation does not seal); construction passes.
- `tests/actions/test_state_transitions.py::test_payload_json_via_fields_rejected` — the issue's exact repro raises, payload and status unchanged, no `ActionEvent`.

Each guard was mutation-tested: reverting the order, disabling the transition check, disabling the seal, dropping the pending scan, and widening the seal to any confirmation each produced a red test, then the code was restored.

Audit `detail` strings are unchanged; `PayloadSealed` carries only the action id.

## Ran

```
uv run ruff check .                                   # All checks passed!
uv run python -m pytest tests/ -q -p no:cacheprovider # 3165 passed, 13 skipped, 1 xfailed, 2 warnings in 106.47s
```

The 2 warnings are pre-existing (`tests/test_actions_routes.py:47` LegacyAPIWarning, `tests/test_ingest_resilience.py:96` SAWarning), untouched by this PR. Ratchets in `tests/test_ratchets.py` are green; `r6/routes.py` is not touched.

## What was NOT run

- No run against a live HealthClaw or the Postgres CI lane locally (SQLite in the suite; the Postgres lane is CI's).
- No e2e / Playwright.
- Not deployed; not merged.

## Noticed, deliberately not touched

- #528's suggested shape moves `reviewed_qr_id` out of the executable payload into its own column. That is a schema change beyond this guard; with the reorder the confirmation now covers the annotated payload, which is coherent, but the annotation still lives inside `payload_json`.
- The seal is keyed on the confirmation, not on leaving `proposed` (spec R2 says "after commit"). Keying on status would refuse the review route's own annotation and force the column move above. Follow-on if the team wants the stricter rule.

Closes #528

### DIFF
--- a/r6/actions/confirmations.py
+++ b/r6/actions/confirmations.py
@@ -33,6 +33,21 @@ def issue_confirmation(action_id, approved_via, ttl_minutes):
     return c
 
 
+def has_confirmation(action_id):
+    """True if any ActionConfirmation names action_id — committed, or still
+    pending in this session (the review route mints and commits in one
+    transaction). Backs the payload seal on ProposedAction (#528): once a
+    confirmation exists, the payload is what the human approved. Runs
+    without autoflush so a validator never flushes unrelated state as a
+    side effect; the pending scan covers what that flush would have found."""
+    for obj in db.session.new:
+        if isinstance(obj, ActionConfirmation) and obj.action_id == action_id:
+            return True
+    with db.session.no_autoflush:
+        return ActionConfirmation.query.filter_by(
+            action_id=action_id).first() is not None
+
+
 def consume_confirmation(action_id):
     """Atomically claim every unconsumed, unexpired confirmation for
     action_id. Returns True iff at least one row was consumed by THIS call
--- a/r6/actions/models.py
+++ b/r6/actions/models.py
@@ -17,6 +17,8 @@ import json
 import uuid
 from datetime import datetime, timedelta, timezone
 
+from sqlalchemy.orm import validates
+
 from models import db
 
 PROPOSAL_TTL_MINUTES = 30
@@ -44,6 +46,11 @@ _TRANSITIONS = {
 def _utcnow():
     # Stored naive-UTC; columns aren't timezone-aware, so this matches what other models' aware defaults become after DB round-trip.
     return datetime.now(timezone.utc).replace(tzinfo=None)
+
+
+class PayloadSealed(RuntimeError):
+    """A write to payload_json after a confirmation exists for the action.
+    Same idiom as the AuditEvent immutability listeners in r6/models.py."""
 
 
 class ProposedAction(db.Model):
@@ -87,6 +94,23 @@ class ProposedAction(db.Model):
 
     def is_expired(self):
         return _utcnow() >= self.expires_at
+
+    @validates('payload_json')
+    def _seal_payload_once_confirmed(self, key, value):
+        """The confirmation is the human's signature over the payload they
+        saw; once one exists the payload is what was approved, and any write
+        to it means the ledger cannot say what was (human-gate spec §9 R2,
+        #528). Fires on every ORM assignment, including __init__ — a new row
+        has no id, so no confirmation, so construction passes. Bulk
+        Query.update() never reaches a validator; transition_action() refuses
+        payload_json in **fields for that writer."""
+        # Local import: confirmations.py imports _utcnow from this module.
+        from r6.actions.confirmations import has_confirmation
+        if self.id is not None and has_confirmation(self.id):
+            raise PayloadSealed(
+                'payload_json is sealed: a confirmation exists for action %s'
+                % self.id)
+        return value
 
     @property
     def payload(self):
--- a/r6/actions/review.py
+++ b/r6/actions/review.py
@@ -37,8 +37,8 @@ from dataclasses import dataclass, field
 from flask import render_template, request
 
 from models import db
-from r6.actions.confirmations import issue_confirmation
-from r6.actions.models import ProposedAction
+from r6.actions.confirmations import has_confirmation, issue_confirmation
+from r6.actions.models import PayloadSealed, ProposedAction
 from r6.actions.routes import _error, _tenant_or_none, actions_blueprint
 from r6.audit import record_audit_event
 from r6.models import R6Resource
@@ -386,6 +386,20 @@ def review_submit(action_id):
     if action is None:
         return _error(404, 'Unknown action')
 
+    # A submitted review is final. This route does NOT transition the action
+    # (the confirm route's claim does), so the loader above keeps accepting an
+    # action that has already been reviewed — and the payload is sealed the
+    # moment the first submit mints a confirmation (#528). Without this the
+    # second submit reaches the sealed assignment below and raises out of the
+    # handler as a 500; the review page calls r.json() on every response, so a
+    # Werkzeug HTML 500 rejects the promise and the patient sees nothing move
+    # at all. A phone double-tap is the ordinary way to reach that. Answer
+    # instead, before any FHIR work or QuestionnaireResponse row is created.
+    if has_confirmation(action_id):
+        return _error(409, 'This review has already been submitted. Your '
+                           'answers were recorded and approved; nothing '
+                           'further is needed.')
+
     # RE-POPULATE from FHIR: the server, not the client, decides which rows
     # exist and must be acted on. This is what makes the gate un-craftable.
     _questionnaire, patient, draft_qr, _content = _draft_qr(action, tenant_id)
@@ -446,11 +460,23 @@ def review_submit(action_id):
     # consent record. The confirmation is the human's signature over the
     # payload as it stands, and payload_json is sealed once it exists (#528)
     # — the other order minted the signature and then changed what it signed.
-    payload = action.payload
-    payload['reviewed_qr_id'] = qr_row.id
-    action.payload_json = json.dumps(payload)
-    issue_confirmation(action_id, approved_via='review-page', ttl_minutes=15)
-    db.session.commit()
+    # The pre-check above is not a lock: two submits milliseconds apart both
+    # pass it, and the loser reaches the seal here with the winner's
+    # confirmation already committed. Same answer as the pre-check — this is
+    # the branch that made it a 500, and the FHIR re-population between the
+    # two is exactly wide enough for a double-tap to land in it.
+    try:
+        payload = action.payload
+        payload['reviewed_qr_id'] = qr_row.id
+        action.payload_json = json.dumps(payload)
+        issue_confirmation(action_id, approved_via='review-page',
+                           ttl_minutes=15)
+        db.session.commit()
+    except PayloadSealed:
+        db.session.rollback()
+        return _error(409, 'This review has already been submitted. Your '
+                           'answers were recorded and approved; nothing '
+                           'further is needed.')
 
     record_audit_event(
         'update', resource_type='ProposedAction', resource_id=action.id,
--- a/r6/actions/review.py
+++ b/r6/actions/review.py
@@ -442,11 +442,14 @@ def review_submit(action_id):
     db.session.add(qr_row)
     db.session.flush()          # assign qr_row.id
 
-    # (6) Consent record for the review approval + hand-off marker for Task 8.
-    issue_confirmation(action_id, approved_via='review-page', ttl_minutes=15)
+    # (6) Hand-off marker for Task 8 goes into the payload FIRST, then the
+    # consent record. The confirmation is the human's signature over the
+    # payload as it stands, and payload_json is sealed once it exists (#528)
+    # — the other order minted the signature and then changed what it signed.
     payload = action.payload
     payload['reviewed_qr_id'] = qr_row.id
     action.payload_json = json.dumps(payload)
+    issue_confirmation(action_id, approved_via='review-page', ttl_minutes=15)
     db.session.commit()
 
     record_audit_event(
--- a/r6/actions/state.py
+++ b/r6/actions/state.py
@@ -16,7 +16,7 @@ def transition_action(action_id, from_states, to_state, actor, detail=None,
     (and writes one ActionEvent in the same commit), False if the WHERE matched
     nothing (concurrent claim / already advanced). Raises IllegalTransition if
     to_state isn't reachable from every from_state, if from_states is empty,
-    or if 'status' is passed via fields.
+    or if 'status' or 'payload_json' is passed via fields.
 
     extra_criteria: optional iterable of SQLAlchemy predicates appended to
     the guarded UPDATE's WHERE, making the claim STRICTER — e.g. the expiry
@@ -29,6 +29,14 @@ def transition_action(action_id, from_states, to_state, actor, detail=None,
     which apply atomically with the transition."""
     if 'status' in fields:
         raise IllegalTransition('status cannot be passed via fields')
+    if 'payload_json' in fields:
+        # The executable payload is what the human saw; no transition may
+        # rewrite it (#528). The bulk UPDATE below bypasses the model-level
+        # seal, so this is the only place this writer can be refused — and
+        # it is unconditional because the claim runs BEFORE the confirmation
+        # row is minted (see the confirm route), so "a confirmation exists"
+        # is not yet true at the one moment a swap would matter.
+        raise IllegalTransition('payload_json cannot be passed via fields')
     if not from_states:
         raise IllegalTransition('from_states must be non-empty')
     for fs in from_states:
--- a/tests/actions/test_payload_immutability.py
+++ b/tests/actions/test_payload_immutability.py
@@ -116,3 +116,49 @@ def test_constructing_a_new_action_is_not_sealed(app):
         db.session.add(a)
         db.session.commit()
         assert _stored_payload(a.id) == ORIGINAL
+
+
+# ---------------------------------------------------------------------------
+# The seal is an ORM-layer control. It fires on attribute assignment, and a
+# bulk UPDATE, raw SQL, or set_committed_value writes straight past it —
+# verified, not assumed. transition_action() is refused because it is the one
+# bulk writer that exists; nothing structural stops the NEXT one, and there is
+# no payload digest on ActionConfirmation that would make a bypass visible
+# after the fact. So pin the writer set.
+#
+# Blind spot, stated rather than discovered later: this is a file-level
+# allowlist. A NEW write added inside a file already on the list does not trip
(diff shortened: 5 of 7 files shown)


aks129:
## QA review — **FAIL** on one HIGH and one MEDIUM, both fixed on the branch (`60d498a`)

Reviewed `eaad41b` in an isolated worktree. **The guard itself is sound**, and every non-trivial line of it is load-bearing — the mutation matrix below says so rather than the test count. The failure is the retro's exact shape: a control that looks like "prevent a payload swap" and quietly also does "500 the review page when a patient taps Submit twice."

Do not merge on my say-so; a maintainer still approves. Auto-merge checked and **not armed** before and after the push.

---

### 1. Spec compliance vs. the ruling — four touch points, and each one earns its place

The council asked for a "one-line guard now" (D2 dissent). This is four touch points, so I mutation-tested each one separately: break it, run `tests/actions/ tests/test_actions_routes.py`, restore.

| # | Mutation | Tests red | Load-bearing? |
|---|---|---|---|
| a | Revert the `review.py` reorder alone | **4** — `test_post_payload_is_final_before_the_confirmation_is_issued`, plus **3 pre-existing** flow tests (`test_post_confirming_real_allergy_succeeds`, `test_post_with_nka_affirmed_succeeds`, `test_end_to_end_propose_review_confirm_download`) | **Yes** |
| b | Remove the seal alone (`models.py:109`) | **3** — the committed / pending / consumed seal tests | **Yes** |
| c | Remove the `transition_action` refusal alone (`state.py:32`) | **1** — `test_payload_json_via_fields_rejected` (the issue's exact repro) | **Yes** |
| d | Drop the pending-session scan (`confirmations.py:43-45`) | **1** — `test_payload_sealed_while_the_confirmation_is_still_pending` | **Yes** |
| e | Drop `no_autoflush` (`confirmations.py:46`) | **0** | **No** — see LOW-1 |
| f | Revert the reorder **and** remove the seal | **4**, incl. the ordering test | Ordering is pinned **independently** of the seal |
| g | Widen the seal to "any confirmation exists" (unscoped) | **1** — `test_confirmation_for_another_action_does_not_seal` | Scoping is pinned |

Mutation (a) is the interesting row: reverting the reorder reddens **three pre-existing tests**, so the reorder is not optional tidiness — the seal makes the old ordering impossible. And (f) shows the ordering test stands on its own rather than passing only because the seal throws.

**Verdict: no defensive padding except `no_autoflush`.** The claim in the PR body that each guard was mutation-tested red is accurate; I reproduced it independently.

**One correction to the record.** There is no hash anywhere. `ActionConfirmation` is `id, action_id, approved_via, approved_at, expires_at, consumed_at` — **no payload digest column**. The PR body says "signature" metaphorically and does not overclaim, but writing it down so nobody later cites this PR as cryptographic binding: the confirmation binds the payload **temporally**, not cryptographically. Spec §8.3 step 9's byte equality holds only for as long as every writer stays sealed or refused, and a bypass would leave **no evidence at all**. That is what MEDIUM-1 is about.

---

### 2. Write-path enumeration — every path exercised, not reasoned about

`grep -rn "payload_json" r6/ careagents/ services/` → the only `ProposedAction` writers are `models.py` (constructor), `review.py`, and `state.py` `**fields`. The `r6/agent_runs` hits are a different model's column. The PR body's writer table is accurate.

I then drove each *mechanism* against a real action with a committed confirmation:

| Path | Result |
|---|---|
| 1. `setattr(action, 'payload_json', …)` | **SEALED** (`PayloadSealed`) |
| 2. `action.payload["x"]=…` + `flag_modified` | **no-op** — `.payload` is a property returning a fresh `json.loads` dict, so in-place mutation cannot reach the column. Not a bypass. |
| 3. `Query.update({'payload_json': …})` | **OPEN** (mechanism) — no live caller; `transition_action` is the only one and it refuses |
| 4. `session.execute(update(ProposedAction)…)` | **OPEN** (mechanism) — no live caller |
| 5. Raw SQL `UPDATE proposed_actions` | **OPEN** (mechanism) — no live caller |
| 6. `set_committed_value` + `flag_modified` | **OPEN** (mechanism) — no live caller |
| 7. `session.merge(detached)` | **SEALED** — `merge()` fires validators |
| 8. `__dict__` poke + `flag_modified` | **OPEN** (mechanism) — no live caller |

**No open path has a live caller**, so this is not a HIGH by the brief's rule. The two other bulk updates on `ProposedAction` (`routes.py:134` `external_ref`, `routes.py:620` `provider_request_at`) use literal dicts that cannot carry `payload_json`, as the PR body says. → **MEDIUM-1** is the durability gap, not a reachable bypass.

### 3. `has_confirmation` correctness — verified in all three states

- **Pending in session** (added, unflushed) → sealed, via the `session.new` scan.
- **Flushed but not committed** (no longer in `session.new`) → sealed, via the query in the same transaction. This is the gap the scan alone would leave and the query alone would leave; together they cover it.
- **Committed** → sealed.
- **Scoped to the action** → pinned by mutation (g).
- **No autoflush recursion**: the validator queries `ActionConfirmation`, never `ProposedAction`, so there is no validator re-entry. Probed with a dirty unflushed `ProposedAction` in the session: sealed, no recursion.

### 4. `/confirm` ordering — confirmed by reading the code

`r6/actions/routes.py`: the claim (`awaiting_confirmation → executing`) is step **(c)** at `:546`; `issue_confirmation` is step **(d)** at `:580`. The claim genuinely precedes the mint, exactly as the PR body and `state.py`'s comment argue. So a confirmation-keyed check in `transition_action` would be false at the one moment the swap matters, and the **unconditional** refusal is the right call. The issue's repro is pinned by `test_payload_json_via_fields_rejected` and reddens on mutation (c).

### 5. Hash coverage / later writes

After the reorder the confirmation is minted over the annotated payload (pinned by the snapshot test). **Nothing writes the payload between the confirmation and execution**: `rails/phone.py`, `rails/sms.py`, `rails/webhook_poster.py` and `rails/form_fill.py` only *read* `action.payload`; `_resolve_from_executing` writes `status`, `external_ref`, `outcome_summary`. The seal does **not** break execution. See §1 for the "no digest" correction.

### 6. Audit + PHI

`PayloadSealed`'s message is id-only — no payload content. Audit `detail` strings are untouched. One note that the fix removes: there is **no app-wide error handler**, so the uncaught exception's traceback frame held the full payload in `value`; Flask's default logger does not emit frame locals, but any error-tracking integration that captures locals would have. Moot now.

### 7. Ratchets — the soft-delete concern is **unfounded**, and the helper placement is not a dodge

`uv run python -m pytest tests/test_ratchets.py tests/test_guardrail_conformance.py -q -p no:cacheprovider` → **50 passed**. Grade A intact.

On whether `has_confirmation` was moved to dodge `test_soft_delete_blind_query_files_only_decrease`:

- **Neither table has `is_deleted`.** `proposed_actions` and `action_confirmations` are not soft-deletable, so there is no soft-delete blindness in fact.
- The placement is **forced**: `confirmations.py` already does `from r6.actions.models import _utcnow`, so a query on `ActionConfirmation` cannot live in `models.py` without a circular import — which is why the validator uses a function-local import in the other direction.
- Worth naming for the ratchet's own sake: `models.py` mentions `R6Resource` **only in a docstring** (line 12) and has no `.query` node. Had the query been inlined there, the ratchet would have counted it — a **false positive off a docstring keyword match**. The ratchet's `'R6Resource' in source` proxy is crude. Not this PR's problem; flagging it (LOW-3).

### 8. Gates

```
  eaad41b: 3165 passed, 13 skipped, 1 xfailed, 2 warnings in 135.26s
  60d498a: 3168 passed, 13 skipped, 1 xfailed, 2 warnings in 107.94s
uv run ruff check .                    All checks passed!   (both commits)
ratchets + conformance                 50 passed            (both commits)
```
The 2 warnings are pre-existing and untouched, as the PR body states.

---

## Findings

### HIGH-1 — a resubmitted review 500s, and on a phone the screen does not move at all
`r6/actions/review.py:336-338` (loader) + `r6/actions/models.py:109` (seal) — **fixed in `60d498a`**

`review_submit()` **never transitions the action**; the status stays `awaiting_confirmation`, so `_load_form_fill_action` keeps accepting an action that has already been reviewed. The first submit mints a confirmation, which seals `payload_json`. The second submit reaches `action.payload_json = json.dumps(payload)` and `PayloadSealed` raises out of the handler. There is **no app-wide error handler** → Werkzeug HTML 500.

That is worse than an error page here. `templates/action_review.html:249` does `r.json()` on **every** response with no `.catch()`, so an HTML body rejects the promise: the alert never renders, the Submit button stays enabled, nothing on screen changes. A patient on a phone taps Submit again — which is exactly how the second submit happens.

**Scenario:** patient reviews their intake form, taps Submit on a flaky in-app browser, sees no response, taps again → 500 → still no response.

**Reproduction (pre-fix, `eaad41b`):**
```
FIRST  -> 200   status=awaiting_confirmation  confirmations=1
SECOND -> r6.actions.models.PayloadSealed: payload_json is sealed:
          a confirmation exists for action 0542814f-…
```
Pre-PR baseline on `89b42fb`, same probe: `FIRST -> 200`, `SECOND -> 200`. **A live path this PR closed.**

**Fix:** two guards, because the first is not a lock.
- `has_confirmation()` pre-check right after the 404 load — refuses **before** `_draft_qr` re-reads the patient's record out of FHIR and before a QuestionnaireResponse row is created for a doomed request.
- `try/except PayloadSealed` around annotate + issue + commit — two submits milliseconds apart both pass the pre-check and the loser meets the seal with the winner's confirmation already committed; the FHIR re-population between the two is exactly wide enough. Same answer, other branch.

Both return **409** with a JSON body the page already renders via `res.b.error`.

Each half is independently mutation-pinned (restored after each):
| Mutation | Test red |
|---|---|
| pre-check disabled | `test_second_submit_answers_409_instead_of_raising_the_seal` |
| `except` branch disabled | `test_second_submit_racing_the_precheck_answers_409_not_500` |

The first test pins the pre-check via its *actual* property — `_draft_qr` monkeypatched to fail if called — because without that the `except` branch answers 409 either way and the rollback hides the QR insert. It also asserts one confirmation, the original `reviewed_qr_id`, unchanged status, and no orphan QuestionnaireResponse.

> Worth noting the pre-PR behaviour was not good either: the second submit minted a **second confirmation** with a different `reviewed_qr_id` and last-write-wins on the payload. 409 is strictly better than what it replaced.

**Known remaining window, stated rather than hidden:** if the loser's validator query runs *before* the winner commits, both submits succeed and two confirmations exist with last-write-wins payload. That is not a #528 swap — both are the human's own decisions and both are in the ledger — and the timing-independent control is `transition_action`'s unconditional refusal. The seal is ORM-layer defence-in-depth with a concurrency window; only a claim transition on the review route would close it, which is #215/#520 territory and out of scope here.

### MEDIUM-1 — nothing stopped the *next* bulk writer
`r6/actions/state.py:32-39` — **fixed in `60d498a`** (test added, no production change)

Paths 3, 4, 5, 6 and 8 above write straight past the seal. `transition_action` is refused because it is the one bulk writer that exists **today**; nothing structural stopped the next one, and with no digest on `ActionConfirmation` a bypass would leave no trace. Given the retro's lesson about controls that pass for reasons unrelated to the property they protect, "no live caller" is a fact about today, not a guarantee.

Added `test_only_the_allowlisted_files_touch_the_executable_payload` — a file-level allowlist over `r6/` (`ProposedAction` co-mention scopes it past `agent_runs`' unrelated column) pinned to `{models.py, review.py, state.py}`. Mutation-tested: dropping in a module doing `ProposedAction.query.filter_by(id=aid).update({"payload_json": blob})` turns it red with `Added: ['r6/actions/zz_fake_writer.py']`; removed, green again.

Its blind spot is **named in the test**: a new write inside a file already on the list does not trip it, nor would a writer outside `r6/`. It catches the realistic drift — another module starting to touch the executable payload — in the PR that does it.

### LOW-1 — `no_autoflush` is unpinned
`r6/actions/confirmations.py:46`. Mutation (e) removes it and the suite stays green. The reasoning in the docstring is sound and I would keep it, but nothing proves it is needed. Not fixed — pinning it would mean asserting an internal flush side effect, which is a worse test than no test.

### LOW-2 — `reviewed_qr_id` still lives inside the executable payload
#528's suggested shape was to move the post-commit annotation into its own column. The PR body already flags this as deliberately not done, and the reorder makes the current state coherent. Agreed as follow-on, not a blocker.

### LOW-3 — the soft-delete ratchet's `'R6Resource' in source` proxy matches docstrings
`tests/test_ratchets.py:299`. Would have false-positived on `r6/actions/models.py` (a docstring mention on line 12) had the query been inlined there. Pre-existing; noted so the next person who trips it knows why. Not fixed — out of this PR's scope.

---

## What I did **not** cover

- **No run against a running HealthClaw or CareAgents.** The 500 → 409 finding is from the real Flask route via the test client, not a live server; the mobile rendering claim is read from `templates/action_review.html:242-284`, not observed in a browser.
- **No Postgres lane.** SQLite only; that is CI's. Relevant because `ProposedAction.query.filter(...).update()` and `no_autoflush` behaviour can differ under a real concurrent transaction.
- **No genuine concurrency test.** The race branch is exercised by neutralising the pre-check, not by two simultaneous requests. The remaining window described under HIGH-1 is reasoned from the code, not reproduced.
- **No Playwright / e2e**, and no check of how CareAgents' relay (`careagents/app.py`) surfaces a 409 from the HealthClaw review route — worth one look before beta, since that relay is the actual caller on the consumer path.
- **Not deployed, not merged, auto-merge not armed.** `gh pr checks` deliberately not run.

<sub>Reviewed with Claude Opus 5 (1M context).</sub>

PR #544, opened 2026-09-02
chore(deps): npm audit fix — three transitive advisories turned dependency-audit red on every PR

aks129:
## What

`npm audit fix` in `services/agent-orchestrator`. Lockfile only, no `package.json` change.

- browserslist 4.28.2 → 4.28.8 (GHSA-c83g-rgw3-j3cx, GHSA-73wf-gq98-2v4g)
- fast-uri 3.1.5 → 3.1.7 (GHSA-5jgf-p345-68v8, GHSA-f65p-4m7j-42xc, GHSA-fph4-wmhf-6fwf, GHSA-jqff-g426-hqxp)
- qs (GHSA-x5fp-wj9c-mxmx)

## Why now

`dependency-audit` is red on #540 and #541 for these advisories, which were published after main's last green audit (run 33350710762 on `89b42fb`). Every PR from here inherits the red until this merges.

## Ran

- `npm audit --audit-level=high` → found 0 vulnerabilities
- `npm run lint` (tsc --noEmit) → clean
- `npm test` (jest) → green

## What was NOT run

The Python side of the audit (pip-audit) — unchanged and green on the same runs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #556, opened 2026-09-03
candidate: MCP phase 1 — protected-resource metadata and challenge behind MCP_CANONICAL_RESOURCE (#523, council D4)

### LINKED ISSUE #290: Locked MCP endpoint is unreachable from hosted connectors: no OAuth metadata, and the 401 gives no way forward

aks129:
## What

Phase 1 of `docs/specs/2026-08-16-mcp-authorization.md`, behind `MCP_CANONICAL_RESOURCE`. MCP server only; no Flask change, no phase 2, no phase 3.

Two commits: `a701ad9` serves the metadata and enriches the challenge; `b11a359` is the QA fix to boot validation described under Review below.

- `MCP_CANONICAL_RESOURCE` unset (today, and after this merges): both well-known routes 404 and the challenge is a bare `Bearer`. Asserted in tests, and measured at the socket by QA against a server built from `origin/main` — 33 requests, byte-identical after stripping `Date`, `ETag` and `/health`'s timestamp.
- Set to a canonical value, **and** the request `Host` names that host: the RFC 9728 document is served at `/.well-known/oauth-protected-resource/mcp` (the header-pointed sub-path) and at `/.well-known/oauth-protected-resource` (root fallback), with `Access-Control-Allow-Origin: *` on those two routes only, and the 401 on `/mcp` carries `resource_metadata` and `scope`.
- Set to a value that is not canonical: **the server refuses to start.** The rule is the property, not a list — the configured string must already be, character for character, the identifier a client derives from the URL we advertise (`origin + pathname`, or `origin` alone for a root path). The boot error names the variable and the shape, never the value.
- `resource` and the advertised metadata URL are both built from the constant, never from `req.hostname` (amendment P1-a) — because RFC 9728 §3.3 has a client reject a document whose `resource` is not the identifier it inserted into the well-known path. A mutation that builds the URL from the request host turns three tests red.
- `error="invalid_token"` appears only when a credential was presented; a credential-less 401 carries no `error` (amendment P1-c, RFC 6750 §3.1).

Who gets in does not change: `MCP_AUTH_TOKEN` remains the only accepted credential, `assertMCPAuthConfigured` is untouched, and `isPublicDemo()` still bypasses the challenge entirely (pinned).

Two things worth naming rather than leaving to be found:

- **The challenge carries `scope="fhir.read context.read"` as well as `resource_metadata`.** §3.4's two worked 401 examples both carry it, and assertion A1 requires `WWW-Authenticate` to contain `resource_metadata=` **and** `scope=`. The MCP specification marks `scope` SHOULD.
- **The enriched challenge applies to `/sse` and `/messages` too**, because it is built in the one auth middleware that guards all three transports. The same 401 needs the same way forward on any transport.

## Why

#523 — a conformant client that receives our 401 has no path to a credential, which is the mechanism behind #290. Council ruling D4 (2026-09-02): canonical resource `https://mcp.healthclaw.io/mcp`, issuer `https://app.healthclaw.io`, phase 1 merges now behind the flag and does not deploy until DNS.

This PR closes nothing on its own. #523 closes when phase 1 is deployed and verified; #568 tracks phases 2 and 3.

## Review

QA reviewed against two real servers rather than supertest: **FAIL at `a701ad9`**, **PASS at `b11a359`**.

The finding: the boot check refused a value that was not absolute or not `https:`, which is three of twelve malformed shapes an operator can plausibly type. The other nine — mixed-case scheme or host, an explicit `:443`, `?query`, `#fragment`, `user:pass@` userinfo, a bare trailing slash, dot segments, an empty authority — parse as https URLs and boot a server whose challenge advertises a document that the challenge's own reader then refuses under RFC 9728 §3.3. That is the spec's §9.3 mode: confidently wrong, which reads to a partner as our bug where a 404 reads as a gap.

The userinfo shape is worse than the others: the configured string is served verbatim as `resource`, so `user:pass@` puts a credential into an unauthenticated, CORS-open document. `b11a359` replaces the rule with the property above; ten new boot rows pin it, and removing the rule turns all ten red.

## Owner-only, before this is deployed

Nothing here is deployable by me, and the flag is unset in every environment, so merging changes no running behaviour.

1. **#522 first — repoint `mcp.healthclaw.io` from Vercel to the Railway service.** Verify over DoH, never `dig` on the operator machine:
   ```
   curl 'https://dns.google/resolve?name=mcp.healthclaw.io&type=CNAME'
   ```
2. Then set `MCP_CANONICAL_RESOURCE=https://mcp.healthclaw.io/mcp` on the MCP service, and only there.
3. Then the four verification curls:
   ```
   curl -sS -o /dev/null -D - -X POST https://mcp.healthclaw.io/mcp \
        -H 'content-type: application/json' \
        -d '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{}}'
   # expect: 401, WWW-Authenticate with resource_metadata= and scope=, no error=

   curl -sS https://mcp.healthclaw.io/.well-known/oauth-protected-resource/mcp
   # expect: 200 JSON, resource == https://mcp.healthclaw.io/mcp

   curl -sS -o /dev/null -w '%{http_code}\n' \
        https://mcp-server-production-5112.up.railway.app/.well-known/oauth-protected-resource/mcp
   # expect: 404 — the platform hostname must not serve a document whose
   # resource disagrees with the host it was fetched from (RFC 9728 §3.3)

   curl -sS -o /dev/null -D - -X POST https://mcp.healthclaw.io/mcp \
        -H 'content-type: application/json' -H "authorization: Bearer $MCP_AUTH_TOKEN" \
        -d '{"jsonrpc":"2.0","id":1,"method":"tools/list"}'
   # expect: 200 and the full tool list — the static path is unchanged
   ```

**Two open items belong on that checklist. Neither blocks merging a branch that ships inert; both block the deploy.** QA handed them back rather than patch them.

- **`resource_documentation` points at a URL that returns 400 to the reader it is for.** The document sends a partner to `https://app.healthclaw.io/r6/fhir/docs/privacy-policy`, which answers `400 {"diagnostics":"X-Tenant-Id header is required"}` unauthenticated (live, 2026-09-03). The tenant gate stands in front of the one human-readable pointer we publish, and the caller reading the PRM has no tenant by construction. The value comes from spec §3.2 and this PR implements it faithfully; the choice between exempting the route, repointing the field, and dropping it (RFC 9728 §2 marks it OPTIONAL) is an owner decision. Being tracked separately by the coordinator; the three options are recorded in the spec's §9.3 on #560.
- **CORS on the 401 is deferred, and the reason is stronger than scope.** Spec §9.4 asks for `Access-Control-Allow-Origin` on the refusal as well as on the PRM paths. Measured on this build, `OPTIONS /mcp` with `Origin: https://claude.ai` returns 204 with no `access-control-allow-origin`: a browser's `POST /mcp` with a JSON content type is preflighted, and **the preflight is refused before the 401 exists**. Adding the header to the 401 alone would look like the fix and not be one — a control that appears to do one thing and does another, which is the `docs/2026-08-02-retro.md` shape. The refusal and its preflight have to move together. That is part 2 of #523, and §9.4 on #560 now says so. Doing it here would also break this PR's byte-identity property, since it changes the flag-off path.

Positively: the PRM document itself *is* readable from a browser context. It is the challenge pointing at it that is not, so phase 1's reachability claim covers a server-side fetcher and not a browser one.

**Railway preserves `Host`, so the gate works — measured, not assumed.**

```
curl -sS 'https://app.healthclaw.io/r6/fhir/Condition?_count=1'
# fullUrl: http://app.healthclaw.io/r6/fhir/Condition/…
```

That `fullUrl` is built from `request.host_url` (`r6/routes.py:1128`), so the container behind the proxy sees the public custom domain. Measured 2026-09-03. No `X-Forwarded-Host` handling is needed, and none is added — QA confirmed the header is not read and does not get the document out of the platform hostname.

What the same measurement does show is that the proxy loses the **scheme**: every production authorization-server URL is `http://` on an https-only deployment. That is #567, not this PR's, and it is the live example of what building URLs from a request costs.

## Ran

Re-run at `b11a359`, in `services/agent-orchestrator`:
- `npm test` — `Test Suites: 9 passed, 9 total` / `Tests: 239 passed, 239 total` (43 of them in `protected-resource-metadata.test.ts`)
- `npm run lint` (`tsc --noEmit`) — clean
- `npm run build` (`tsc`) — clean

At the repo root:
- `uv run ruff check .` — `All checks passed!`
- `uv run python -m pytest tests/ -q -p no:cacheprovider` — `3157 passed, 13 skipped, 1 xfailed, 2 warnings in 104.46s`

CI: every check green except `dependency-audit`, which fails on `npm audit --audit-level=high` for two `browserslist` advisories (GHSA-c83g-rgw3-j3cx, GHSA-73wf-gq98-2v4g). Not this PR — it touches no `package.json` or lockfile, and the same job is red on the docs-only PR #560. The fix is in unmerged #544.

## NOT run

- **Nothing was executed against a deployed server.** QA's measurements were against locally built servers; the four curls above are the first measurement against production, and they are the owner's.
- **No client was run.** This makes no claim about what claude.ai or any other hosted connector does with the document. That is the spec's §8.4 end-user run, which is what closes #290.

### DIFF
--- a/services/agent-orchestrator/src/index.ts
+++ b/services/agent-orchestrator/src/index.ts
@@ -109,15 +109,40 @@ const AUTHORIZATION_SERVER_ISSUER = "https://app.healthclaw.io";
 const RESOURCE_SCOPES = ["fhir.read", "context.read"];
 const PRM_BASE_PATH = "/.well-known/oauth-protected-resource";
 
+// The check that matters is not "is this a URL" but "will the document we serve
+// survive RFC 9728 §3.3". A client derives the resource identifier by removing
+// the well-known segment from the URL it fetched, and rejects the document if
+// `resource` is not that string. So the configured value must already BE the
+// identifier a client derives, character for character — the constant is
+// advertised verbatim as `resource`, while the URL that carries it is built
+// from the parsed origin and path.
+//
+// Nine shapes parse as https URLs and fail that: a host or scheme in mixed case
+// (the URL parser lowercases both), an explicit `:443` (dropped), a `?query` or
+// `#fragment` (not in the path), `user:pass@` userinfo (also a credential we
+// would then publish in an unauthenticated, CORS-open document), a bare
+// trailing slash, and dot segments (resolved). Each one boots a server whose
+// own challenge points at a document the challenge's reader will refuse — the
+// §9.3 "confidently wrong" mode, which is the failure this check exists to
+// catch. Verified by starting the server on each value and fetching the
+// document at the URL its own challenge advertises.
 function parseCanonicalResource(raw: string | undefined): URL | null {
   if (!raw || !raw.trim()) return null;
+  const trimmed = raw.trim();
   let parsed: URL;
   try {
-    parsed = new URL(raw.trim());
+    parsed = new URL(trimmed);
   } catch {
     return null;
   }
   if (parsed.protocol !== "https:" || !parsed.hostname) return null;
+  // `origin` re-serializes: lowercased scheme and host, no userinfo, no default
+  // port. `pathname` is normalized and excludes query and fragment. A root path
+  // contributes nothing, because protectedResourceMetadataURL appends nothing
+  // for it and the identifier a client derives therefore has no trailing slash.
+  const canonicalForm =
+    parsed.pathname === "/" ? parsed.origin : `${parsed.origin}${parsed.pathname}`;
+  if (trimmed !== canonicalForm) return null;
   return parsed;
 }
 
@@ -857,8 +882,13 @@ function assertMCPCanonicalResourceConfigured(
   const raw = env.MCP_CANONICAL_RESOURCE;
   if (!raw || !raw.trim()) return; // Unset is the supported off state.
   if (!parseCanonicalResource(raw)) {
+    // Names the variable and the shape, never the value: the rejected value may
+    // be the reason it was rejected (userinfo credentials), and a boot crash is
+    // the loudest place in the system to print one.
     throw new Error(
-      "MCP_CANONICAL_RESOURCE must be an absolute https:// URL " +
+      "MCP_CANONICAL_RESOURCE must be an absolute https:// URL in canonical " +
+        "form — lowercase scheme and host, no userinfo, no default port, no " +
+        "query, no fragment, no trailing slash on a root path " +
         "(e.g. https://mcp.healthclaw.io/mcp)"
     );
   }
--- a/services/agent-orchestrator/src/index.ts
+++ b/services/agent-orchestrator/src/index.ts
@@ -92,6 +92,102 @@ function isMCPBearerCredential(authorization: string | undefined): boolean {
   );
 }
 
+// --- RFC 9728 protected-resource metadata (phase 1, behind the constant) ---
+//
+// docs/specs/2026-08-16-mcp-authorization.md §3.2/§3.4, as amended 2026-09-02.
+// The canonical resource identifier is one pinned string, and everything below
+// is derived from it. Unset, none of this runs: the well-known paths 404 and
+// the challenge stays a bare `Bearer`, which is today's behaviour. Read
+// per-request like MCP_AUTH_TOKEN so config is authoritative at call time;
+// assertMCPCanonicalResourceConfigured re-reads it at boot so a malformed value
+// refuses to start rather than disabling the feature in silence.
+//
+// This admits nobody. It adds two unauthenticated documents that hold no
+// secret, name no tenant and grant nothing, and two parameters on a refusal
+// that is still a refusal.
+const AUTHORIZATION_SERVER_ISSUER = "https://app.healthclaw.io";
+const RESOURCE_SCOPES = ["fhir.read", "context.read"];
+const PRM_BASE_PATH = "/.well-known/oauth-protected-resource";
+
+function parseCanonicalResource(raw: string | undefined): URL | null {
+  if (!raw || !raw.trim()) return null;
+  let parsed: URL;
+  try {
+    parsed = new URL(raw.trim());
+  } catch {
+    return null;
+  }
+  if (parsed.protocol !== "https:" || !parsed.hostname) return null;
+  return parsed;
+}
+
+function canonicalResource(): URL | null {
+  return parseCanonicalResource(process.env.MCP_CANONICAL_RESOURCE);
+}
+
+// RFC 9728 §3.1 inserts the resource's path after the well-known segment, so
+// `https://host/mcp` is described at `<host>/.well-known/oauth-protected-resource/mcp`.
+// Built from the constant and never from req.hostname. RFC 9728 §3.3 has a
+// client reject a document whose `resource` is not the identifier it inserted
+// into the well-known path, so what we advertise cannot follow the request —
+// whatever any proxy in front of us does or does not rewrite.
+function protectedResourceMetadataURL(resource: URL): string {
+  const path = resource.pathname === "/" ? "" : resource.pathname;
+  return `${resource.origin}${PRM_BASE_PATH}${path}`;
+}
+
+// `resourceIdentifier` is the configured string as configured, not a
+// re-serialized URL, so it string-equals the audience the client will request
+// and the authorization server will record. Nothing here comes from the
+// request.
+function protectedResourceMetadata(resourceIdentifier: string): Record<string, unknown> {
+  return {
+    resource: resourceIdentifier,
+    authorization_servers: [AUTHORIZATION_SERVER_ISSUER],
+    scopes_supported: RESOURCE_SCOPES,
+    bearer_methods_supported: ["header"],
+    resource_name: "HealthClaw Guardrails",
+    resource_documentation: `${AUTHORIZATION_SERVER_ISSUER}/r6/fhir/docs/privacy-policy`,
+  };
+}
+
+// A document whose `resource` disagrees with the host it was fetched from is
+// rejected by a conformant client (RFC 9728 §3.3), and that reads to a partner
+// as our bug. Until DNS points the canonical name here, the platform hostname
+// keeps answering exactly as it does today.
+function hostIsCanonical(req: express.Request, resource: URL): boolean {
+  const host = req.headers.host;
+  if (typeof host !== "string" || !host) return false;
+  try {
+    return new URL(`https://${host}`).hostname === resource.hostname.toLowerCase();
+  } catch {
+    return false;
+  }
+}
+
+function challengeHeader(req: express.Request): string {
+  const resource = canonicalResource();
+  if (!resource || !hostIsCanonical(req, resource)) return "Bearer";
+
+  const authorization = req.headers.authorization;
+  const credentialPresented =
+    typeof authorization === "string" && authorization.trim().length > 0;
+  // RFC 6750 §3.1: SHOULD NOT send `error` when no credential was presented —
+  // there is no token to call invalid. The description below distinguishes
+  // nothing on purpose: expired, wrong audience and not-a-token recover
+  // identically, and telling them apart is an oracle for a caller with no
+  // business getting one. It never carries the credential or a tenant.
+  const params = credentialPresented
+    ? [
+        'error="invalid_token"',
+        'error_description="The access token is invalid, expired, or was not issued for this resource."',
+      ]
+    : [];
+  params.push(`resource_metadata="${protectedResourceMetadataURL(resource)}"`);
+  params.push(`scope="${RESOURCE_SCOPES.join(" ")}"`);
+  return `Bearer ${params.join(", ")}`;
+}
+
 // Health probes and CORS preflight remain public. When configured, every MCP
 // network transport requires the deployment-scoped bearer credential.
 app.use((req, res, next) => {
@@ -106,7 +202,7 @@ app.use((req, res, next) => {
   }
 
   if (!isMCPBearerCredential(req.headers.authorization)) {
-    res.setHeader("WWW-Authenticate", "Bearer");
+    res.setHeader("WWW-Authenticate", challengeHeader(req));
     return res.status(401).json({ error: "Unauthorized" });
   }
   next();
@@ -677,6 +773,38 @@ app.post("/mcp/rpc", async (req, res) => {
   }
 });
 
+// --- Protected-resource metadata (RFC 9728) ---
+//
+// Mounted rather than declared per-path so the sub-path form always matches the
+// path inside the constant, which is the same path the challenge advertises. A
+// header pointing at a URL this server does not serve is the failure mode the
+// spec's §9.3 calls the worst one: confidently wrong beats absent, for the
+// partner reading the error.
+//
+// Both forms are served because clients try the sub-path first and the root
+// second. `next()` on every other case leaves the response byte-identical to
+// today's 404.
+app.use(PRM_BASE_PATH, (req, res, next) => {
+  if (req.method !== "GET" && req.method !== "HEAD") return next();
+
+  const resource = canonicalResource();
+  if (!resource || !hostIsCanonical(req, resource)) return next();
+
+  const isPathInsertionForm = req.path === resource.pathname;
+  const isRootForm = req.path === "/";
+  if (!isPathInsertionForm && !isRootForm) return next();
+
+  // Unauthenticated and readable cross-origin. A document a client must read
+  // before it can authenticate cannot itself require authentication, and a
+  // browser-context client that cannot read it is back at the dead end this
+  // phase removes (spec §9.4). The transport endpoint's origin allowlist is
+  // untouched.
+  res.setHeader("Access-Control-Allow-Origin", "*");
+  return res.json(
+    protectedResourceMetadata((process.env.MCP_CANONICAL_RESOURCE || "").trim())
+  );
+});
+
 // --- Health Check ---
 
 app.get("/health", (_req, res) => {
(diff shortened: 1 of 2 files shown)


aks129:
## QA review — FAIL on arrival at `a701ad9`, one MEDIUM fixed in `b11a359`, re-verified

Independent adversarial review. Everything below was run against **two real servers** built from `origin/main` and from this branch, not against supertest — unit fakes accept wire formats production rejects, and two of the properties this PR claims are only visible at the socket.

### The claim I most wanted to break, and could not

**Flag off is byte-identical to today.** Built `origin/main`'s `index.ts` and this branch's, ran each on the same port with `MCP_AUTH_TOKEN` set and `MCP_CANONICAL_RESOURCE` unset, and captured 33 requests against each: both well-known paths, an uppercase spelling of the well-known path, `POST /mcp` initialize with no credential and with a wrong one, `/sse`, `/messages`, `/mcp/rpc`, `/health`, `tools/list` with the token, `OPTIONS` with an `Origin`, and ten Host spellings.

```
diff <normalized main> <normalized branch>   ->  exit 0, zero differences
```

The only unnormalized differences across 76 lines are `Date`, the `ETag` of `/health`, and `/health`'s own `timestamp`. The 404 body, `Content-Length: 179`, the header set and their order are identical. `tools/list` returns the same 27 tools in all three configurations (main flag-off, branch flag-off, branch flag-on). **This half of the PR holds.**

### Host gate — nothing gets the document or the enriched challenge out of the platform hostname

Tried, flag on. `-> 404` means the metadata 404s and the challenge stays bare.

| Host presented | Result |
|---|---|
| `mcp-server-production-5112.up.railway.app` | -> 404, bare `Bearer` |
| that host + `X-Forwarded-Host: mcp.healthclaw.io` | -> 404, bare `Bearer` |
| absolute-form request line `GET https://mcp.healthclaw.io/... ` with the platform Host | -> 404, bare `Bearer` |
| no `Host` at all (HTTP/1.0) | -> 404, bare `Bearer` |
| empty `Host:` | -> 404 |
| `mcp.healthclaw.io.` (trailing dot) | -> 404 (fail-closed; WHATWG keeps the dot) |
| `notmcp.healthclaw.io`, `mcp.healthclaw.io.evil.com` | -> 404 |
| `[::1]`, `[::ffff:mcp.healthclaw.io]` | -> 404 |
| duplicate `Host:` platform-then-canonical | -> 404 (Node keeps the first) |
| duplicate `Host:` canonical-then-platform | -> 200, and the document is the constant |
| `MCP.HEALTHCLAW.IO`, `:8443`, `evil@mcp.healthclaw.io`, `mcp.healthclaw.io/evil`, trailing space | -> 200, and the document is the constant |

The last two rows are the gate being lenient about *spellings of the canonical hostname* — port, case, userinfo and trailing garbage are not compared. **In no case does a non-constant value appear in `resource` or `resource_metadata`**, which is the property that matters, so these are LOW. `X-Forwarded-Host` is never read, and the design note saying no such handling is needed is correct.

### MEDIUM — nine malformed constants booted, into exactly the document §9.3 calls the worst failure. Fixed in `b11a359`

The PR says "a malformed value refuses to boot". It refuses three shapes: not absolute, not `https:`, not a URL. Twelve shapes an operator can plausibly type are malformed; **nine of them parse as https URLs and boot.**

Measured by starting the server on each value, reading the `resource_metadata` URL out of its own challenge, fetching the document at that URL, and deriving the identifier the way RFC 9728 §3.3 has the client derive it:

| `MCP_CANONICAL_RESOURCE` | boots | document `resource` | client derives | |
|---|---|---|---|---|
| `https://MCP.healthclaw.io/mcp` | yes | `https://MCP.healthclaw.io/mcp` | `https://mcp.healthclaw.io/mcp` | rejected |
| `HTTPS://mcp.healthclaw.io/mcp` | yes | `HTTPS://…` | `https://…` | rejected |
| `https://mcp.healthclaw.io:443/mcp` | yes | `…:443/mcp` | `…/mcp` | rejected |
| `https://mcp.healthclaw.io/mcp?x=1` | yes | `…/mcp?x=1` | `…/mcp` | rejected |
| `https://mcp.healthclaw.io/mcp#frag` | yes | `…/mcp#frag` | `…/mcp` | rejected |
| `https://user:pass@mcp.healthclaw.io/mcp` | yes | `https://user:pass@…` | `https://…` | rejected |
| `https://mcp.healthclaw.io/` | yes | `https://mcp.healthclaw.io/` | `https://mcp.healthclaw.io` | rejected |
| `https://mcp.healthclaw.io/mcp/../admin` | yes | `…/mcp/../admin` | `…/admin` | rejected |
| `https:///mcp` | yes | `https:///mcp` | `https://mcp` | rejected |

Every one of these is the failure the boot check exists to prevent, stated in its own comment: *"a value that cannot be an audience is a configuration error, and the failure it would otherwise produce is silence."* The silence it actually produces is worse than the one described — not a phase 1 that looks deployed and is off, but a phase 1 that looks deployed, answers 200, and hands a conformant client a document it must refuse. §9.3 calls that the highest-severity mode because the partner reads our error as our bug.

The `user:pass@` row does one thing more. `protectedResourceMetadata` returns the configured string verbatim, so the credential is served as `resource` in a document that is unauthenticated and carries `Access-Control-Allow-Origin: *`.

**Fixed** in `b11a359`, as one rule stated as the property rather than a list of rejections: the trimmed value must already string-equal `origin + pathname` (or `origin` alone for a root path) — the identifier a client derives. All nine now refuse to boot. Values that are self-consistent still boot even when they are not the chosen constant: `:8443`, a trailing slash on a non-root path, an uppercase path segment, a bare origin, surrounding whitespace. The check rejects documents a client would refuse, not values it dislikes.

After the fix, the same matrix: **every value that boots agrees under §3.3, zero mismatches.** Flag-off byte-identity re-measured after the fix — still `diff` exit 0.

Boot behaviour that was already right and stays right: the process exits non-zero, and the message names the variable and now the shape, never the value. Asserted, because one of the shapes it rejects is a credential.

### The metadata document — checked against RFC 9728 §3 and the pinned MCP revision

`resource` REQUIRED and present; `authorization_servers`, `scopes_supported`, `bearer_methods_supported`, `resource_name`, `resource_documentation` are all §2-legal; nothing forbidden, nothing invented. `content-type: application/json`. Unauthenticated with `MCP_AUTH_TOKEN` set — correct, a document you must read before you can authenticate cannot require authentication. Names no tenant and is identical for every caller, so it is not an existence oracle. `POST` to it 404s; `GET` and `HEAD` answer. It is rate-limited like everything else.

**The path-insertion form agrees exactly** with what the challenge advertises. The root form does not, and cannot — root-form insertion implies `https://mcp.healthclaw.io`, the document says `…/mcp`. That is P1-d's acknowledged non-conformance, correctly implemented as written, and A2/A3 assert against the sub-path.

One note that is not a defect: `app.use()` mounts are case-insensitive, so `/.WELL-KNOWN/OAUTH-PROTECTED-RESOURCE/mcp` also serves the document when the flag is on. It 404s identically to main when the flag is off, holds no secret, and no client will request it. LOW, informational.

### MEDIUM — the one pointer we publish for a human returns 400

`resource_documentation` sends a partner to `https://app.healthclaw.io/r6/fhir/docs/privacy-policy`. Fetched unauthenticated, live, 2026-09-03:

```
HTTP/2 400
{"issue":[{"code":"security","diagnostics":"X-Tenant-Id header is required"…}],"resourceType":"OperationOutcome"}
```

The tenant gate stands in front of the page, and the caller reading the PRM has no tenant yet by construction. This is §9.3 in miniature: a URL we publish, that errors, in the document that exists so a partner does not hit a dead end.

**Not fixed here, deliberately.** The value comes from spec §3.2 and this PR implements it faithfully; choosing between exempting the route, repointing the field, and dropping it (RFC 9728 §2 marks it OPTIONAL) is a product/CTO call, not QA's. Recorded in §9.3 on #560 with the three options, and it belongs in the pre-deploy checks alongside the four curls. Inert until the flag is set.

### Spec gap, carried to #560 rather than patched here — CORS on the 401

Spec §9.4 says **"Required in phase 1:** the two PRM paths **and the 401 response** must carry `Access-Control-Allow-Origin: *`". The PRM paths do; the 401 does not. The PR discloses this as part 2 of #523, and deferring it is the right call — but not for the reason of scope alone. Measured on this build:

```
OPTIONS /mcp   Origin: https://claude.ai, ACRM: POST, ACRH: content-type
   -> 204, and NO access-control-allow-origin
```

A browser's `POST /mcp` with a JSON content type is preflighted, and the preflight is refused before the 401 exists. **Adding the header to the 401 alone would look like the fix and not be one** — the retro's control-that-does-two-things shape. The refusal and its preflight have to move together. §9.4 has been amended on #560 to say that rather than to keep asserting a phase-1 requirement phase 1 does not meet. Fixing it here would also have broken this PR's own byte-identity property, since it changes the flag-off path.

Worth stating positively: the PRM document *is* readable from a browser context (`ACAO: *` on a simple GET). It is the challenge that points at it that is not.

### Test honesty

Ran this PR's new suite against `origin/main`'s `index.ts` (boot check stubbed so the file compiles): **10 of 24 fail, 14 pass.** The 14 that pass are the flag-off, platform-hostname, static-token, `/health` and demo-mode rows — they pin no-change, and passing on main is what they should do. Nothing in the suite is decorative.

Both claimed mutations reproduce exactly: building the metadata URL from a host other than the constant turns **3** red; dropping the canonical-host gate turns **2** red. Removing my new canonical-form rule turns **10** red.

`/sse` and `/messages` do get the enriched challenge, as the PR says. That is right, not just harmless: the client derives the identifier from the `resource_metadata` URL it was handed, never from the transport path it hit, so §3.3 is satisfied on all three. `/mcp/rpc` gets it too.

### Gates, verbatim

```
services/agent-orchestrator:
  npm ci                          clean
  npm test    (at a701ad9)        Test Suites: 9 passed  Tests: 220 passed
  npm test    (at b11a359)        Test Suites: 9 passed  Tests: 239 passed
  npm run lint  (tsc --noEmit)    clean
  npm run build (tsc)             clean

repo root:
  uv run ruff check .             All checks passed!
                                  3157 passed, 13 skipped, 1 xfailed in 104.36s
                                  37 passed          (Grade A gate intact)
```

No security invariant in `docs/development.md` §"Security invariants" is touched: no write path, no PHI, no audit call, no step-up path, no redaction change.

`dependency-audit` is red here from npm advisories fixed in unmerged #544, unrelated to this branch. Deployment stays owner-gated behind the DNS move in #522. I did not merge, arm auto-merge, deploy, or run `gh pr checks`; `autoMergeRequest` was null before and after my push.

### What I did NOT cover

- **Nothing ran against the deployed MCP server.** Both servers were local builds. The four curls in the PR body remain the first real measurement and they are the owner's.
- **No MCP client was run.** No claim here about what claude.ai, ChatGPT or Perplexity do with this document — that is the §8.4 end-user run.
- **The Host gate was never exercised through the platform proxy.** The "Railway preserves Host" measurement is the PR's, not re-run by me; I verified the code reads `req.headers.host` and never `X-Forwarded-Host`, which is what makes that measurement the load-bearing one.
- **No concurrency or load testing**, and no check of PRM behaviour under rate-limit exhaustion beyond confirming the route sits behind the limiter.
- IPv6 was probed only as a Host-header spelling, not as a real listening socket.

### Verdict

**FAIL at `a701ad9`** on the boot-validation MEDIUM. **PASS at `b11a359`**, with two MEDIUMs handed back rather than patched — the `resource_documentation` 400, which is a product call, and the 401 CORS, which is #523 part 2 and needs the preflight fixed with it. Neither blocks merging a branch that ships inert; both block the deploy, and both belong on the pre-deploy checklist next to the four curls.

Not merging, and not approving my own fix — a maintainer owns that gate.

aks129:
The `resource_documentation` item in the pre-deploy section is now tracked as #574.

It turned out to be wider than this PR's metadata document. The same URL is published in `_disclaimer.url` on every FHIR resource we return and as `service_documentation` in OAuth discovery, and it answers 400 without a tenant header in all three places. So the broken link reaches anyone reading a resource outside our own client, not just a metadata reader.

Confirmed against production on 2026-09-03. The public policy page at `/privacy` answers 200; it is the API-flavoured copy that is gated.

PR #560, opened 2026-09-03
docs(spec): MCP authorization — the council's amendments, P1-a..d, P2-a..f, P3-a (council D4)

aks129:
## What

The council's spec amendments (2026-09-02, Interop seat, ruling D4) applied to `docs/specs/2026-08-16-mcp-authorization.md`, plus the corrections QA made when it reviewed them against the code. Each amendment is an edit in the section it belongs to, with a dated "Amendments 2026-09-02" index at the end (new §12) listing them by id. Docs only — no code in this PR; the phase 1 code is #556.

Two commits: `5095f92` applies the amendments; `9ddb21a` is QA's pass, described under Review.

Phase 1, merged behind the flag:

- **P1-a** (§3.2, §3.4) — the `resource_metadata` URL is built from `MCP_CANONICAL_RESOURCE`, never `req.hostname`. The document already said this for the PRM `resource`, and the header needs the same rule. See the correction below: the reason the amendment gave is not the real one.
- **P1-b** (§3.1, §7) — the fail-closed rule is replaced. It keyed on OAuth acceptance being enabled, which phase 1 predates, so there was nothing for a boot check to read. New rule: unset means 404 and a bare challenge; set means both ship; a non-canonical value refuses to boot; either is served only when `Host` names the canonical host.
- **P1-c** (§3.4) — citation fix. RFC 6750 §3.1 says SHOULD NOT, not "does not permit". The behaviour asked for is unchanged.
- **P1-d** (§3.2, §8.1) — "byte-identically at both paths" is not conformant for the root form. Both keep being served, and A2/A3 now assert against the header-pointed sub-path URL.

Phase 2, not yet built: **P2-a** the `302` with `Location` and `iss` (§3.5, and A6 follows it); **P2-b** tenant binding from config in a new §3.5.1, with the allowlist becoming a map; **P2-c** RFC 8707 `resource` agreement across both endpoints; **P2-d** the four RFC 7591 registration fixes; **P2-e** introspection protected, with the credential named; **P2-f** the token lifetime stated in a new §6.8.

Phase 3, not yet built: **P3-a** (§3.6, §8.2) — the JSON-RPC tool-argument overrides set the same three downstream headers §3.6 was written to stop. Now covered, and asserted as R8.

Header rows record the canonical resource, the issuer, which issue phase 1 answers (#523, once deployed), and that #290 closes on the §8.4 end-user run with #568 tracking what remains.

**This PR closes no issue.** It is a design document. #523 is answered by phase 1 running in production (#556 plus the DNS move in #522), and a docs merge that closed it would be a green check whose subject never ran.

## Why

Council ruling D4: the amendments land as a docs PR before phase 2 is built. Two of them are corrections to this document rather than to the design, and both would have been found by whoever tried to implement it — P1-b by a boot check with nothing to key on, P1-c by reading RFC 6750.

## Review — four more places the document described code that is not there

QA re-checked all seven Flask claims against the source and they hold. It found four statements about our own system that did not:

- **P1-b's malformed rule, "not absolute, or not `https:`", was three of twelve shapes.** Nine others parse as https URLs and boot a server whose challenge advertises a document its own reader refuses under RFC 9728 §3.3 — including `user:pass@`, which is served verbatim as `resource` in an unauthenticated, CORS-open document. §3.1 now states the rule as the property the check enforces: the configured value must already be the identifier a client derives. #556 implements it.
- **§3.1's Host gate said "equals the canonical host".** The comparison is on hostname — port, case and userinfo in the header are not part of it, and `X-Forwarded-Host` is not read. Measured across ten Host spellings, duplicate headers and an absolute-form request target.
- **§9.4 carried two line numbers into `src/index.ts`** (`:97`, `:131`) that are correct on `main` and wrong the moment phase 1 merges — the same staleness this PR fixed for P3-a, in a section it had not touched. Replaced with the ordering, which is the fact that matters.
- **§9.4 called the 401 CORS a phase 1 requirement, which phase 1 does not meet — and one header would not meet it either.** A browser's `POST /mcp` is preflighted, and `OPTIONS /mcp` also answers without `Access-Control-Allow-Origin`, so the preflight is refused before the 401 exists. The requirement stands; it is no longer labelled phase 1's, it says the refusal and its preflight have to move together, and it says plainly that phase 1's reachability claim covers a server-side fetcher and not a browser one. Tracked as part 2 of #523.

**§9.3 gains a fifth row, and it is ours.** §3.2's document points a partner at `https://app.healthclaw.io/r6/fhir/docs/privacy-policy`, which returns `400 X-Tenant-Id header is required` unauthenticated (live, 2026-09-03). The tenant gate stands in front of the one human-readable pointer in a document whose reader has no tenant yet. Recorded with three options — exempt the route, repoint the field, or drop it, since RFC 9728 §2 marks it OPTIONAL — and left undecided, because it is a product call. It belongs in #556's pre-deploy checks.

## The Flask claims, checked against the code

The P2 amendments assert what `r6/oauth.py` does today. Those assertions arrived with the ruling, and this document's header says nothing in it was inferred from reading code alone, so all seven were checked against the source on 2026-09-03 and each carries a `file:line`. Read, not run — source reading, not a measurement against a deployment, and dated because line numbers move.

**All seven held; nothing needed correcting.**

| Claim | Where |
|---|---|
| `/oauth/authorize` returns the code in a JSON body, no `302`, no `iss` | `r6/oauth.py:307-313` |
| the tenant comes from `X-Tenant-Id`, and the 403 that limits it to public tenants | `r6/oauth.py:284`, `r6/oauth.py:286` |
| no `resource` parameter is read at either endpoint | no occurrence in `r6/oauth.py` |
| no introspection endpoint exists, and `token` never checks a `client_secret` | no occurrence in `r6/oauth.py` |
| `OAUTH_TOKEN_TTL` defaults to 3600 and no `refresh_token` is ever issued | `r6/oauth.py:33`, `r6/oauth.py:371-375` |
| `_oauth_authorizes` checks tenant and scope, not audience | `r6/read_auth.py:35-41` |
| `APP_ENV`, `READ_AUTH_ENABLED`, `PUBLIC_TENANTS` are real knobs for P2-b's pre-deploy check | `r6/runtime_config.py:49`, `:105`, `r6/command_center/access.py:95` |

Two details the reading added to P2-d, which the amendment as ruled did not name:

- `/oauth/register` ignores the requested `token_endpoint_auth_method` and always answers `client_secret_post` (`r6/oauth.py:243`), while discovery advertises both `client_secret_post` and `none` (`r6/oauth.py:186`). A public client is told one thing and registered as another.
- `client_id` at `/oauth/token` is checked only when the client sends it (`r6/oauth.py:343`), which is exactly what P2-d makes mandatory for public clients.

## One correction to the amendment as ruled

**P1-a's stated reason, that the platform proxy does not preserve the Host header, is false**, and it is corrected in the spec rather than repeated in a document people will cite.

```
curl -sS 'https://app.healthclaw.io/r6/fhir/Condition?_count=1'
# fullUrl: http://app.healthclaw.io/r6/fhir/Condition/…
```

That `fullUrl` is built from `request.host_url` (`r6/routes.py:1128`), so the container behind the proxy sees the public custom domain. Host survives. Measured 2026-09-03.

The instruction is unchanged and its real reason is stronger: RFC 9728 §3.3 has a client reject a document whose `resource` is not the identifier it inserted into the well-known path, so what we advertise cannot follow the request — under any proxy, including one that does rewrite Host. §3.2 and the §12 index now say that.

The same measurement found what the proxy does lose: the **scheme**. Every production authorization-server URL is `http://` on an https-only deployment, which is what §3.3 already describes. Now filed as #567 and cited in §3.2, §3.3 and §11 as the live example of what building URLs from a request costs. Not fixed here.

## Two fixes beyond the amendments

- **#568 is cited, in four places.** The successor to #290 ("MCP authorization phases 2 and 3: a hosted connector still cannot reach a tenant") appears in the header table, in §6.1 where the follow-on consent specification is described, in §10's answer to open question 4, and in §11. A reader of the spec finds what remains without leaving the document, which is what D4 asked for.
- **§3.2's CORS cross-reference said §8.4; the content is §9.4.** One character, in the paragraph a partner reads on the way to the metadata document.

## Consistency edits, beyond the listed amendments

Small, and called out so a reviewer can reject them separately:

- §3.5's "three additions" became "five" — the list is now five items.
- §3.6's heading became "Three rules", since P3-a adds one.
- §7 phase 0 records D4's choice; phase 1 gains a paragraph saying merging is not deploying.
- §10 questions 1 and 4 are marked ruled rather than deleted; 2 and 3 gain what P2-b changed about them.
- §11 gains #523, #522, #567, #568 and the council ruling.
- **P1-b gained a third state.** The ruling names two, unset and set. The spec also says a non-canonical value refuses to boot, which comes from the phase 1 build brief rather than the amendment. QA then found the first version of that state too narrow, which is the first review item above.

## Ran

Re-run at `9ddb21a`:

- `uv run python scripts/check_table_stakes.py --base origin/main` — `table stakes: clean (1 file(s) checked)`
- `uv run ruff check .` — `All checks passed!` (a no-op for a docs change, run because it gates CI)
- `uv run python -m pytest tests/ -q -p no:cacheprovider` — `3157 passed, 13 skipped, 1 xfailed, 2 warnings in 105.54s`

CI has no markdown linter (`.github/workflows/ci.yml` lint job is ruff, mypy, and the table-stakes script), so the table-stakes run is the prose gate. `dependency-audit` is red here as on every open PR — `browserslist` GHSA-c83g-rgw3-j3cx / GHSA-73wf-gq98-2v4g, fixed in unmerged #544, and this PR touches no lockfile.

The first commit's message was rewritten before review: it said the challenge URL avoids the request host on the grounds that the platform proxy rewrites Host, and `squash_merge_commit_message` is `COMMIT_MESSAGES`, so that sentence would have landed in `main`'s history with its correction three commits later.

## NOT done

- Nothing in this PR was verified against a running system, because there is nothing to run: it is a design document. The measurements it cites were run by QA and are attributed in the sections above.
- The `resource_documentation` 400 recorded in §9.3 is left undecided on purpose. It is an owner call, and it blocks the phase 1 deploy rather than this merge.

### DIFF
--- a/docs/specs/2026-08-16-mcp-authorization.md
+++ b/docs/specs/2026-08-16-mcp-authorization.md
@@ -4,11 +4,14 @@
 
 | | |
 |---|---|
-| Closes | [#290](https://github.com/aks129/HealthClawGuardrails/issues/290) |
+| Closes | phase 1 closes [#523](https://github.com/aks129/HealthClawGuardrails/issues/523) |
+| Also open | [#290](https://github.com/aks129/HealthClawGuardrails/issues/290) closes on the §8.4 end-user run. [#568](https://github.com/aks129/HealthClawGuardrails/issues/568) tracks what remains after it |
+| Canonical resource | `https://mcp.healthclaw.io/mcp` (council ruling D4, `docs/2026-09-02-council-ruling.md`) |
+| Issuer | `https://app.healthclaw.io` (same ruling) |
 | Feature set | 6 — Surfaces (`docs/prd/06-surfaces.md` §6 names this as the missing spec) |
 | Pipeline step | 3, architecture review (`docs/2026-08-16-delivery-process.md`) |
 | Author | owner-surfaces |
-| Decision required from | the founder, on §3.1 (the canonical resource URI) and §7 phase 3 (deploy authorization) |
+| Decision required from | the founder, on §7 phase 3 (deploy authorization). §3.1 is settled by D4 |
 
 Everything asserted about live behaviour in this document was executed on
 2026-08-16/17 against the running deployments. The raw responses are quoted
@@ -133,10 +136,34 @@ hostname baked into a token audience and a partner's saved config is a
 migration we would have to run later under worse conditions. The dangling
 record is separately a risk and is logged in §9.6.
 
-This is the founder's call, and it blocks phase 1. The chosen value is pinned
-in one place, `MCP_CANONICAL_RESOURCE`, and the server refuses to start with
-OAuth acceptance enabled and that variable unset — the same fail-closed shape
-as `MCP_AUTH_TOKEN`.
+This was the founder's call and it blocked phase 1. Council ruling D4 settled
+it: `https://mcp.healthclaw.io/mcp`. The chosen value is pinned in one place,
+`MCP_CANONICAL_RESOURCE`.
+
+**The flag rule, replacing the fail-closed sentence this paragraph used to
+carry (amendment P1-b).** The old rule was that the server refuses to start
+with OAuth acceptance enabled and the variable unset. That rule cannot work
+for phase 1, because phase 1 serves the metadata before OAuth acceptance
+exists to be enabled. There is nothing for the boot check to key on. The rule
+is now:
+
+- **Unset.** The two well-known routes return 404 and the challenge stays a
+  bare `Bearer`. That is today's behaviour, unchanged, and it is what merges
+  before DNS moves.
+- **Set.** Both routes serve the document and the challenge carries
+  `resource_metadata` and `scope`.
+- **Set, but malformed** — not absolute, or not `https:`. The server refuses
+  to start. A value that cannot be an audience would otherwise turn phase 1
+  off in silence, which looks the same from outside as a deploy that worked.
+- **Host gate.** The document and the enriched challenge are served only when
+  the request `Host` equals the canonical host. The Railway hostname keeps
+  `/health` and the static-token path exactly as they are.
+
+The host gate is not defence in depth, it is conformance. RFC 9728 §3.3 has a
+client reject a document whose `resource` is not the identifier it inserted
+into the well-known path. Serving our document on the platform hostname would
+hand that client a rejection where it currently gets a 404, and the rejection
+reads to a partner as our bug.
 
 Throughout this document `<RESOURCE>` stands for the chosen value.
 
@@ -151,7 +178,7 @@ route and removes a whole class of client disagreement:
   resource whose path is `/mcp`)
 - `/.well-known/oauth-protected-resource` (root fallback)
 
-Both return, byte-identically:
+Both return the same document:
 
 ```json
 {
@@ -164,11 +191,34 @@ Both return, byte-identically:
 }
 ```
 
+**Serving the same bytes at the root path is not conformant, and we do it
+anyway (amendment P1-d).** Under RFC 9728 §3.3, root-form insertion implies
+the resource identifier `https://mcp.healthclaw.io`. Our document says
+`https://mcp.healthclaw.io/mcp`. A client that fetched the root form and
+checked would be right to reject it. MCP 2026-07-28 lists the root form as the
+fallback to try second, so a client that reaches it has already missed the
+sub-path form, and a 404 there helps nobody. We keep serving both. The
+consequence for the test plan is §8.1: A2 and A3 assert against the sub-path
+URL the challenge points at, never the root.
+
 Notes on each field, because each one has a way to be wrong:
 
 - `resource` is REQUIRED by RFC 9728 and must string-equal `<RESOURCE>`. It is
   not derived from `request.host` — a proxy that rewrites Host would then
   silently mint a different identity. It is read from the pinned constant.
+  **The same rule binds the `resource_metadata` URL in the 401 (amendment
+  P1-a).** That URL is built from `MCP_CANONICAL_RESOURCE` and never from
+  `req.hostname`. The reason is not that a proxy is known to rewrite Host.
+  Railway preserves it: production answers with
+  `fullUrl: http://app.healthclaw.io/r6/fhir/Condition/…`, built from
+  `request.host_url` (`r6/routes.py:1128`), and that is the public custom
+  domain. Measured 2026-09-03 against `GET /r6/fhir/Condition?_count=1`.
+  The reason is RFC 9728 §3.3. A client rejects a document whose `resource` is
+  not the identifier it inserted into the well-known path. What we advertise
+  therefore comes from the constant, whatever the request carries.
+  Building a URL from the request is not a hypothetical cost. The same
+  measurement shows every production authorization-server URL as `http://`,
+  because `request.host_url` loses the scheme behind the proxy (§3.3, #567).
 - `authorization_servers` is what makes the document useful; RFC 9728 marks it
   OPTIONAL but the MCP specification requires at least one entry. Its value is
   an **issuer identifier**, and §3.3 exists entirely because the issuer we
@@ -185,7 +235,7 @@ document contains no secret, identifies no tenant, and grants nothing. Its
 entire content is already public knowledge or is a URL. Refusing to serve it
 is what produces the dead end.
 
-They also need CORS that the transport endpoint does not have — see §8.4.
+They also need CORS that the transport endpoint does not have — see §9.4.
 
 ### 3.3 The authorization server must become discoverable from its own issuer
 
@@ -228,7 +278,8 @@ And its contents, fetched live, fail validation twice over:
 
 The cause of the scheme is not an OAuth bug: `oauth_discovery` builds every
 URL from `request.host_url`, and behind Railway's TLS-terminating proxy that
-yields `http`. This is the same class of trap CLAUDE.md already records about
+yields `http`. Re-measured on 2026-09-03 and filed as #567: the proxy loses
+the scheme, and preserves Host. This is the same class of trap CLAUDE.md already records about
 Railway and ports.
 
 **Three required changes, all in `r6/oauth.py` and its config:**
@@ -264,8 +315,14 @@ field.
 ### 3.4 What changes in the 401
 
 Two cases, and conflating them is how #290's own "option 1" sketch got it
-wrong. RFC 6750 does not permit an `error` parameter when no credential was
-presented, because there is no token to call invalid.
+wrong. RFC 6750 §3.1 says a server SHOULD NOT include `error` when no
+credential was presented, because there is no token to call invalid. This
+document said "does not permit", which overstates the requirement; the
+behaviour it asks for is unchanged (amendment P1-c).
+
+`<RESOURCE_ROOT>` below is the origin of `MCP_CANONICAL_RESOURCE`, and the
+whole URL is built from that constant. Never from `req.hostname` — see §3.2,
+amendment P1-a.
 
 **No credential presented** — the ordinary first contact:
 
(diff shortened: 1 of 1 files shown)


aks129:
## QA review — PASS on the amendments, FAIL on three descriptions of code. Fixed in `9ddb21a`

This document's job in this PR is to describe the code truthfully. The amendments themselves land as ruled, and the fact-check in the PR body is honest — I re-ran all seven claims plus the two the reading added, and they hold. Three other places describe code that is not there.

### The seven Flask claims, re-checked line by line

Read against the source at `origin/main`, independently of the PR body's own reading.

| Claim | Cited | What the source says | |
|---|---|---|---|
| `/oauth/authorize` returns the code in a JSON body, no `302`, no `iss` | `r6/oauth.py:307-313` | `return jsonify({'redirect': location, 'code': code, 'state': state})` at 309-313. No `redirect(...)`, no `iss`. | holds |
| tenant comes from `X-Tenant-Id`; 403 limits it to public tenants | `:284`, `:286` | 284 is `requested_tenant = request.headers.get('X-Tenant-Id', 'default')`; 286 is `if read_auth_enabled() and not is_public(requested_tenant):` | holds, exactly |
| no `resource` parameter read at either endpoint | no occurrence | the only `resource` hits are scope descriptions and `resourceType` in an OperationOutcome | holds |
| no introspection endpoint; `token` never checks a `client_secret` | no occurrence | zero `introspect`; `client_secret` appears only in the store and in `/oauth/register` | holds |
| `OAUTH_TOKEN_TTL` defaults to 3600, no `refresh_token` ever issued | `:33`, `:371-375` | 33 is `TOKEN_TTL_SECONDS = int(os.environ.get('OAUTH_TOKEN_TTL', '3600'))`; the token response is 371-376 and carries no refresh token; zero `refresh_token` in the file | holds (range ends 376, not 375) |
| `_oauth_authorizes` checks tenant and scope, not audience | `r6/read_auth.py:35-41` | `def` at 35, `return` at 41; checks `tenant_id` and scope intersection; zero `aud` in the file | holds, exactly |
| `APP_ENV`, `READ_AUTH_ENABLED`, `PUBLIC_TENANTS` are real knobs | `runtime_config.py:49`, `:105`, `access.py:95` | 49 `if "APP_ENV" in env:`; 105 `if env.get("READ_AUTH_ENABLED", …)`; 95 `def is_public(tenant_id: str) -> bool:` | holds, exactly |

The two details the reading added also hold: `/oauth/register` answers `'token_endpoint_auth_method': 'client_secret_post'` at `:243` while discovery advertises `['client_secret_post', 'none']` at `:186`; `client_id` at `/oauth/token` is `if client_id and auth_code['client_id'] != client_id:` at `:343`. So does `r6/routes.py:1128`, the `request.host_url` citation behind the Host measurement.

**No commit message or PR body on either branch still asserts that the platform rewrites Host.** Grepped both branches' logs and both file trees. Every surviving occurrence is either the explicit refutation or hypothetical framing ("a proxy that rewrites Host *would* then…"), which is the RFC argument, not a claim about Railway.

### FAIL 1 — §9.4 carried two line numbers into the MCP server

`src/index.ts:97` and `:131` are correct on `main`. Phase 1 inserts 148 lines above both, so on the merge of #556 they become 193 and 227. That is the same staleness this PR fixed for P3-a — caught there because the amendment touched it, missed here because the amendment did not. Replaced with the ordering, which is the fact §9.4 actually needs.

### FAIL 2 — §9.4 said the 401 CORS was required in phase 1, and phase 1 does not do it

"**Required in phase 1:** the two PRM paths **and the 401 response** must carry `Access-Control-Allow-Origin: *`". #556 does the PRM paths and not the 401, and discloses that as part 2 of #523.

Deferring is right, and for a stronger reason than scope. Measured against the phase 1 build, 2026-09-03:

```
OPTIONS /mcp   Origin: https://claude.ai
               Access-Control-Request-Method: POST
               Access-Control-Request-Headers: content-type
  -> 204, and NO access-control-allow-origin
```

A browser's `POST /mcp` with a JSON content type is preflighted, and the preflight is refused before a 401 exists to read a header on. **`Access-Control-Allow-Origin` on the 401 alone would look like the fix and not be one** — the `docs/2026-08-02-retro.md` shape, a control that appears to do one thing and quietly does another. The refusal and its preflight have to move together.

The requirement stands. What changed: it is no longer labelled phase 1's, it says why the two must move together, and §9.4 now says plainly that phase 1's reachability claim covers a **server-side fetcher, not a browser-context one**. The PRM document itself *is* browser-readable — it is the challenge that points at it that is not, and that distinction was not in the document.

### FAIL 3 — §3.1's malformed rule described three of twelve shapes

"**Set, but malformed** — not absolute, or not `https:`." I measured what actually boots. Nine more shapes parse as https URLs and start a server whose own challenge advertises a document its reader must refuse under RFC 9728 §3.3 — mixed-case scheme or host, explicit `:443`, `?query`, `#fragment`, `user:pass@` userinfo, a bare trailing slash, dot segments, and an empty authority. Full table on #556.

`user:pass@` is the one to look at twice: the configured string is served verbatim as `resource`, in a document that is unauthenticated and carries `Access-Control-Allow-Origin: *`.

§3.1 now states the rule as the property the check enforces — the value must already be the identifier a client derives — with the shapes named. #556's `b11a359` implements it; every value that boots now agrees under §3.3.

### Also corrected

- **§3.1's Host gate said "equals the canonical host".** The comparison is on hostname: port, case and userinfo in the header are not part of it, and `X-Forwarded-Host` is not read. Measured across ten Host spellings plus duplicate headers and an absolute-form request target — nothing reaches the document from the platform hostname, and nothing changes what is advertised, so the leniency is cosmetic. But "equals" is not what the code does and a reader would design against it.
- **§9.3 gains a fifth row, which is ours and is measured.** §3.2's document points a partner at `https://app.healthclaw.io/r6/fhir/docs/privacy-policy`. Fetched unauthenticated, live, 2026-09-03: `400`, `{"code":"security","diagnostics":"X-Tenant-Id header is required"}`. The tenant gate stands in front of the one human-readable pointer in a document whose reader has no tenant yet — §9.3 in miniature, in the document that exists to stop a partner hitting a dead end. Recorded with three options (exempt the route, repoint the field, drop it — RFC 9728 §2 marks it OPTIONAL) and **left undecided**: it is a product call, not QA's, and it belongs in the pre-deploy checks beside the four curls.

### What I did not change

The amendments as ruled, P1-a..d, P2-a..f and P3-a — every one lands where the ruling put it, and the two that correct the document rather than the design say so. §3.2's P1-d paragraph is right that serving the root form is deliberate non-conformance; I confirmed the served root-form document carries `…/mcp` while root-form insertion implies `https://mcp.healthclaw.io`, and that A2/A3 assert against the sub-path. The `<RESOURCE>` and issuer values match D4. #568 is cited in all four promised places.

Each of my edits is separable — reject any one without touching the others.

### Ran

```
uv run python scripts/check_table_stakes.py --base origin/main
  table stakes: clean (1 file(s) checked)
uv run ruff check .
  All checks passed!
  3157 passed, 13 skipped, 1 xfailed, 2 warnings in 104.36s
```

`dependency-audit` is red here from npm advisories fixed in unmerged #544 — the same job is red on #556 and on every open PR, and this branch touches no lockfile.

### What I did NOT cover

- **Nothing in this document was verified against a running system except the two live measurements above** (the privacy-policy 400, and the preflight against a local phase 1 build). The phase 2 and phase 3 claims are checks of source, not of behaviour — same standard the PR body sets for itself.
- **I did not re-run the §3.3 discovery-chain measurements** (the three 404/400s, the `http://` issuer). They are dated 2026-08-16/17 and I took them as given.
- **No RFC conformance opinion on P2-a..f beyond reading them**, since none is built.
- I did not merge, arm auto-merge, or run `gh pr checks`; `autoMergeRequest` was null before and after my push.

### Verdict

**PASS on the amendments. FAIL on three descriptions of the code, fixed in `9ddb21a`.** A design document whose whole purpose is that people cite it instead of reading the code has to be right about the code; two of the three would have been read as instructions and one of them ("required in phase 1") would have been read as a thing that shipped.

Not merging — a maintainer owns that gate.

Issue #226, opened 2026-08-01
integrations: DocumentReferences are ingested and counted but no tool can read them


aks129:
MEDENT ingest brings in DocumentReferences and `healthclaw.py COUNTED_TYPES` counts them — so the patient is told '247 records synced' while the agent **cannot open the clinical notes that dominate that count**.

`careagents/agent.py` `search_records` enum omits `DocumentReference` entirely (Condition, Observation, MedicationRequest, AllergyIntolerance, Immunization, Procedure only).

Notes and discharge summaries are the substance of a real record. The MCP server already proves the read path works (35 resource types incl. DocumentReference), so this is the highest value-to-effort item in the integrations set — days, not weeks.

**Value gate before building:** sample what the DocumentReference rows actually contain. If they carry only external URLs rather than readable text or an inline attachment, the tool cannot help and we should say so instead of shipping it.

From the 2026-08-01 integrations audit.

aks129:
## Gate G2 result: blocked — do not build this yet

Measured against the live demo tenant on 2026-08-01. DocumentReference attachments come back **empty**:

```
DOCREF: {'empty': 2}   # neither data nor url present
```

**Root cause:** `r6/redaction.py::_redact_recursive` strips `data`, `url` **and** `title` from any object that has a `contentType` — which is every attachment. So the document body never survives the redacted read path.

A `get_documents` tool built today would return a list of documents the agent cannot open. That is worse than not having it: the person is told notes exist and the agent still cannot help.

**What has to be decided first:** how a patient reads *their own* clinical note. Options, roughly in order of effort:

1. A patient-controlled read path that preserves attachment content for the record's own subject (analogous to `apply_patient_controlled_redaction`, which already preserves birthDate on the same reasoning).
2. Server-side extraction into a redacted summary, so the note is summarised rather than handed over whole.
3. Leave documents out of the agent surface entirely and say so plainly in the UI, rather than counting them in "247 records synced".

Option 3 is the honest stopgap and costs almost nothing — see #225 for the counting-what-we-can't-read problem.

Note the free-text risk is real here and cuts the other way from the code-label fix (#207 / PR #240): a clinical note is *full* of PHI by nature, so none of this can be solved by a lookup table. It needs a deliberate decision about the read path.

Found by the 2026-08-01 integrations audit; measured as gate G2 in `docs/2026-08-01-webinar-delivery-plan.md`.

aks129:
[...]

PR #561, opened 2026-09-03
candidate: stop counting DocumentReferences nothing can open (#226)

aks129:
## What

DocumentReferences leave the synced count, and where the count is shown it says so.

- **`careagents/healthclaw.py:459`** — `DocumentReference` moves out of `COUNTED_TYPES` into a new `UNCOUNTED_TYPES`, with `uncounted_record_count()` as the source for the sentence. It **raises** on failure on the same terms as `record_count`: a `0` there reads as "no documents arrived", which is not the same fact as "could not ask" (#403).
- **`careagents/models.py:87`** — a new `last_uncounted` baseline on `ca_connections`, added through the existing inline `_ensure_columns` pattern. Plain `ADD COLUMN ... INTEGER`, which SQLite and Postgres both accept. NULL means "no baseline recorded", which is never read as an arrival.
- **`careagents/accounts.py:359`** — `mark_synced` takes an optional `uncounted` and leaves the stored figure alone when it is None: a caller that could not measure documents must not erase the last one that did.
- **`careagents/app.py:639`** — both refresh/upload call sites baseline documents; the poll emits one of three sentences, or none.
- **`careagents/static/home.js:471`** — four outcomes, and the interim ones do not stop the poller.

**Ingest is unchanged.** The DocumentReferences are still fetched and still stored. This PR changes what the number *claims*, not what is kept — the day a read path lands, the type moves back into `COUNTED_TYPES`.

## Why

Closes the counting half of #226 (council ruling **D2**, clinical — **option 3** from the issue's own list).

MEDENT ingest brings in DocumentReferences and `COUNTED_TYPES` counted them, so the patient was told "N records synced" while the notes and discharge summaries dominating that number could not be opened by anything:

- `careagents/agent.py` `search_records` omits `DocumentReference` from its resource enum, so the agent cannot ask for one.
- Gate **G2**, measured against the live demo tenant on 2026-08-01, found the attachments come back **empty** regardless — `r6/redaction.py::_redact_recursive` strips `data`, `url` **and** `title` from any object carrying a `contentType`, which is every attachment.

So the number counted rows nothing could read — the retro's recurring shape (`docs/2026-08-02-retro.md`): a measure of one thing presented as a measure of another. G2's own stated exit was *"the tool cannot help; drop it and say so"* (`docs/2026-08-01-webinar-delivery-plan.md:51`). This is that.

### What the patient sees — the four outcomes

| after a refresh | rendered |
| --- | --- |
| readable records arrived, tenant holds documents | `5 new records added. Notes and documents are not yet readable here.` |
| readable records arrived, no documents | `5 new records added.` |
| **only documents arrived** | `No new records you can read. Notes and documents arrived, and they are not readable here yet.` |
| nothing arrived | *(nothing — the sync message is left as it was)* |
| the document probe failed | `5 new records added. We could not check whether notes or documents were left out.` |

### Two rulings this went through, and why the second one moved the scope

**Hedge, don't suppress.** An earlier revision dropped the *whole count* when the document probe raised. That trades one silence for another — the growth is a fact we did measure — so the count now stands with a sentence naming what could not be checked. Known part stated, unknown named, the posture the review relay already takes on an unestablished `confirmed`.

**The fourth outcome needed a measurement that did not exist.** `home.js` gated its entire render on `new_records > 0`, so a refresh delivering only DocumentReferences showed **nothing** — indistinguishable from a sync that did nothing, and worse than the misleading count this PR set out to remove.

Saying so honestly turned out to require a schema change. `uncounted_record_count` is a tenant **total**, so "documents arrived" could not be established from it; and a clause gated on `uncounted > 0` would have fired at the first 5s tick of *every* refresh, including every no-op, for exactly the tenants that already hold notes. Announcing an arrival from a number that cannot distinguish arrival from pre-existing would have put this PR's own defect class inside the fix for it.

So the refresh now baselines documents the way it already baselines readable records, and the arrival sentence is gated on the delta. That is a column on `ca_connections`, which is wider than this PR's original "changes what the number claims, not what is stored" scope — **authorized deliberately**, because the measurement showed the narrower scope could not produce an honest answer.

### The first refresh for an account that already holds documents

The baseline is recorded **by the refresh itself**, before re-authorization, so a delta is available from the very first refresh after this ships: an account holding four notes baselines at 4, and only notes arriving *after* that are reported. Nothing already on file is ever announced as newly arrived.

A connection whose row predates the column has `last_uncounted = NULL`. That is treated as "arrival unknown", not zero, so no arrival is claimed until a refresh records a baseline. Both are pinned (`test_documents_that_predate_the_baseline_are_not_called_new`, `test_a_connection_with_no_document_baseline_claims_no_arrival`).

## Ran

```
uv run ruff check .
All checks passed!

3178 passed, 13 skipped, 1 xfailed, 2 warnings in 121.85s (0:02:01)
```

In `e2e/`, against the app booted by the harness:

```
npx playwright test careagents-sync-message --project=chromium
7 passed (47.6s)
```

Green including `tests/test_ratchets.py` and `tests/test_guardrail_conformance.py`. No TypeScript touched, so the orchestrator's npm lane is unaffected by this branch.

**The branch selection is verified in a browser, not by inspection.** `e2e/tests/careagents-sync-message.spec.ts` serves a fixture page carrying only the markup `home.js` touches, loads the **real shipped file** from `/static/home.js`, and fulfils both requests the journey makes (the refresh POST and the poll GET) with `page.route`. No backend, no HealthClaw, and no re-implementation of the rule in another language. All five outcomes above run there, plus the upgrade case: a document-only message followed by a poll carrying readable records replaces it, because `clearInterval` fires only on a readable record.

Mutations run and confirmed red, then restored:

| mutation | result |
| --- | --- |
| gate the arrival clause on `uncounted > 0` instead of the delta | 3 red (`..._stays_quiet`, `..._predate_the_baseline...`, `..._no_document_baseline...`) |
| restore `if (d.new_records === 0) return;` in home.js | 2 red **in the browser suite** — the exact defect, and nothing in the Python suite catches it |
| delete the `last_uncounted` block from `_ensure_columns` | 1 red (`..._added_to_an_existing_table`) |

The "nothing arrived stays quiet" test asserts the poll **actually fired** before concluding it stayed silent — otherwise it would pass on a route that never ran.

## What was NOT run

- **No read path for documents.** Options 1 and 2 from the issue (patient-controlled attachment preservation; server-side extraction into a redacted summary) are untouched, and `search_records` still omits `DocumentReference`. This is only the honest stopgap — **#226 should stay open** for the read-path decision.
- **The Postgres lane was not run locally** — no `psql` and no Docker daemon on this machine. The migration is exercised against SQLite (a table built in the pre-#226 shape, then migrated) and the ALTER is plain `ADD COLUMN ... INTEGER`, valid on both. **CI's Postgres lane is the check that matters here**; a reviewer should confirm it went green before merging, since a column that appears only on the local lane is precisely the failure this note exists for.
- **Only the sync-message spec was run in the browser**, not the whole Playwright suite. The retro flags this gate's `webServer` boot as fragile; the other specs were left alone rather than fought with.
- **The endpoint tests use `FakeClient`**, so per the repo's standing trap they prove the call is made and the sentence is wired — *not* that the engine answers. Mitigating: the Flask route already served `_summary=count` for `DocumentReference` (it was in the counted set until this commit), so no new engine surface is introduced. The unit tests cover the sum logic against the real client.
- **Not run against a live stack.** No `flask init-db` / seeded-tenant walkthrough.
- No change to redaction, audit, tenant scoping, or any write path.
- **This does add a schema column** (`ca_connections.last_uncounted`) — see the scope note above. It is a nullable integer holding a count; no patient data, nothing about a record's content, and CareAgents' no-PHI rule is untouched.

## Sweeps, so nobody repeats them

- `grep -rn "records synced" docs/ careagents/` — **no hits.** No doc or template quotes the wording; the issue's "247 records synced" is the author's paraphrase of the UI, which actually says "N new records added."
- `grep -rn "last_count\|record_count" careagents/` — the only patient-visible surface fed by `record_count` is this poll. `last_count` is a refresh baseline and never reaches a template.
- `grep -rn "def record_count" careagents/ scripts/ tests/` — only `HealthClawClient` and the test `FakeClient` implement it, so no third client can reach the poll's new call without the method. `mark_synced`'s new argument is optional, so no caller breaks.
- The greeting totals (`careagents/app.py:~720`, conditions / medications / labs) never included `DocumentReference` and are unchanged.
- `docs/2026-08-01-webinar-delivery-plan.md` mentions `DocumentReference` at lines 34 and 51; neither claims documents are in the synced count, so nothing there went stale.

## Cost

One extra `_summary=count` request per poll tick after a refresh (every 5s, capped at 60 ticks by `watchForNewRecords`), plus one more at each refresh to record the baseline. Negligible, but they are real added calls.

Also: a document-only or probe-failed message no longer ends the watch, so those polls now run to the 60-tick ceiling instead of stopping early. That is the point — the message upgrades when readable records land — but it is more polls than before on those paths.

### DIFF
--- a/careagents/accounts.py
+++ b/careagents/accounts.py
@@ -356,12 +356,18 @@ class AccountService:
                  .filter_by(id=conn_id, account_id=account_id).first())
             return _conn_dict(c) if c else None
 
-    def mark_synced(self, conn_id: str, count: int) -> dict:
+    def mark_synced(self, conn_id: str, count: int,
+                    uncounted: int | None = None) -> dict:
         """Record the end of a sync and report what changed since the last one.
 
         Returns {"new": int, "total": int} where `new` is the growth since the
         previous sync (0 on the first one, since there's no baseline to compare
         against and every record is arguably 'new').
+
+        `uncounted` baselines the types the count excludes (#226) on the same
+        terms. Passing None leaves the stored baseline untouched rather than
+        clearing it — a caller that could not measure documents must not erase
+        the last figure that was measured.
         """
         with self.session() as s:
             c = s.query(Connection).filter_by(id=conn_id).first()
@@ -369,6 +375,8 @@ class AccountService:
                 return {"new": 0, "total": count}
             previous = c.last_count
             c.last_count = count
+            if uncounted is not None:
+                c.last_uncounted = uncounted
             c.last_synced_at = now()
             new = 0 if previous is None else max(0, count - int(previous))
             return {"new": new, "total": count}
@@ -478,7 +486,8 @@ def _conn_dict(c: Connection) -> dict:
     return {"id": c.id, "kind": c.kind, "tenant_id": c.tenant_id,
             "label": c.label, "status": c.status, "provider": c.provider,
             "connected_at": c.connected_at,
-            "last_synced_at": c.last_synced_at, "last_count": c.last_count}
+            "last_synced_at": c.last_synced_at, "last_count": c.last_count,
+            "last_uncounted": c.last_uncounted}
 
 
 def _agent_dict(a: Agent) -> dict:
--- a/careagents/app.py
+++ b/careagents/app.py
@@ -636,24 +636,32 @@ def create_app(config: Config | None = None,
             # of decaying to zero while the patient is still reading it.
             baseline = conns[conn_tenant].get("last_count")
             if baseline is not None:
-                # The count covers what the patient can actually reach;
-                # DocumentReferences are ingested but unreadable, so they are
-                # out of it and the clause below says so (#226). The two are
-                # fetched together and reported together: a number whose
-                # "and there are notes missing from this" could not be
-                # established is the same silent omission the clause exists to
-                # end, so it is either both or neither.
                 try:
                     current = hc.record_count(conn_tenant)
+                except HealthClawError:
+                    current = None
+                # The count covers what the patient can actually reach;
+                # DocumentReferences are ingested but unreadable, so they are
+                # out of it and the sentence below says so (#226). When the
+                # probe itself fails we hedge rather than suppress: the count
+                # is a fact we did measure, and hiding it would trade one
+                # silence for another. State the known part, name the unknown
+                # — the posture the review relay already takes on an
+                # unestablished `confirmed`.
+                try:
                     uncounted = hc.uncounted_record_count(conn_tenant)
                 except HealthClawError:
-                    current = uncounted = None
+                    uncounted = None
                 if current is not None:
                     out["record_count"] = current
                     out["new_records"] = max(0, current - int(baseline))
-                    if uncounted:
+                    if uncounted is None:
                         out["uncounted_note"] = (
-                            "notes and documents are not yet readable here")
+                            "We could not check whether notes or documents "
+                            "were left out.")
+                    elif uncounted > 0:
+                        out["uncounted_note"] = (
+                            "Notes and documents are not yet readable here.")
             return jsonify(out)
         return jsonify({"status": "pending"})
 
--- a/careagents/app.py
+++ b/careagents/app.py
@@ -510,7 +510,8 @@ def create_app(config: Config | None = None,
                 logger.warning(
                     "could not flip connection %s to active", conn_id)
             try:
-                svc.mark_synced(conn_id, hc.record_count(conn["tenant_id"]))
+                svc.mark_synced(conn_id, hc.record_count(conn["tenant_id"]),
+                                hc.uncounted_record_count(conn["tenant_id"]))
             except HealthClawError:
                 logger.warning("record_count after upload failed for %s",
                                conn_id)
@@ -593,9 +594,12 @@ def create_app(config: Config | None = None,
                             "consent_version": CONSENT_VERSION}), 428
 
         # Baseline the count BEFORE re-authorizing so the follow-up poll can
-        # report what the refresh actually added.
+        # report what the refresh actually added — documents on the same
+        # terms, so "notes arrived" is a measured delta rather than a restatement
+        # of "this tenant holds notes" (#226).
         try:
-            svc.mark_synced(conn_id, hc.record_count(conn["tenant_id"]))
+            svc.mark_synced(conn_id, hc.record_count(conn["tenant_id"]),
+                            hc.uncounted_record_count(conn["tenant_id"]))
         except HealthClawError:
             return jsonify({"error": "records service unavailable"}), 503
 
@@ -653,15 +657,36 @@ def create_app(config: Config | None = None,
                 except HealthClawError:
                     uncounted = None
                 if current is not None:
+                    new_records = max(0, current - int(baseline))
                     out["record_count"] = current
-                    out["new_records"] = max(0, current - int(baseline))
+                    out["new_records"] = new_records
+                    # Documents that demonstrably arrived since the refresh
+                    # baselined them. None when the probe failed or the
+                    # connection predates the baseline column — in both cases
+                    # arrival is unknown, and unknown is never claimed.
+                    doc_baseline = conns[conn_tenant].get("last_uncounted")
+                    new_documents = (
+                        None if uncounted is None or doc_baseline is None
+                        else max(0, uncounted - int(doc_baseline)))
                     if uncounted is None:
                         out["uncounted_note"] = (
                             "We could not check whether notes or documents "
                             "were left out.")
-                    elif uncounted > 0:
+                    elif new_records > 0 and uncounted > 0:
+                        # A standing caveat on a number the patient can act
+                        # on: what you can read grew, and this excludes notes.
                         out["uncounted_note"] = (
                             "Notes and documents are not yet readable here.")
+                    elif new_records == 0 and new_documents:
+                        # The refresh delivered documents and nothing readable.
+                        # Silence here is indistinguishable from a sync that
+                        # did nothing, so say what happened. Gated on the
+                        # delta, never on `uncounted > 0` — the latter is true
+                        # from the first tick for every tenant that already
+                        # holds notes, and would fire on every no-op refresh.
+                        out["uncounted_note"] = (
+                            "Notes and documents arrived, and they are not "
+                            "readable here yet.")
             return jsonify(out)
         return jsonify({"status": "pending"})
 
--- a/careagents/app.py
+++ b/careagents/app.py
@@ -636,13 +636,24 @@ def create_app(config: Config | None = None,
             # of decaying to zero while the patient is still reading it.
             baseline = conns[conn_tenant].get("last_count")
             if baseline is not None:
+                # The count covers what the patient can actually reach;
+                # DocumentReferences are ingested but unreadable, so they are
+                # out of it and the clause below says so (#226). The two are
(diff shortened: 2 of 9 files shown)


aks129:
## QA review — **FAIL** (one MEDIUM)

The counted-set change is right and well pinned. It fails on the rendering: in the case #226 is most about — a delivery that is *only* documents — the sentence the server correctly emits is never shown, and the patient is told nothing at all.

Not approved on GitHub (an approval arms auto-merge here); `autoMergeRequest` confirmed `null` before and after my push.

### What I ran, on `cd2898b`

```
uv run ruff check .
All checks passed!

3164 passed, 13 skipped, 1 xfailed, 2 warnings in 104.11s

50 passed
```

Your numbers reproduce exactly. No TypeScript touched, so the npm lane is unaffected — agreed.

### 1. The hedge — correct, and it cannot be bypassed. Two mutations, both red

```
drop the `if uncounted is None:` branch, keep the elif
  -> FAILED test_the_count_is_hedged_not_hidden_when_the_document_probe_fails (KeyError: 'uncounted_note')
default a failed probe to 0 instead of None
  -> FAILED test_the_count_is_hedged_not_hidden_when_the_document_probe_fails
```

Also red: restoring `DocumentReference` to `COUNTED_TYPES` (2 tests), and swallowing the probe error to return 0. The tests are honest.

A probe failure **cannot** produce a bare number: `HealthClawError` is the only failure mode, and it is the one branch that always emits a note. The remaining silent-zero path (a 200 whose Bundle omits `total` → `int(None or 0)` → 0 → no sentence) needs upstream-proxy mode; in local mode `_summary=count` always sets `total`, which I confirmed live.

### 2. The number matches what a person can reach — verified live, real client, no fakes

Seeded a tenant on a running HealthClaw (`:5099`) with 25 Observation, 3 Condition, 2 Procedure, **4 DocumentReference**, 1 each MedicationRequest / AllergyIntolerance / Immunization. Then the real `HealthClawClient`:

```
record_count()            = 31        <- 3+25+1+1+1, documents excluded
uncounted_record_count()  = 4

agent can search Condition=3  Observation=25  MedicationRequest=1
                 AllergyIntolerance=1  Immunization=1  Procedure=2   (total 33)

DocumentReference after redaction:
  {'content': [{'attachment': {'contentType': 'text/plain'}}], ...}
```

Every counted type is in `careagents/agent.py`'s `search_records` enum, so **counted ⊆ readable**. G2 reproduced: the attachment comes back with `data`, `url` **and** `title` stripped — nothing can open one. The premise holds.

*(LOW: `Procedure` is readable but not counted, so the number under-reports by 2. Pre-existing, opposite direction, harmless.)*

### 3. Rendering — MEDIUM found here

Drove the real poll endpoint for every combination and applied the exact rule at `home.js:471-479`:

| server JSON | what the card shows |
|---|---|
| `new_records=5`, docs note | `5 new records added. Notes and documents are not yet readable here.` |
| `new_records=1`, docs note | `1 new record added. Notes and documents are not yet readable here.` |
| `new_records=5`, no note | `5 new records added.` |
| `new_records=0`, **docs note present** | **`<NOTHING RENDERED>`** |
| `new_records=0`, no note | `<NOTHING RENDERED>` |

The first three join cleanly — one space, one full stop, correct singular/plural. No fragment, no doubled stop.

---

#### MEDIUM — `careagents/static/home.js:471`: the caveat is unreachable when the delivery is documents only

**Scenario.** A patient refreshes a MEDENT connection. The pull delivers 12 discharge summaries and no new readable records. The server does its job — the poll returns `{"new_records": 0, "record_count": 100, "uncounted_note": "Notes and documents are not yet readable here."}` — and then `if (... d.new_records > 0)` discards it. The card keeps saying *"Finish signing in to your provider — new records appear here."* for five minutes and then stops quietly. **The patient is told nothing.**

Before this PR that same refresh reported **"12 new records added."** Misleading, which is why #226 exists — but an acknowledgement. This diff replaces it with silence, in the one case where documents dominate. That is the posture this PR's own description rejects: *"it trades one silence for another, and the growth is a fact we did measure."* G2's stated exit was *"the tool cannot help; drop it and say so"* — dropped, not said.

**Reproduction** (against this branch):
```python
d = _poll_after_sync(cfg, svc, monkeypatch, documents=12)   # counted unchanged
assert d["uncounted_note"]        # server emits it
assert d["new_records"] == 0      # home.js:471 gate discards it -> nothing renders
```

**I did not fix this.** The fix is one line — render `uncounted_note` whenever it is present, not only behind `new_records > 0` — but it needs a sentence for the zero-readable case, and the ruling enumerates exactly three outcomes. Inventing a fourth string is unrequested patient-facing copy and that call is Product's, not QA's. Naming the departure from "QA fixes MEDIUMs" rather than quietly writing copy.

---

#### LOW (pre-existing, not this PR) — the upload card still counts documents

`careagents/app.py` returns the engine's `ingested`; `home.js:369` renders it as `${ing} record(s) added`. That is a different number from `record_count`, so your `grep record_count careagents/` sweep could not have found it — the sweep was sound, the question was just wider than the source.

Measured on the live stack, uploading 5 Condition + 12 DocumentReference:

```
engine returned ingested=17 skipped=0 failed=0
UPLOAD CARD SHOWS ->  17 records added
record_count()        ->  5      (what the refresh path would now say)
uncounted_record_count() -> 12
```

This predates the branch and the diff does not worsen it. What the diff *introduces* is the inconsistency: two "records added" counters on the same page now mean different things and only one carries the caveat. Belongs on **#226**, which stays open anyway.

#### LOW — `uncounted_record_count` was missing from `EVERY_SEAM_METHOD`

`careagents/app.py` catches `HealthClawError` and only that, so a raw `requests` exception from the probe would 500 the poll and the patient would get neither number nor hedge. The guarantee held by construction (the probe delegates to `search`) and was exercised by nothing — the #287 ensemble gap, in the list whose stated contract is *"every method on the seam."* **Fixed** (below).

### Pushed to this branch — `e3f750a` (tests only, no production code)

- `test_healthclaw_transport.py` — added `uncounted_record_count` to `EVERY_SEAM_METHOD` and its dispatch, plus a wrongly-shaped-200 sibling. Dropping the `isinstance(body, dict)` arm from `_json_object` → **red** on the new rows. Confirmed.
- `test_careagents.py` — the clause shape asserted on the string the poll *actually emitted*, for both sentences, plus the rendered join. Removing the trailing period, or padding with a leading space, each → **red**. Confirmed. (This is what makes `home.js`'s `"... added." + " " + note` safe here and on the next surface to carry it.)

Re-ran after: ruff clean, **3171 passed, 13 skipped, 1 xfailed**, ratchets + conformance **50 passed**.

### Not covered

- **No browser.** The table above is the real poll JSON plus a hand port of `home.js:471-479` into Python. The rule matches by reading; it was not executed as JavaScript.
- No read path for documents assessed — correctly out of scope, and **#226 should stay open**.
- Redaction, audit, tenant scoping, write paths, migrations: untouched by the diff, not re-probed.
- The stale-baseline worry does not apply: `watchForNewRecords` runs only after `/refresh`, which re-baselines with the same new `record_count` seconds earlier. Checked, clean.

> `dependency-audit` is red on this PR and on every other open one — npm advisories fixed in unmerged **#544**. Not from this change.

Issue #219, opened 2026-08-01
reliability: thread saturation causes a self-inflicted restart loop that wipes all chats


aks129:
One worker, 8 threads (`deploy/careagents/Dockerfile`). Each chat turn holds a thread for its whole duration — up to 6 tool rounds × 25s HealthClaw timeout, plus a 90s LLM timeout.

**Failure sequence:** 8 concurrent turns block everything **including `/healthz`** → Railway marks the container unhealthy → restart → **all chat histories and login rate-limit state wiped mid-conversation** for every user.

**Bites at ~10-30 concurrent users.**

Mitigations, cheapest first: bound the turn (see the history issue), drop the HealthClaw read timeout to ~10s, raise to `--threads 16`. Persisting history removes the data-loss half of the failure.

From the 2026-08-01 architecture audit.

aks129:
**Probe result: reproduces differently.** Measurement only — no production code
changed, and no load was sent to `careagents.cloud`. PR #573 carries the rig
(`scripts/probe_219_thread_saturation.py`) and the transcript
(`docs/evidence/2026-09-03-probe-219-thread-saturation.md`).

**What reproduces.** `/healthz` starves — totally at 8 concurrent turns,
intermittently from 6. It never refuses a connection; it connects instantly and
waits for a free thread, so the wait tracks the turn length rather than N. At
the code's own worst case (`run_deadline_seconds` = 120) **`/healthz` waits
118.12s**. Shipping config, 8 threads as 2×4, 20s turns:

| N | turns started | 5s budget | 30s budget | true wait |
|---|---|---|---|---|
| 4 | 4/4 | 4/4 ok | 1/1 ok | 0.00s |
| 6 | 6/6 | 4/4 ok | 1/1 ok | 0.00s |
| 8 | 7/8 | 3/4 ok | 1/1 ok | **18.19s** |
| 10 | 8/10 | 1/4 ok | 1/1 ok | **18.13s** |
| 16 | 8/16 | 0/4 ok | 0/1 ok | **38.32s** |

Repeating N=6: across three attempts of six runs each, bursty and staggered
arrivals alike, **2, 4 and 4 of 6 stalled ~18s** and the rest ~0.01s. On six
trials that spread is noise, so read it as intermittent rather than as a rate;
what is stable is the bimodal shape. It is two gunicorn workers sharing an
accept socket — six turns land unevenly and saturate one while the other idles,
and the wait depends on which one accepted the probe.

**What does not.**

[...]

PR #562, opened 2026-09-03
candidate: $populate reads a bounded %patient projection, redacted content, 24h links (council D10)

aks129:
## What

`$populate` no longer reads whatever a Questionnaire asks for.

- **Expressions evaluate against a bounded projection**, not the stored Patient. `build_context` puts a new dict in the FHIRPath environment holding only `name.given`, `name.family`, `birthDate`, `gender`, `telecom` (phone and email), and `address` (line, city, state, postalCode). `%patient.identifier` and `%patient.photo` resolve to nothing, and nothing in the code inspects the expression text to make that happen. There is deliberately no `%resources`.
- **Auto-loaded clinical content goes through `apply_redaction`** before the engine sees it, then labels come from `r6/terminology.py` keyed by code.
- **Tombstoned rows are not read.** Both queries in the route filter `is_deleted=False`.
- **An item withheld by the projection reports an issue naming its `linkId`**, so an empty answer that means "we refused" is distinguishable from one that means "the record has no such value".
- **The delivery PDF link drops from 7 days to 24 hours.**

## Why

Council ruling 2026-09-02 D10. Three seats — interop, privacy and clinical — independently called this a live violation of the non-negotiables. It was: an expression reached the whole stored Patient, and the tenant's Observations, MedicationRequests, AllergyIntolerances and Conditions were loaded verbatim with upstream `display` and `CodeableConcept.text` intact. Real feeds put patient names in those fields, which is the one thing `CLAUDE.md` says never to preserve.

A projection rather than a filter, because a denylist over FHIRPath source is one `where()`, one `descendants()`, one alias away from being wrong, and it fails open. A projection fails closed by construction.

Redaction before population rather than over the response, because the engine copies free text into answers from four resolvers and from `Observation.valueCodeableConcept`. A later pass could not tell an answer that came from a record from one the caller typed.

## What this does not buy

Stated plainly so nobody reads more into it than is there.

- Exiting through the kernel with `Profile.INTAKE` makes the operation a counted, shaped FHIR exit. INTAKE's own strip pops top-level `note`, `text` and SSN-class identifiers and does not recurse, so on this `Parameters` wrapper it changes nothing. The projection and the redaction are what bound the payload.
- The two queries stay in the route, so the raw-tenant-read count is unchanged rather than reduced. The kernel has no resource selector yet, which is playbook F5 and unlanded.
- The subject itself is not redacted. An intake form exists to carry the patient's own name, date of birth and address, and redaction would truncate all three. The projection is what bounds it.

## The ruled trade — CORRECTED

The original wording here said this affected "SNOMED allergens, today". That was measured and it is far too narrow. On a seeded synthetic patient with two allergies, one condition and one medication, **seven of eight clinical fields come back empty**: every allergen, every allergy reaction, every SNOMED-coded condition, and every medication dose. Only the medication name populates, because its code happens to be in the table.

`r6/terminology.py` holds 121 labels — 49 LOINC, 23 ICD-10, 20 RxNorm, 15 ICD-9, 14 CVX and **zero SNOMED** — and the terminology lookup flag appears in no deploy configuration. Since allergies, reactions and conditions are SNOMED-coded in practice, redacting auto-loaded content without a label for the code empties them.

Worse, the response carried **no signal at all** that anything was dropped. A caller reading it could not distinguish "this patient has no allergies recorded" from "we read three and could not name any of them".

The CTO ruling on this PR settles what to do, and the fix ships with this change rather than after it, because this change causes the regression. Leaving the corrected claim here rather than quietly rewriting it: a claim in a PR body is product surface, and this one was wrong.

## Ratchets

| Pin | Before | After |
|---|---|---|
| Soft-delete-blind files | 10 | 9 |
| Raw tenant reads | 5 | 5 |
| `r6/routes.py` lines | 3927 | 3927 |

## Outstanding from D10

The ruling's second half — pull `questionnaire_populate` from the model-facing tool tier for the beta — is **not** in this PR. Nothing in the demo scripts or e2e depends on the tool, so the ruling's own escape clause does not apply. But the reason for pulling it was that this path returned unredacted PHI, and that premise is what this PR removes. Cutting a tool is a user-visible capability change, so it goes back to the council rather than being decided in a code review.

## Ran

```
uv run ruff check .
All checks passed!

3178 passed, 13 skipped, 1 xfailed, 2 warnings in 105.60s (0:01:45)
```

Green including `tests/test_ratchets.py` and `tests/test_guardrail_conformance.py` (Grade A). Eleven new tests in `tests/test_sdc_populate_bounded.py` cover the negative cases the ruling names, a tombstoned Observation, a tombstoned Questionnaire, an upstream name in `code.text` asserted absent from the whole response, the unlabelled-allergen trade, inline content passthrough, and a PHI-free audit detail.

## What was NOT run

No live FHIR server and no real feed. No Postgres lane. The MCP host was not exercised, so the tool-tier question above is untested either way. Not deployed.

### DIFF
--- a/r6/sdc/delivery.py
+++ b/r6/sdc/delivery.py
@@ -26,8 +26,14 @@ from flask import Blueprint, Response, request
 from r6.sdc.documents import get_document_pdf_bytes
 from r6.audit import record_audit_event
 
-# One week — a patient/clinic link that is emailed and opened days later.
-DEFAULT_TTL_SECONDS = 7 * 24 * 3600  # 604800
+# 24 hours (council ruling D10, cut from one week). The link is a bearer
+# credential for a PDF of somebody's intake form: the signature in the URL is
+# the whole authorization, so anyone who ends up holding the URL — a
+# forwarded email, a shared screen, a mail archive — can open it for as long
+# as it lives. A day covers "sent this morning, opened this evening", which
+# is the case this exists for; a week was six extra days of exposure bought
+# for a case nobody measured.
+DEFAULT_TTL_SECONDS = 24 * 3600  # 86400
 
 
 def _link_secret():
--- a/r6/sdc/expressions.py
+++ b/r6/sdc/expressions.py
@@ -249,6 +249,14 @@ def resolves_outside_projection(expression):
     on a patient with no email) from being reported as withheld: it resolves
     against the probe on both sides, so it is not "outside".
 
+    WHAT THIS IS NOT: exhaustive. A withheld path `_PROBE_PATIENT` happens
+    not to populate — `%patient.deceasedDateTime`, `%patient.name.period`,
+    `%patient.telecom.rank` today — answers False, so that item comes back
+    silently empty instead of naming itself. That is a gap in the REPORTING,
+    never in the bound: the projection still withholds it, and nothing about
+    the patient reaches the caller either way. Widen the probe to close one;
+    do not widen it with anything but placeholders.
+
     Cached because the answer depends only on `expression`, which also caps
     the work a caller can buy per request.
     """
--- a/r6/sdc/expressions.py
+++ b/r6/sdc/expressions.py
@@ -3,6 +3,30 @@
 Thin wrapper over fhirpathpy. Evaluation failures return None rather than
 raising, so a single bad expression in a Questionnaire never aborts the
 whole populate/extract run (the caller records an issue instead).
+
+THE %patient PROJECTION — read this before widening `_PROJECTED_*` below.
+
+`build_context` does NOT put the stored Patient into the FHIRPath
+environment. It puts a NEW dict carrying only the elements an intake form is
+entitled to ask for (council ruling D10): name.given, name.family, birthDate,
+gender, telecom (phone/email), address (line/city/state/postalCode). An
+expression cannot reach what is not in the object, so `%patient.identifier`
+and `%patient.photo` resolve to nothing without this module ever inspecting
+the expression text.
+
+That is deliberate, and it is the whole design. A denylist over FHIRPath
+source — "reject an expression mentioning `identifier`" — is one `where()`,
+one `descendants()`, one alias away from being wrong, and it would fail open
+(an expression it did not understand would still run against the whole
+record). A projection fails closed by construction: the only way to widen
+what `%patient` can answer is to add an element to the allowlist here, in a
+diff a reviewer sees.
+
+There is deliberately no `%resources`. Handing a questionnaire author the
+tenant's clinical bundle as an environment variable is the unbounded read
+D10 names; the populate engine reaches auto-loaded content through its own
+code-matching and list-group paths, which are bounded by the questionnaire's
+own structure.
 """
 
 import logging
@@ -11,22 +35,132 @@ import fhirpathpy
 
 logger = logging.getLogger(__name__)
 
+#: telecom entries survive the projection only for these systems, and only
+#: `system` + `value` travel with them.
+_PROJECTED_TELECOM_SYSTEMS = ('phone', 'email')
 
-def build_context(subject=None, resources=None, extra=None):
+#: The only address elements an intake form may read.
+_PROJECTED_ADDRESS_KEYS = ('line', 'city', 'state', 'postalCode')
+
+#: Top-level scalars that survive whole.
+_PROJECTED_SCALARS = ('birthDate', 'gender')
+
+
+def patient_projection(subject):
+    """Return the bounded `%patient` object for `subject`, or None.
+
+    THE ONE PROPERTY: the returned dict is NEW, and every value in it was
+    copied element by element from an allowlisted path. No key of `subject`
+    reaches the result by default — `identifier`, `photo`, `contact`,
+    `name.text`, `extension` and everything else are absent because nothing
+    here copies them, not because something removed them.
+
+    `resourceType` is carried as a type tag, not as data: fhirpathpy raises
+    KeyError evaluating the resource-root form (`Patient.name.family`) against
+    a dict without it, and `evaluate` would swallow that as "no value", which
+    is a silent behaviour change for any Questionnaire not using `%patient`.
+    """
+    if not isinstance(subject, dict):
+        return None
+
+    projection = {'resourceType': 'Patient'}
+
+    names = []
+    for entry in subject.get('name') or []:
+        if not isinstance(entry, dict):
+            continue
+        kept = {}
+        given = [g for g in (entry.get('given') or []) if isinstance(g, str)]
+        if given:
+            kept['given'] = given
+        if isinstance(entry.get('family'), str):
+            kept['family'] = entry['family']
+        if kept:
+            names.append(kept)
+    if names:
+        projection['name'] = names
+
+    for key in _PROJECTED_SCALARS:
+        value = subject.get(key)
+        if isinstance(value, str):
+            projection[key] = value
+
+    telecom = []
+    for entry in subject.get('telecom') or []:
+        if not isinstance(entry, dict):
+            continue
+        if entry.get('system') not in _PROJECTED_TELECOM_SYSTEMS:
+            continue
+        if not isinstance(entry.get('value'), str):
+            continue
+        telecom.append({'system': entry['system'], 'value': entry['value']})
+    if telecom:
+        projection['telecom'] = telecom
+
+    addresses = []
+    for entry in subject.get('address') or []:
+        if not isinstance(entry, dict):
+            continue
+        kept = {}
+        for key in _PROJECTED_ADDRESS_KEYS:
+            value = entry.get(key)
+            if key == 'line':
+                lines = [ln for ln in (value or []) if isinstance(ln, str)]
+                if lines:
+                    kept['line'] = lines
+            elif isinstance(value, str):
+                kept[key] = value
+        if kept:
+            addresses.append(kept)
+    if addresses:
+        projection['address'] = addresses
+
+    return projection
+
+
+def build_context(subject=None):
     """Build the FHIRPath environment-variable context.
 
-    %patient / %subject resolve to the populate subject; named entries in
-    `extra` (e.g. launchContext or variable values) are passed through.
+    %patient and %subject both resolve to the BOUNDED projection of the
+    populate subject, and nothing else is in the environment. See the module
+    docstring for why this is a projection rather than a filter.
+
+    Both names share one projection object: FHIRPath evaluation reads its
+    environment, it never writes to it.
     """
-    context = {}
-    if subject is not None:
-        context['patient'] = subject
-        context['subject'] = subject
-    if resources:
-        context['resources'] = resources
-    if extra:
-        context.update(extra)
-    return context
+    projection = patient_projection(subject)
+    if projection is None:
+        return {}
+    return {'patient': projection, 'subject': projection}
+
+
+def resolves_outside_projection(expression, subject, resources=None):
+    """Would `expression` have resolved against the UNBOUNDED record?
+
+    THE ONE PROPERTY: this returns a bool. The value it evaluates is never
+    returned, logged, or stored — it exists only long enough to tell the two
+    reasons an item came back empty apart:
+
+      - the record simply has no such element  -> no answer, no issue
+      - the projection withheld it             -> no answer, and an issue
+        naming the linkId, so a caller learns the operation refused rather
+        than guessing the patient has no phone number
+
+    Call it ONLY after the bounded evaluation returned None; on its own it
+    says nothing about what the caller is allowed to see.
+    """
+    if not expression:
+        return False
+    if not isinstance(subject, dict) and not resources:
+        return False
+    unbounded = {
+        'patient': subject,
+        'subject': subject,
+        # The name the projection withholds entirely, present here so that an
+        # expression reaching for it is reported rather than silently empty.
+        'resources': list(resources or []),
+    }
+    return evaluate(expression, subject, unbounded) is not None
 
 
 def evaluate(expression, resource, context=None):
--- a/r6/sdc/expressions.py
+++ b/r6/sdc/expressions.py
@@ -27,9 +27,17 @@ tenant's clinical bundle as an environment variable is the unbounded read
 D10 names; the populate engine reaches auto-loaded content through its own
 code-matching and list-group paths, which are bounded by the questionnaire's
 own structure.
+
+AND NOTHING IN THIS MODULE MAY EVALUATE A CALLER'S EXPRESSION AGAINST THE
+REAL RECORD. Not to answer a yes/no question, not to decide whether to warn.
+`resolves_outside_projection` reads a constant probe for exactly that reason
+— its docstring has the reproduction. A bool derived from a patient's data
+and handed to the caller is a read of that data; repeat it once per
+questionnaire item and it is a fast one.
 """
 
 import logging
+from functools import lru_cache
 
 import fhirpathpy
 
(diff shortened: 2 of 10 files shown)


aks129:
## QA: FAIL on `d0b3971` — one HIGH, fixed in `ecd29fa`; two MEDIUMs stay open

Adversarial review against a running app (`PORT=5099`, seeded tenant), not the fakes.

**The good news first, because it is most of the PR.** The projection holds. I could not
reach `identifier`, `photo`, `name.text`, `contact`, `extension`, `maritalStatus` or an
`sms` telecom through any of 32 expressions, including the structural walkers —
`descendants()`, `children()`, `select($this)`, `repeat()`, `toString()`, `ofType()` — and
the resource-root form `Patient.identifier.value`. fhirpathpy's own env vars (`%context`,
`%resource`, `%rootResource`, `%ucum`) resolve to the projection or to a constant, never
to the record. Building a new dict instead of filtering an old one is the right call and
it did what the module docstring claims.

The redaction holds too. I planted a distinct fake patient name in `code.text`,
`coding.display`, `valueCodeableConcept.text`, `valueString`, `note[].text`,
`performer[].display` and `subject.display` on Observation, MedicationRequest,
AllergyIntolerance and Condition. **Zero** of those 31 markers appear anywhere in the
response, via either the code-matching or the list-group path. Audit is one AuditEvent,
`detail='populate; issues=1'`, PHI-free. The 10→9 soft-delete ratchet is real — both
`R6Resource.query` sites in `r6/sdc/routes.py` carry `is_deleted=False`, so the file drops
out of the scan; the pin was lowered to match a real change, not edited to hide one. Nine
of the new tests fail against `main`'s production code, which I checked by reverting it.

---

### HIGH — the refusal was a one-bit oracle over exactly the data the projection withholds

`r6/sdc/expressions.py:137-163` (`resolves_outside_projection`), reached from
`r6/sdc/populate.py:326`.

The function decides whether to emit the `WITHHELD_BY_PROJECTION` issue by evaluating
**the caller's own expression against the unbounded record** and reporting whether the
result was non-empty. That is a yes/no question the caller writes the wording of, asked of
the withheld data, and answered in a field the caller reads.

```
%patient.identifier.value.where($this.startsWith('1'))   -> issue names the linkId
%patient.identifier.value.where($this.startsWith('2'))   -> no issue
```

**Eleven HTTP requests recovered a stored SSN, character by character.** One request per
character; eleven items per request test all eleven candidates at once.

```
=== THE ORACLE: recover the withheld SSN over HTTP ===
  after  1 HTTP request(s): '1'
  ...
  after 11 HTTP request(s): '123-45-6789'
RECOVERED (never appeared in any answer): '123-45-6789'
```

`%resources` was a second door to the same record: `_gather_content` appends the
**unredacted** subject (`r6/sdc/routes.py:205`) and the oracle's environment carried it,
so `%resources.where(resourceType='Patient').identifier.value.where($this='123-45-6789')`
distinguishes a right guess from a wrong one just as well.

Everything else in the PR behaved correctly throughout this: the value never appeared in
an answer, the audit row stayed a count, and every assertion in
`tests/test_sdc_populate_bounded.py` passed. That file measures whether the *value* is in
the response. It never was.

**Same root cause, second symptom — an amplification DoS.** The oracle re-evaluates
against the full content once per unresolved item.
`%resources.descendants().descendants()`, 50 items over 500 stored resources:
**32.53 s of single-threaded CPU in one request**, against 0.04 s for the same shape with
no oracle call. ~800×, from a request body, on the thread pool #219 already worries about.

Reproduction (against a seeded running app):

```
uv run flask --app main init-db
PORT=5099 STEP_UP_SECRET=dev-secret PUBLIC_TENANTS=qa-tenant python main.py
# then POST Questionnaire/$populate with one item per guess:
#   %patient.identifier.value.where($this.startsWith('<prefix><char>'))
# and read which linkIds the OperationOutcome names.
```

**Fixed in `ecd29fa`.** `resolves_outside_projection(expression)` now takes **no record**.
It evaluates twice against a constant `_PROBE_PATIENT` — once through
`patient_projection`, once around it — and answers "outside" when the unbounded form
resolves and the bounded one does not. `%patient.identifier` is outside the allowlist for
everyone; that is a fact about the allowlist, so a constant can state it.

The ruling's UX is unchanged: the D10 negative list still produces an issue naming each
linkId, and an allowlisted path the patient simply lacks still stays silent (it resolves
on both sides of the probe). Live, after the fix: `RECOVERED: ''`, right and wrong guesses
indistinguishable, the three D10 refusals still named, the real intake Questionnaire
raises **0** issues. DoS shape: 32.53 s → 0.01 s.

Three tests in `tests/test_sdc_expressions.py:95-129` **pinned the oracle** — they asserted
`True` for `%patient.identifier.value` against the real patient, which is the leak. They
are rewritten to assert the allowlist property (nine withheld paths, eleven allowlisted),
not weakened. The route-level regression test fails against the pre-fix code with
`the issue list singled out ['M']`.

`da16068` follows up: the DoS test I first wrote asserted wall-clock time, but `lru_cache`
made 199 of 200 calls hits, so it measured the cache. Replaced with a signature pin —
adding a record parameter back goes red there before it goes red anywhere else.

---

### MEDIUM — D10 bullet 5 is not in this PR and is not deferred anywhere

The ruling has six bullets; the diff addresses five. Bullet 5 — *"Pull
`questionnaire_populate` from the model-facing read tier for the beta, unless the demo kit
depends on it (verify before removing; if it stays, answers carry a 'from your records'
marker)"* — is untouched. It is still in the read tier at
`adapters/tools.manifest.json:198` and `services/agent-orchestrator/src/tools.ts:423`, no
marker was added, and the PR body does not name it or link a deferral. Not mine to fix:
Dev/Product call, but it should be an explicit decision rather than an omission.

### MEDIUM — the allergen field is now blank for every allergy, and the test that caught it was edited to match

`tests/test_sdc_routes.py:144-159` changed
`assert allergen["answer"][0]["valueString"] == "Penicillin"` to
`assert allergen is not None and "answer" not in allergen`.

The PR body discloses this as a ruled trade ("SNOMED allergens, today"), and the
non-negotiable is intact — the repeat is still emitted and `no-known-allergies` is still
never inferred, which I re-ran and confirmed. But the scope is larger than "today"
suggests. `r6/terminology.py` holds **121 labels and zero SNOMED entries**, and
`TERMINOLOGY_LOOKUP_ENABLED` is unset in every deploy config, so the resolver is off in
production. Its own docstring records the live measurement: *"1 of 15 distinct ICD-10-CM
codes and 0 of 11 SNOMED codes carried a label"* on a real import.

Live, on a seeded patient:

```
allergies.item.allergen  = <no answer>     <- SNOMED-coded allergy
allergies.item.reaction  = <no answer>
conditions.item.name     = <no answer>     <- SNOMED-coded condition
medications.item.dose    = <no answer>     <- dosageInstruction[].text stripped
medications.item.name    = "Metformin 500 mg"   <- RxNorm code recognised
```

A patient on a phone reviewing their intake form sees an allergy row with no allergen. On
a real MEDENT-shaped import that is most conditions too. That is the D14 clinical-honesty
shape, one surface over. I am not overturning a council trade — routing it to
Product/Clinical with the numbers, because "SNOMED allergens" reads narrower than the
measurement is.

### LOW

- `r6/sdc/expressions.py:177` logs the caller-supplied expression at WARNING on every
  evaluation failure. No PHI (only `type(exc).__name__` accompanies it), but it is
  caller-controlled text in the log, newlines included. Pre-existing, not introduced here.
- `_gather_content` still appends the unredacted subject to `content`. After `ecd29fa`
  nothing reads it — Observation matching and the three list-group types all exclude
  Patient — so it is dead weight that was one code path away from being a leak. Worth
  removing while the reasoning is fresh.
- `_coerce` returns dicts unchanged, so `%patient.select($this)` yields a `valueString`
  holding an object — invalid FHIR, bounded content. Pre-existing.
- Missing/blank `X-Tenant-Id` now returns **400** `security` (kernel handler) where the
  old path returned 401. Both refuse; measured on the PR, and `tests/test_access_kernel.py`
  records `r6/sdc/routes.py`'s kernel adoption deliberately in `_ADOPTION_ALLOWED` with a
  reason, which is that pin working as designed rather than being widened quietly.

---

### What I ran

```
  3180 passed, 13 skipped, 1 xfailed        (baseline on d0b3971: 3178 passed)
uv run ruff check .
  All checks passed!
uv run python -m pytest tests/test_ratchets.py tests/test_guardrail_conformance.py -q -p no:cacheprovider
  50 passed                                  (Grade A asserted and green)
```

Plus, against a running app on `:5099` with a seeded tenant: the 32-expression escape
sweep, the 31-marker redaction sweep across all four auto-loaded types, the SSN oracle
before and after, the `%resources` oracle, the DoS timings, the audit-row count and
content, the tenant-header matrix on both `$populate` and `$extract`, and the real intake
Questionnaire end to end.

### What I did NOT cover

- The inline `content` Bundle pass-through. Left unredacted by design (caller's own data);
  I never exercised it, so the argument stands unmeasured.
- `$extract` beyond the tenant-header matrix — no roundtrip, and specifically **not** what
  `$extract` writes back from a form carrying a blank allergen row. Whether that can
  commit an AllergyIntolerance with no `code` is untested and worth someone's hour.
- The delivery route live. The 7d→24h cut is verified only by the PR's own unit tests —
  which are good ones (they pin the enforced boundary, not just the constant).
- The Postgres CI lane; SQLite only here.
- The MCP `questionnaire_populate` caller against the changed response shape.
- Any browser/mobile rendering of a populated form.

### Two process notes

1. **`ecd29fa` and `da16068` are production code written by QA and reviewed by nobody.**
   The review task directed the fix; it did not authorize skipping review. Please have the
   CTO read `r6/sdc/expressions.py` before this merges — particularly whether answering
   from a constant probe is the shape they want, versus emitting the issue
   unconditionally.
2. `dependency-audit` is red on every open PR right now. It is npm advisories fixed in
   **unmerged PR #544**, not anything from this branch.

**Do not arm auto-merge until the CTO has reviewed `ecd29fa`.** (`autoMergeRequest` was
null before both of my pushes; I checked each time.)

aks129:
## CTO ruling — PR #562 (D10 `$populate` bound)

**This comment is not a GitHub approval and must not be read as one.** On this repo an
approval arms `gh pr merge --squash --auto` (`.github/workflows/claude-pr-review.yml:148`),
so clicking Approve *is* the merge. `autoMergeRequest` was null when I started and null
when I posted. The only thing holding this PR now is `dependency-audit`, which is red for
an unrelated reason (#544) — that is not a safety margin.

**Verdict: APPROVE-WITH-CHANGES. Do not split.** The leak is closed; the shape QA reached
for is not the one to keep, and the allergen regression has to ship with its own fix.

---

### 1. The oracle fix — `ecd29fa` / `da16068`

**The fix holds. Keep it until the replacement lands, then delete the whole mechanism.**

Verified, not taken on trust. In-process against the pure function (the route only calls
it, so the interesting surface is the function): both evaluations run against module
constants, the result is a pure function of the expression text, repeated calls with
`cache_clear()` between them are identical, and neither `_PROBE_PATIENT` nor
`_PROBE_RESOURCES` is mutated by any expression I tried, including `combine()`, `|`,
`iif()`, `children()`, `descendants()`, `%context`, `%resource`, `%rootResource`, `%ucum`.
Hostile shapes cost 0.001–0.008 s. Over HTTP against a seeded synthetic tenant, QA's own
eleven-request SSN walk recovers `''`. **The probe cannot be steered.**

**But it is the wrong control, and the live numbers say so.** Same running app, one
expression per request, reading whether the OperationOutcome named the linkId:

```
issue=True   %patient.identifier.value                      <- D10 negative
issue=True   %patient.photo.url                             <- D10 negative
issue=True   %resources.where(resourceType='Observation').code.text
issue=True   %patient.maritalStatus.text
issue=True   %patient.telecom.where(system='sms').value
issue=True   %patient.identifier.first().value

issue=False  %patient.identifier.value.where($this.startsWith('1'))
issue=False  %patient.identifier.value.where($this.startsWith('9'))
issue=False  %patient.identifier.value.combine(%patient.gender)
issue=False  %patient.contact.gender
issue=False  %patient.deceasedDateTime
issue=False  %patient.name.period.start
```

Six of ten withheld paths come back **silently empty** — including the exact expression QA
used to walk the SSN out. The classifier fires iff the caller's filter happens to match
the placeholder literal: `where($this.startsWith('probe'))` returns True,
`where($this.startsWith('1'))` returns False. `da16068`'s docstring admits
non-exhaustiveness and names three leaf paths; the real gap is whole expression classes,
and it is the ordinary way an author writes a filter.

That is `docs/2026-08-02-retro.md` pattern #1, verbatim: a control that looks like "we tell
you when we refused" and does "we tell you when your expression matches our placeholder."
And the failure lands exactly where D10 said not to — a caller reads silence and concludes
the patient has no such value.

The three rewritten tests at `tests/test_sdc_expressions.py:95-129` do not save it. Nine
hand-picked plain paths that happen to hit the placeholder inventory is the implementation,
not the property. Add `where($this.startsWith('1'))` to that `withheld` list and it goes red
today. (The oracle test at :123 and the signature pin at :163 are good and pin real
properties — those two survive in spirit.)

**Ruled: emit the issue unconditionally, drop both evaluations, delete the probe.**

The zero-bits argument is what decides it. `_populate_item` already emits every leaf and
attaches `answer` only when a value resolved (`r6/sdc/populate.py:232-235`), so *which
linkIds have no answer is already in the response*. An issue per unanswered
expression-bearing leaf is a redundant encoding of information the caller is holding. Not
"we checked it doesn't leak" — there is nothing to leak, by construction, and no one has to
re-verify that when they next touch the file. It is also *complete*, which the probe is not:
every withheld path names itself, not just the ones the inventory covers.

Concretely:

- Delete `resolves_outside_projection`, `_PROBE_PATIENT`, `_PROBE_RESOURCES` and the
  `lru_cache` (~130 lines). `patient_projection` and `build_context` are the bound; they stay
  exactly as they are — QA's 32-expression escape sweep says they hold, and I agree.
- In `_resolve_answer`, append an issue whenever an expression resolved nothing.
- Retext. `WITHHELD_BY_PROJECTION` becomes a statement that is true in both cases and claims
  nothing about the patient — roughly *"not populated: no value resolved. $populate reads
  only the %patient projection (name, birthDate, gender, telecom phone/email, address
  line/city/state/postalCode) and coded content this server recognises; anything else is not
  read."* The allowlist is public; hand it over and let the caller compare it against their
  own expression, instead of asking the server a question about a patient. This matters
  extra because `questionnaire_populate` is in the model-facing tier — the current text
  would have a model tell a patient their record was withheld when they simply have no email.
- **Severity must drop to `information`.** `r6/sdc/routes.py:328` emits
  `severity: warning, code: incomplete`; under unconditional emission a conformant client
  would read an ordinary form with one empty optional field as a failed operation.
- The noise is small and measured: the intake form carries nine `initialExpression` items
  (`r6/sdc/intake.py`), and on a seeded patient with full demographics all nine answer —
  **0 issues**. One absent demographic, one issue.

Replacement test, and this is the one that has to exist: for a matrix of patients and
adversarial expressions, assert **`answer present ⟺ no issue for that linkId`, per item**.
That is the zero-bits property stated directly, it is implementation-independent, and any
re-introduced record-driven classifier goes red on the first item it silences. Keep D10's
three named negatives as an explicit row, and add one for an allowlisted path the fixture
patient genuinely lacks. The signature pin at :163 dies with the function; this replaces it.

---

### 2. The allergen labelling

QA's MEDIUM understates it, and so does the PR body. Live, seeded synthetic patient, two
SNOMED allergies + one SNOMED condition + one RxNorm medication:

```
demographics.*            all nine populate correctly
medications.item.name     Metformin 500 mg     <- RXNORM 860975 is in the table
medications.item.dose     <NO ANSWER>          <- dosageInstruction[].text stripped
allergies.item.allergen   <NO ANSWER>          <- SNOMED 91936005
allergies.item.reaction   <NO ANSWER>          <- SNOMED 247472004
allergies.item.allergen   <NO ANSWER>          <- SNOMED 300916003
allergies.item.reaction   <NO ANSWER>
conditions.item.name      <NO ANSWER>          <- SNOMED 44054006
ISSUES (0)
```

Seven of eight clinical fields blank, and **zero issues** — the response carries no signal at
all that anything was dropped. "SNOMED allergens, today" should read: every allergen, every
allergy reaction, every SNOMED-coded condition, and every medication dose. `r6/terminology.py`
holds 121 labels — 49 LOINC, 23 ICD-10, 20 RxNorm, 15 ICD-9, 14 CVX, **0 SNOMED** — and
`TERMINOLOGY_LOOKUP_ENABLED` appears in no deploy config. Please correct the PR body; a claim
in a PR body is product surface (retro, pattern 2).

**Ruled, of the five options:**

- **Omit the row — BLOCKED.** A patient with three allergies seeing one is a false negative on
  an allergy list, and an allergy section that renders empty is one click from someone ticking
  "no known allergies". Row count must equal record count. It does today (two stored → two
  repeats); pin it.
- **Hold the redaction of AllergyIntolerance — BLOCKED.** That restores upstream `code.text` on
  the resource type most likely to carry free text. It is the non-negotiable D10 just closed.
- **Turn `TERMINOLOGY_LOOKUP_ENABLED` on — DEFER.** It sends `(system, code)` to an external
  service per read. That is a subprocessor and privacy-posture decision (D3 P6 /
  `docs/2026-08-04-shakeout-rulings.md:49`), not a code review one.
- **Marker in the answer — NOT AS A `valueString`.** I checked `$extract` live, which QA flagged
  as untested, and it is worse than a round-trip risk. `allergies.item.allergen` is
  `type: "string"` with `definition: AllergyIntolerance#AllergyIntolerance.code.text`, and
  `_extract_by_definition` walks *every* item with a definition into the single target type
  (Patient). A committed intake form produced:

  ```
  POST /r6/fhir/QuestionnaireResponse/$extract   (step-up token, not dryRun) -> 200
  stored row 52674a15-…:
    {"resourceType":"Patient","name":[{"family":"Synthetic"}],
     "code":{"text":"Allergy on file - name not available"}}
  ```

  A **new duplicate Patient**, carrying the allergen field's text in `Patient.code.text`. Any
  marker string we put in that field becomes fabricated clinical text in the record. This is
  pre-existing and documented as a v1 limit (`r6/sdc/intake.py:46-56`) — it means a real
  allergen name does the same thing today — but it rules out the marker-as-answer shape.
- **Carry the coded value — DEFER** to the same fix. The item is `type: "string"`; a
  `valueCoding` needs a Questionnaire change with extract consequences.

**So: the unavailable state is reported, not answered.** Extend ruling 1's rule to list-group
leaves — a leaf that resolved no value gets an issue naming its linkId, same text, same
`information` severity, one mechanism for both paths. The response then says
`allergies.item.allergen: not populated` twice and `conditions.item.name: not populated` once,
the renderer has what it needs to show "name not available" without any string entering the
answer set, and `$extract` has nothing new to round-trip. On a real MEDENT-shaped import that
is ~25 issues on one form — that count is the point, and it is the honest measure of coverage.

**The actual fix for the common case is a dict addition:** SNOMED entries for the frequent
allergens (penicillin, sulfa, latex, peanut, shellfish, NSAIDs) and the frequent chronic
conditions. No new surface, no network, nothing patient-specific. It is a clinical-safety
change though — a wrong allergen label is dangerous — so it needs the physician advisor's
initials on the mapping (D14's rule-register pattern). **Named follow-up, not a merge gate**,
so the PR does not wait on a signature.

**Before this reaches a real patient (D3 P5):** row count equals record count, pinned; no
blank clinical row without an issue naming it; `no-known-allergies` still never inferred
(existing test, re-ran, green); clinician initials on any label added; and **`$extract` fixed**
— it currently forks the Patient record and writes answers into elements the target resource
does not have. That last one is pre-existing and does not block #562, but it blocks the
real-patient path and should be filed today.

---

### 3. Merge

**APPROVE-WITH-CHANGES, do not split.** The projection, the redaction, the `is_deleted`
filters and the 7d→24h cut are all correct and I want them. The blank-allergen regression is
*caused by* the redaction in this PR, so it ships with it — retro rule 2, ship a feature and
its hardening as one unit or ship neither.

Required before merge (a dev implements; I wrote no production code):

1. Ruling 1 — unconditional issue, probe deleted, severity `information`, neutral text, the
   `answer ⟺ no issue` property test.
2. Ruling 2 — list-group leaves get the same issue; row count pinned.
3. Delete the unredacted subject append at `r6/sdc/routes.py:199`. Nothing reads it after
   `ecd29fa`, it was the second door to the oracle, and it is one `%resources` away from being
   a leak again.
4. D10 bullet 5 (`questionnaire_populate` in the model-facing tier): the "from your records"
   marker, or a written deferral in the PR body naming the bullet. The ruling's premise for
   pulling it is genuinely removed by this PR, so leaving it in is defensible — silently
   omitting it is not.
5. PR-body correction on the scope of the labelling trade.

Not blocking; please file: `$extract` duplicate-Patient and wrong-element write; the SNOMED
label additions (clinician initials); `r6/sdc/expressions.py:289` logging caller-supplied
expression text at WARNING (log injection, newlines included, pre-existing); `_coerce`
returning dicts.

On QA's process note — correct call, and the fix was right to land as written. It is a good
patch of a live leak. It is also 130 lines of mechanism computing a fact that is already
public, which is the kind of cleverness this codebase should spend down rather than keep.

**What I ran.** Read `CLAUDE.md`, the retro, the access-kernel spec, D10, the full diff and
QA's review. In-process steering and purity checks on `resolves_outside_projection`. A
running app on `:5199` with a synthetic seeded tenant (SNOMED allergies/condition, RxNorm
medication, an SSN-shaped identifier — synthetic throughout): the intake `$populate` end to
end, the fifteen-expression issue matrix above, QA's eleven-request SSN walk (recovered `''`),
and `$extract` in both dryRun and committed modes with the stored row read back. I changed no
repository code, so `ruff` and the suite were not re-run — QA's `3180 passed` / ruff-clean on
`da16068` stands and my worktree is clean at that commit.

aks129:
### CTO addendum — one correction and four clarifications

Still not an approval; `autoMergeRequest` still null.

**Correction.** My ruling said the `no-known-allergies` invariant was "existing test, re-ran,
green" while also saying I ran no suite. Both could not be true. What I had actually verified
was NKA staying unanswered on a patient *with* two allergies on file — not the zero-allergies
case the invariant is about. I have now run the test:

```
uv run python -m pytest tests/test_populate_lists.py -q -p no:cacheprovider
8 passed in 0.07s
```

Green, including `test_zero_allergies_never_infers_no_known_allergies`. The ruling stands;
the sentence supporting it did not, and I would rather correct it here than leave it.

**Line number.** The unredacted subject append is `r6/sdc/routes.py:205`, not `:199`.

**The exclusion predicate — the part a dev will otherwise get wrong.** "Every unanswered leaf
gets an issue" must exclude leaves with **no population mechanism at all**. `allergies.no-known-allergies`
and `medications.no-current-medications` carry no `code` and no `initialExpression` precisely
so nothing ever touches them (`r6/sdc/intake.py:217-225`, `r6/sdc/populate.py` docstring). An
issue reading "not populated: no value resolved" against `no-known-allergies` tells a
model-facing reader the system tried to determine NKA and failed, which is a hair from
inferring it. Rule: the issue attaches only where population was *attempted* — an
`initialExpression`, an `item.code` Observation match, or a list-group leaf with a `definition`.
Pin it as a named row: **those two linkIds never appear in the issue list, on any patient,
with any content.**

`_observation_answer` items belong in the rule too — a code-matched item that found no
Observation is an attempted population that resolved nothing. The intake form has no
`item.code` items, so this is invisible today; another Questionnaire will have them.

**The two issue counts are not in conflict.** "0 issues" in ruling 1 is expression items only
(all nine demographics answered on the seeded patient). Under the full ruling, that same
patient yields **six**: two allergens, two reactions, one condition, one medication dose.

**One thing I did not check.** I wrote that the renderer "has what it needs". The issue channel
is machine-readable and complete, but whether the intake form UI or the PDF in
`r6/sdc/documents.py` actually surfaces it on the affected row is unverified. That is a
real-patient gate alongside the `$extract` fix, not a merge gate: a blank allergen row that is
*reported* in the response and still *renders* blank on a patient's phone has not been fixed
where it matters.

aks129:
Correction to this PR's ratchet table, from a finding on the parallel session's #618.

This body reports `Raw tenant reads | 5 | 5` — accurate as a reading of the pin, and the pin was wrong. An audit of that ratchet found its matcher was a single-quoted, no-default-argument substring needle that **missed 22 of the 27 actual raw tenant-header reads in the codebase**, including one call split across two lines that no text needle could catch at any quoting, and one in `r6/agent_runs/routes.py` that is the access-kernel spec's own open slice — invisible to the pin the entire time.

So the honest statement of what this PR did to that count is: it left the two queries in the forms route in place, and the number beside them was measuring about a fifth of what it named. "Unchanged rather than reduced" is still true. "Five" was never the population.

It also resizes work the council scheduled. Ruling D11 put A5 in the parallel lane as *raw tenant reads 5 → 0*. The real slice is 27 → 0, which is a different piece of work and should be planned as one.

Nothing to change in this PR — the code did what it says. The number it cites came from a broken instrument, and that is worth recording here rather than leaving a future reader to reconcile it against #618.

Issue #416, opened 2026-08-05
A confirmed clinical action runs, and the patient is told to approve it again (edge 502/504 on confirm)


aks129:
Found by adversarial QA driving a fault proxy between CareAgents and a live HealthClaw. **Patient-safety defect, and it is the exact harm #220 and #403 were written to prevent, reached through a door they did not close.**

## What happens

`careagents/healthclaw.py:332`:

```python
r = self._send(..., what="confirm", error=HealthClawUnconfirmed)
if not r.ok:
    raise HealthClawError(f"confirm failed ({r.status_code})", r.status_code)
```

`_send(error=HealthClawUnconfirmed)` covers the case where **no response arrives** — a dropped socket, a read timeout. That works, and QA verified it.

It does not cover the case where a **gateway answers on the upstream's behalf**. An edge 502/503/504 is a response, so `_send` returns it, and `if not r.ok` then classifies it as a refusal. A gateway 504 is not evidence the engine did nothing; it is the *more common* production shape of "we do not know". The #374 work names exactly this: "a 502 after `claim_next` commits".

Driven against a live engine with the confirm executed and the edge answering 504:

```
engine action status: awaiting_confirmation  ->  failed
patient gets HTTP 502
body: {"confirmed": false,
       "message": "Your review was saved, but we couldn't submit the approval.
                   Nothing has been sent — please try approving again."}
```

The action **was** consumed and executed. The patient is told nothing was sent and instructed to approve again. On the human-approval path, for a clinical action. That is a double-execution instruction.

The dropped-socket path is handled correctly by contrast (`confirmed: null`, "Don't approve again yet"), which is what makes the gap easy to miss.

## Narrower sibling, same false message

A confirm answered `200` with a body that is not JSON raises plain `HealthClawError` from `_json_object`, producing the same "Nothing has been sent". A 200 we cannot parse is also not evidence of a refusal.

## Why the suite missed it, which is the more useful finding

Two test shapes each looked sufficient and neither was:

1. `tests/test_careagents.py:304-309` uses `status_code = 409` as its "observed rejection". A 409 genuinely is an engine answer, so the assertion is true — and it silently generalizes to *all* non-2xx, including the ones that are not answers at all.
2. The route-level test at `tests/test_careagents.py:240-245` uses a fake client that **raises `HealthClawUnconfirmed` itself**. It proves the route handles the exception. It cannot prove the client ever produces it.

[...]

PR #566, opened 2026-09-03
candidate: the approve page answers a 409 instead of going quiet (follow-up to #528)

aks129:
## What

Every outcome the approve journey can produce now says something true, and no
branch directs the patient into a retry that cannot succeed.

**The page** (`templates/action_review.html`) — four outcomes, one renderer:

- **`.catch()` on the fetch itself.** A dropped connection never produces a
  response, so the chain went silent: nothing rendered, Approve left enabled
  with no explanation.
- **`.catch()` on the JSON parse.** Any non-JSON body — the Werkzeug HTML 500
  that started #528, a gateway's 502 page — rejected the promise, same silence.
- **A 409 branch, before the generic failure arm.** Neutral, not red.
- **A branch for the relay's confirm-failure**, which disables Approve (its
  retry answers 409) and reads the status its own message points at.
- **`checkStatus()` answers `awaiting_confirmation`** — the status the action
  actually holds in both of those cases. It used to fall through to "we still
  cannot tell", which is weaker than what is on file.
- The success path is unchanged.

**The relay** (`careagents/app.py`) — the confirm-failure copy no longer tells
the patient to approve again, because since #528 that retry 409s.

### The exact patient-facing sentences

> **409 (already approved)** — You have already approved this request. We are
> checking where it stands.

> **Unreadable body** — We could not read the answer to your approval, so we
> cannot tell you whether it went through. We have not sent it again. Reload
> this page to see where this request stands.

> **Service unreachable (fetch rejected)** — We could not reach the service, so
> we cannot tell you whether your approval was sent. Reload this page to see
> where this request stands before approving again.

> **Relay, confirmation failed after a saved review** — Your review was saved
> and your approval is recorded, but we could not complete it just now.
> Approving again here will not resend it. Check where this request stands.

> **Status terminus, approval recorded but not carried out** — Your approval is
> recorded and this request has not been carried out yet. Approving again will
> not resend it.

## Why

**The silent paths.** Neither the fetch rejection nor an unparseable body was
handled. The chain rejected, nothing rendered, and Approve stayed enabled —
the invitation to double-tap that produced #528 in the first place. `60d498a`
removed one *cause* of an unreadable body; it did not close either branch.

**The mislabelled path.** One correction to the framing this task was filed
under: after `60d498a` a 409 is not silent. `_error` returns JSON, so the
server's sentence rendered — as a red `alert-danger`, with Approve still
enabled. That is the #416 shape one status code over: a definite failure shown
for a request that *was* approved. Told it failed, a reasonable person approves
again. Red-plus-enabled is what the 409 branch fixes; the two `.catch`es fix
the genuinely silent cases.

**The loop.** `careagents/app.py` told the patient "Nothing has been sent —
please try approving again" when `confirm_action` failed after a 200 review.
The review route already minted the confirmation, so that retry answers 409 and
the page says they have already approved. The instruction pointed into a loop
the page itself creates, and "nothing has been sent" reads as "start over" when
an approval *is* on file.

No branch says "nothing was sent". A confirmation exists, so that would be a
claim never observed — enforced over the new page copy by the existing
`test_no_branch_can_call_an_unknown_outcome_a_rejection`, and over the new relay
copy by the same list.

### Two rulings worth stating

**Approve stays ENABLED on an unreachable service, alone among these branches.**
A dropped connection most often means the request never left the device, and a
disabled button with no way forward strands a patient mid-approval. Naming the
uncertainty and pointing at the status is the honest option that keeps them
moving. The option ruled out is the silent failure. Pinned, because the next
person tidying these into one shape would disable it for consistency.

**`confirmed` stays `False` on the relay's confirm failure — not changed to
`null`, against the brief's suggestion.** Only a failed approval-token mint (the
confirm never went out) or a 4xx the engine itself answered reaches that branch.
Every ambiguous case — transport loss on the confirm POST, a gateway
5xx/408/429, an undecodable 200 — is routed to `HealthClawUnconfirmed` and
already answers `null` (`careagents/healthclaw.py:352-357`, `:387-397`). The
action did not execute, and that is *known*, so `null` would assert an
uncertainty that does not exist: #220's third answer collapsed from the other
side. `test_review_submit_reports_a_failed_confirmation` already pinned this and
goes red on `null` — verified. The brief's stated rule ("match what is actually
known") is what selects `False`; only its premise that the fate is unknown does
not survive the read. The instruction was the defect, and the instruction is
what changed.

**One template covers both surfaces**, verified by reading rather than assumed:

- `careagents/app.py:1076-1077` — `review()` proxies this page and rewrites only
  the submit URL.
- `careagents/app.py:1102-1153` — `review_submit()` reaches
  `return jsonify(body), status` for a 409: `_answered_about_data(409)` is true
  (`careagents/healthclaw.py:120-140`), so the unanswered branch is skipped and
  the `status == 200` branch is not taken. Body and status relay verbatim.

## Base note

**Stacks on PR #550 — merge that first.** Based on
`fix/528-payload-immutable-after-confirmation`, not `main`. The 409 this page
renders is minted by `60d498a` on that branch; on `main` the second submit is
still a 500 and the new server assertions have nothing to assert against.

## Ran

```
    3181 passed, 13 skipped, 1 xfailed, 2 warnings in 105.70s (0:01:45)

uv run ruff check .
    All checks passed!

    50 passed in 2.52s          (conformance Grade A)
```

`r6/routes.py` untouched at 3924 lines, under the `<= 3927` ratchet
(`tests/test_ratchets.py:341`).

Every new guard is mutation-tested, each mutation restored after:

| mutation | goes red |
|---|---|
| `.catch` removed from the parse | `test_the_json_parse_has_a_catch` |
| `.catch` removed from the fetch | `test_the_fetch_itself_has_a_catch` |
| unreachable branch disables Approve | `..._leaves_the_button_enabled` |
| `checkStatus()` dropped from the 409 branch | `..._disables_the_button_and_checks_the_status` |
| 409 rendered `alert-danger` | `..._does_not_render_as_a_failure` |
| `btn.disabled` dropped, unreadable branch | `..._leaves_the_button_disabled` |
| 409 branch deleted | 4 tests |
| 409 branch moved below the fallback | `..._runs_before_the_generic_failure_message` |
| 409 body made non-JSON | the server test's `mimetype` assertion |
| old relay copy restored | the relay guard, on `try approving again` |
| relay `confirmed` switched to `null` | `test_review_submit_reports_a_failed_confirmation` |
| `awaiting_confirmation` arm deleted | 2 tests |
| "will not resend" clause dropped | `..._never_invites_a_second_approval` |
| `checkStatus()` cut from the 502 branch | `..._gets_the_destination_it_points_at` |
| 502 branch keyed on `confirmed` alone | 2 tests |
| unreachable branch made to poll the status | `..._leaves_the_button_enabled` |

Where the tests landed:

- The brief named `tests/test_mcp_app_care_gaps.py` as the style reference. That
  file does not exist. The page guards went into
  `tests/test_review_says_what_it_knows.py`, which uses the same
  assert-against-the-template-source pattern and already guards this exact
  submit handler.
- Both server-side tests are **extended, not duplicated**:
  `tests/actions/test_review_flow.py` gains the 409 body shape the page reads,
  asserted *before* the existing `['error']` read — that line already failed on
  a non-JSON body, but as a bare `TypeError` naming nothing.
  `tests/test_careagents.py` gains the forbidden-claims pin on the relay copy.
- `tests/test_healthclaw_transport.py` had a docstring saying the removed
  instruction "must keep being said". That test pins the exception *type*, and
  still does; the docstring is corrected, the assertions are untouched.

## What was NOT run

- **No e2e.** `e2e/tests/` has no approve-flow spec (`design-invariants.spec.ts`
  only mentions the word), and per the brief this PR does not create an e2e
  suite.
- **No live CareAgents stack.** The relay pass-through is verified by reading the
  source, as the brief directed. Repo trap applies: the CareAgents tests fake the
  HealthClaw client, so nothing there proves the 409 is *accepted* end to end.
- **No browser.** These guards assert the template source. CI has no browser, so
  the rendered branches are unexercised — the same limit the neighbouring guards
  in that file already state.

## The terminus, and a precondition worth reading

Both the 409 branch and the relay's confirm-failure send the patient to check
the status. `checkStatus()` handled `completed`/`sent`/`pending`/`approved` and
nothing else, so `awaiting_confirmation` — exactly where the action sits in both
cases — fell through to *"We still cannot tell whether your approval went
through"*. The instruction returned less than the branch that issued it already
knew. It now says the approval is recorded, the request has not been carried out
yet, and approving again will not resend it.

**The precondition on that sentence is load-bearing, and is written into the
code rather than assumed.** `awaiting_confirmation` is set at COMMIT, before any
human approval (`r6/actions/models.py:35`, `r6/actions/routes.py:417`) — so the
status *alone* does not establish that an approval exists. "Your approval is
recorded" is true only because every caller of `checkStatus` has already
established a confirmation: a 200 review mints one, and the 409 fires only when
one is present.

The unreachable-service branch is the one place that has **not** established it,
and it deliberately does not call `checkStatus()` — pinned by a test, for two
reasons: its terminal copy ("Please do not approve a second time") would
countermand the ruling that leaves Approve enabled there, and its
`awaiting_confirmation` arm would claim an approval that branch never observed.

The confirm-failure branch is keyed on the 502 **and** `confirmed === false`,
not on the value alone: the two 503s carry that same value for a review that
never reached the engine, where there is no approval on file and the poll would
report on a request that was never made.

## CI note

`dependency-audit` is red on every open PR — npm advisories fixed in unmerged
PR #544, unrelated to this change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,8 +3,12 @@ name: CI - FHIR MCP Guardrails
 on:
   push:
     branches: [main, master, 'claude/**']
+  # No `branches:` filter here on purpose. That filter matches the PR's BASE
+  # branch, so a change stacked on another feature branch matched nothing and
+  # this entire workflow was skipped — every test lane with it (#585). Every
+  # job below is read-only reporting, so it is safe on any base; if one ever
+  # becomes too expensive for a stack, gate that job, not the workflow.
   pull_request:
-    branches: [main, master]
 
 # Least privilege: every job here only reads the repo (tests, lint, gitleaks,
 # audits). No job writes contents, comments, or uploads results.
--- a/.github/workflows/claude-pr-review.yml
+++ b/.github/workflows/claude-pr-review.yml
@@ -26,6 +26,15 @@ permissions:
   pull-requests: write
   issues: write
 
+concurrency:
+  # One run at a time per PR. Two pushes minutes apart produce two runs, and
+  # the notification dedupe below reads back a comment the other run may not
+  # have posted yet — so both would see "nothing said yet" and both would say
+  # it. Queue instead of racing. `false`, not `true`: cancelling a run
+  # mid-flight would abandon the arming step half-done.
+  group: claude-pr-review-${{ github.event.pull_request.number }}
+  cancel-in-progress: false
+
 env:
   AUTO_MERGE_EXTERNAL: "false"
 
@@ -126,30 +135,93 @@ jobs:
           AUTHOR: ${{ github.event.pull_request.user.login }}
         run: |
           set -e
+          # This block decides ONE thing — the state the PR is in — and then
+          # reports it twice, differently. GitHub mails every comment; it
+          # mails nothing about a job summary. So the state always goes to the
+          # summary, and only a state the maintainer must know about or act on
+          # is allowed to become a comment.
+          #
+          # Keep this free of Actions expression syntax — every value it
+          # needs arrives through `env:` above — because
+          # tests/test_pr_review_notifications.py executes this block verbatim
+          # against a stub `gh`, and that is the only place the logic runs
+          # anywhere but a live PR.
+          STATE=""; HEADLINE=""
           # Tier 1: dependabot minor/patch -> full auto (merges when checks pass)
           if [ "$AUTHOR" = "dependabot[bot]" ]; then
             TITLE=$(gh pr view "$PR" --repo "$REPO" --json title --jq .title)
             # minor/patch = the MAJOR version component is unchanged
             FROM_MAJ=$(printf '%s' "$TITLE" | sed -nE 's/.* from ([0-9]+)\..* to [0-9]+\..*/\1/p')
             TO_MAJ=$(printf '%s' "$TITLE" | sed -nE 's/.* from [0-9]+\..* to ([0-9]+)\..*/\1/p')
-            if [ -n "$FROM_MAJ" ] && [ "$FROM_MAJ" = "$TO_MAJ" ]; then
+            if [ -z "$FROM_MAJ" ]; then
+              # Half of this repo's dependabot PRs are grouped or multi-package
+              # bumps with no single "from X to Y" in the title. They are held
+              # for the same reason a major is, but for a different reason than
+              # a major, so they say so in their own words.
+              STATE="held-unparsed-bump"
+              HEADLINE="Not armed. This title carries no single from-X-to-Y version, so the minor/patch rule cannot be applied to it and nothing will merge on its own. A human has to review and merge it."
+            elif [ "$FROM_MAJ" = "$TO_MAJ" ]; then
               gh pr merge "$PR" --repo "$REPO" --squash --auto
-              echo "dependabot minor/patch: auto-merge armed"
+              STATE="armed-dependabot"
+              HEADLINE="Auto-merge is ARMED. Minor/patch bump, so it squash-merges by itself once the required checks are green, and no approval is asked for. To stop it: gh pr merge $PR -R $REPO --disable-auto"
             else
-              echo "dependabot MAJOR (or unparsable) bump — leaving for human review"
+              STATE="held-dependabot-major"
+              HEADLINE="Not armed. This is a MAJOR version bump ($FROM_MAJ to $TO_MAJ), which never merges on its own here. A human has to read the changelog and merge it."
             fi
+          else
+            # Tier 2: external/human code — arm only if a maintainer approved
+            # (or AUTO_MERGE_EXTERNAL flips the policy)
+            APPROVED=$(gh pr view "$PR" --repo "$REPO" --json reviews \
+              --jq '[.reviews[] | select(.state=="APPROVED") | select(.authorAssociation=="OWNER" or .authorAssociation=="MEMBER" or .authorAssociation=="COLLABORATOR")] | length')
+            if [ "$APPROVED" -gt 0 ] || [ "${AUTO_MERGE_EXTERNAL}" = "true" ]; then
+              gh pr merge "$PR" --repo "$REPO" --squash --auto
+              STATE="armed-approved"
+              HEADLINE="Auto-merge is ARMED. This PR squash-merges by itself the moment the last required check goes green — including a check that is red now and gets fixed later, and including mid-review. To stop it: gh pr merge $PR -R $REPO --disable-auto"
+            else
+              STATE="waiting-approval"
+              # The wording this replaced said the PR would merge "the moment a
+              # maintainer approves". It cannot: this workflow has no trigger
+              # on reviews, so an approval on its own runs nothing. The trigger
+              # list quoted here is pinned by
+              # test_the_wording_matches_the_triggers_it_describes.
+              HEADLINE="Not armed. Waiting for a maintainer approval. Approving alone arms nothing, because this workflow runs on opened, synchronize, reopened and ready_for_review and has no trigger on reviews — it arms on the next push, reopen or ready-for-review that happens after an approval exists."
+            fi
+          fi
+
+          # Always. A job summary notifies nobody, which is precisely why the
+          # states that need no action end here and nowhere else.
+          printf '### auto-merge: %s\n\n%s\n' "$STATE" "$HEADLINE" \
+            >> "$GITHUB_STEP_SUMMARY"
+          echo "$STATE: $HEADLINE"
+
+          # Only a state the maintainer must know about or act on is mailed.
+          # "Everything passed, waiting for you" is not one — they can see the
+          # PR is waiting. This job runs on every push, and that sentence was
+          # mailed 65 times across the 25 most recent PRs, six times on one of
+          # them, four minutes apart on another.
+          case "$STATE" in
+            armed-*|held-*) ;;
+            *) exit 0 ;;
+          esac
+
+          # ...and only when it CHANGES, or five pushes send five identical
+          # mails. The previous state is read back out of the marker in our own
+          # last comment: no external storage, and it survives a re-run. The
+          # alternative — editing one comment in place — is silent, but it is
+          # silent at the state change too, and the state change is the only
+          # moment worth a mail. So: post on change, say nothing otherwise.
+          #
+          # `grep` matching nothing does not trip `set -e` here: a pipeline's
+          # status is its LAST command's, and `sed` succeeds on empty input.
+          PREV=$(gh pr view "$PR" --repo "$REPO" --json comments \
+                   --jq '.comments[].body' \
+                 | grep -o 'automerge-state: [a-z-]*' | tail -n1 | sed 's/.*: //')
+          if [ "$PREV" = "$STATE" ]; then
+            echo "state unchanged since the last comment — not mailing it again"
             exit 0
           fi
-          # Tier 2: external/human code — arm only if a maintainer approved
-          # (or AUTO_MERGE_EXTERNAL flips the policy)
-          APPROVED=$(gh pr view "$PR" --repo "$REPO" --json reviews \
-            --jq '[.reviews[] | select(.state=="APPROVED") | select(.authorAssociation=="OWNER" or .authorAssociation=="MEMBER" or .authorAssociation=="COLLABORATOR")] | length')
-          if [ "$APPROVED" -gt 0 ] || [ "${AUTO_MERGE_EXTERNAL}" = "true" ]; then
-            gh pr merge "$PR" --repo "$REPO" --squash --auto
-            echo "auto-merge armed (maintainer-approved)"
-          else
-            gh pr comment "$PR" --repo "$REPO" --body \
-              "✅ Standards review passed (claude:approve) and auto-merge is staged — this PR will merge automatically the moment a maintainer approves and CI is green." \
-              || true
-            echo "awaiting maintainer approval before arming auto-merge"
-          fi
+          # No `|| true`: a comment that failed to post is a notification that
+          # did not happen, and this repository has paid for silence that
+          # cannot be told from success (defect catalogue section 5).
+          gh pr comment "$PR" --repo "$REPO" --body \
+            "$(printf '%s\n\n<!-- automerge-state: %s -->\n' "$HEADLINE" "$STATE")"
--- a/PR_BODY-safe-harbor.md
+++ b/PR_BODY-safe-harbor.md
@@ -1,0 +1,244 @@
+candidate: remove identifier values from redaction instead of keeping the last four characters (#112)
+
+## What
+
+`r6.redaction.apply_redaction` (the Safe Harbor / standard profile) kept the
+last four characters of every `identifier.value` (`'***' + val[-4:]`). It now
+removes `value` entirely and keeps `system` and `type`, so a reader can still
+see that a patient had an SSN or an MRN of some kind, but not any part of the
+value.
+
+`apply_patient_controlled_redaction` had the same shape one layer down: its
+docstring said the injected `https://healthclaw.io/patient-id` identifier is
+the *sole* identifier on the output, but the code filtered by a keyword
+denylist (`'mrn', 'facility', 'org/', 'example.org'`, plus two hardcoded SSN
+systems) and let anything that didn't match through with its value intact —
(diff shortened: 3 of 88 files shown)


aks129:
## QA review — **FAIL**

Independent review of `77f6d75`, scoped to what #566 adds over its base.
**This PR stacks on #550** (`fix/528-payload-immutable-after-confirmation`), so
everything below is read against that branch, not `main`. `dependency-audit` is
red on every open PR in this repo from npm advisories already fixed in unmerged
**#544** — not this PR's doing and not counted against it.

The four submit branches are well reasoned and the author's mutation table
holds — I applied six of the sixteen and each reddened its named test. The
verdict turns on one thing: the new sentence **"your approval is recorded"** is
reachable today in a state where no approval exists, and the instruction beside
it is false in the same state.

I pushed **tests only** (`8223d65`, `2cb56c8`). I did not touch production
code — the two fixes below need a copy decision that belongs to Dev and the
CTO. Naming that deviation from the task's "fix HIGH/MEDIUM" rather than making
the call quietly.

---

### What I ran

```text
3183 passed, 13 skipped, 3 xfailed, 2 warnings in 106.83s

uv run ruff check .
All checks passed!

  tests/test_conformance_endpoint.py tests/test_conformance_report_never_contradicts_itself.py -q
65 passed in 2.73s          # conformance grade A intact
```

Baseline before my tests: `3181 passed, 13 skipped, 1 xfailed`. The delta is my
two strict xfails plus two passing guards.

Not a unit fake: a real HealthClaw on `:5099` (`uv run flask --app main
init-db`, `PORT=5099`) driven through propose → commit → review → confirm with a
real `requests` client over a real socket, plus the real rendered page in a real
browser at 390×844.

---

## HIGH — "your approval is recorded" and "approving again will not resend it" are both reachable while false

`careagents/healthclaw.py:412-420` — `submit_review` is the one method that opts
out of the rule `_json_object` exists to state, four lines away at `:100-116`:
*"A 200 carrying a proxy's HTML interstitial, or valid JSON of the wrong type,
is a failed call and not data."* It substitutes `{"error": "unexpected
response"}` and **returns the status unchanged**. On a 200 the relay at
`careagents/app.py:1116` then believes the engine saved the review and minted a
confirmation.

**Scenario.** Something between CareAgents and HealthClaw answers the review
POST `200 text/html` — an auth interstitial, a captive portal, a misrouted edge.
The approval-token mint then refuses, and a 4xx lands in the `HealthClawError`
arm at `careagents/app.py:1160-1173`.

**Reproduction** — real socket, real `HealthClawClient`, real engine; a proxy
rewrites only the review POST and forwards the rest to `:5099`:

```text
staged action a3383baa-6439-4875-bed2-b8f73fede385 status: awaiting_confirmation
submit_review -> status=200 body={'error': 'unexpected response'}
relay branch taken: ("confirmed: False -> 'your approval is recorded'", 502)

ActionConfirmation rows for this action: []
QuestionnaireResponse rows saved for this tenant: 0
engine status: awaiting_confirmation
```

Zero confirmations, zero saved reviews. The patient is told *"Your review was
saved and your approval is recorded."* Both clauses false.

The page compounds it, verified in the browser: 502 + `confirmed:false` →
`checkStatus()` → the engine really does answer `awaiting_confirmation` for this
action → the new terminus prints

> Your approval is recorded and this request has not been carried out yet.
> Approving again will not resend it.

**The second sentence is the one that matters.** On this path re-approving
*would* send it — there is nothing on file to 409 against, so the review POST
would go through on a retry. Before #566 this branch said "Nothing has been sent
— please try approving again", which on this path was correct and would have
recovered. #566 replaces it with an instruction not to.

To be precise about what is and is not lost: the button-disable does **not**
survive a reload — I checked, and a reloaded page re-arms once the gate is
re-satisfied. So the state is recoverable; what removes the exit is the copy
telling the patient that recovering is pointless, in both the relay message and
the poll terminus.

**Bound on this finding, tested rather than assumed.** I checked whether the
confirm could go through and push a form out with the patient's decisions
missing. It cannot. `r6/actions/rails/form_fill.py:51-57` refuses without a
`reviewed_qr_id` and returns `needs_review` rather than fabricating one. Driven
end to end against the real engine (commit, no review, mint, confirm):

```text
final status: failed
QuestionnaireResponse rows for this tenant: 0
DocumentReference rows for this tenant: 0
```

No form is produced. **This is an honesty defect, not an H5 violation** — the
guardrail below holds. It is still HIGH: it is a false statement to a patient
about whether their approval exists, on the human-gate surface.

Pinned: `tests/test_healthclaw_transport.py::test_an_unreadable_200_review_is_not_a_saved_review`
(`xfail(strict)`).

```bash
uv run python -m pytest tests/test_healthclaw_transport.py -q -k unreadable_200_review
```

**Suggested fix, and the part that needs your decision.** A one-line change
makes the pin XPASS and breaks nothing else — I applied it and ran the full
suite: `1 failed, 3183 passed`, the single failure being the pin doing its job.

```python
if not isinstance(body, dict):
    if r.status_code == 200:
        raise HealthClawError("review submit returned invalid data", r.status_code)
    body = {"error": "unexpected response"}
```

But that routes to `_REVIEW_UNSUBMITTED`, which asserts *"Nothing has been
approved"* — false in the other direction if the engine processed the review and
a proxy mangled only the response. An unreadable 200 on the review POST is
genuinely **the third answer**, exactly as it already is on the confirm path at
`healthclaw.py:391-397`. Please settle the copy rather than take my one-liner as
complete. The engine's real 200 always carries `reviewed_qr_id` and
`approved_via: 'review-page'` (verified against the running instance), so there
is concrete evidence to require instead of a shape to guess at.

---

## MEDIUM — the poll discards the answer in the one case it exists for

`templates/action_review.html:408-443`, read against `_TRANSITIONS`
(`r6/actions/models.py:31-42`):

| | |
|---|---|
| handled, **impossible** | `sent`, `pending`, `approved` |
| **real, unhandled** | `executing`, `failed`, `needs_review`, `expired`, `unknown` |

Three arms are dead and five real statuses fall through to *"We still cannot
tell whether your approval went through."*

Not theoretical. A real form-fill confirm against the running engine:

```text
approval-token mint: 200
confirm: 502 {"error":"Provider execution failed (provider_not_configured)"...}
action status after confirm: failed
```

The engine's 502 is correctly classified `HealthClawUnconfirmed` → the relay
answers `confirmed: null` → the page takes the `confirmed === null` branch
(#220's third answer, **the reason `checkStatus` exists**) → polls → gets
`failed` → and after four tries tells the patient it cannot tell. Confirmed in
the browser:

> We still cannot tell whether your approval went through, so we have not sent
> it again. Your care team will see this request either way. Please do not
> approve a second time.

The status said it ran and failed. The PR is titled *"a recorded approval is not
an unknown outcome"* and adds the one arm needing an unenforced precondition,
while leaving unhandled the arms that need none — including the status a
form-fill action actually reaches wherever no provider is configured, which is
every deployment today.

Pinned: `tests/test_review_says_what_it_knows.py::test_the_poll_reads_statuses_the_state_machine_can_actually_produce`
(`xfail(strict)`). Both pins redden when fixed — per `docs/agent-task-guide.md`
§6, the fixing PR updates them in the same PR.

---

## MEDIUM — a gateway 502 that speaks JSON still renders red with Approve armed

`templates/action_review.html:365`. Keying on `status === 502 && confirmed ===
false` correctly excludes the two 503s — that reasoning is right and I could not
break it. But a gateway 502 carrying a JSON body with no `confirmed` key falls
past it to the danger arm at `:375`. In the browser:

```text
502 JSON, no `confirmed` key  ->  "Bad Gateway" / alert alert-danger / button ENABLED
```

By the project's own `_upstream_answered` reasoning (`healthclaw.py:119-137`) a
5xx is silence and the confirm may already have run. So this is the #416 shape
surviving: a definite red failure, a machine string as the message, Approve
still armed. An HTML 502 is caught by the new unreadable branch and handled
honestly; a JSON one is not, which is an odd seam to leave adjacent to it.

Scope note: I did not check what a real Railway or edge 502 body looks like here
(HTML vs JSON). The finding is the gap in the branch structure, not a
demonstrated live source for it.

Related, LOW: the branch is strict-equality-brittle. `confirmed: "false"` (a
string) renders the "your approval is recorded" message *inside* `alert-danger`
with the button armed — styling and message contradicting each other. The relay
always sends a real boolean today, so this is robustness, not a live bug.

---

## LOW — the 409 comment claims a coverage it does not have

`templates/action_review.html:314-317` says the pre-check and the seal behind it
are "both answers" arriving at this branch. There is a third, and it is the
ordinary one. `_load_form_fill_action` (`r6/actions/review.py:151-164`) requires
`awaiting_confirmation` and runs **before** `has_confirmation`, so once the first
submit's confirm has claimed the action, a second tap answers 404. Against the
real engine:

```text
A. review submitted, confirm never ran   -> second submit: 409 application/json
B. review submitted AND confirmed        -> second submit: 404 {"error":"Unknown action"}
```

Case B renders red *"Unknown action"* for a request that was approved — the #416
shape this branch exists to remove, one status code over. The button is never
disabled at submit time, so a single tab can produce both.

## LOW — nothing will ever carry out what the new copy says is pending

The 502/`confirmed:false` state leaves the action in `awaiting_confirmation`
with an unconsumed confirmation, and nothing retries the confirm: the reaper's
only `awaiting_confirmation` handler is `_reap_lapsed_approval`
(`r6/ops/routes.py:177-191`), which expires it. So *"has not been carried out
**yet**"* points at a future that will not arrive. The action dies at
`expires_at` and the patient's only notification is a Telegram nudge to propose
it again.

## LOW — the precondition was enforced by a paragraph

`templates/action_review.html:428-436` is a careful and correct argument, and it
was the entire enforcement. `docs/2026-08-02-retro.md` is a list of controls
whose only enforcement was their own description. Now mechanized in `8223d65`:
the `checkStatus` caller set is pinned to the three answers that establish a
confirmation, and a `checkStatus()` added to the generic failure arm — a site no
existing guard mentions — reddens it.

Also fixed there: the forbidden-claims guard
(`test_no_branch_can_call_an_unknown_outcome_a_rejection`) matches
**case-sensitively** against source, so `"Nothing was sent"` as a sentence walks
straight past a list written in lower case — which is exactly how the copy it
was written against read. A case-insensitive sibling now covers it. The
`test_careagents.py` guard already lowercases; only the page guard did not.

---

## What I verified, and what I did not

**Verified.** All five new sentences paint in a real browser at 390×844 with the
button state matching the ruling in each, including that the
unreachable-service branch alone leaves Approve enabled. Page reload, browser
back and a bookmarked GET **never** run the poll (`polls: 0`) — `checkStatus` is
reachable only from the submit handler, so the author's claim holds for all of
those. Six mutations applied and reverted — delete the parse `.catch`; disable
the button when unreachable; key the confirm-failure on `confirmed` alone;
delete the `awaiting_confirmation` arm; render the 409 red; restore "try
approving again" in the relay — each reddened its named test with nothing
silently covering for it. My own two new guards were mutation-tested the same
way, at sites no existing guard mentions. The `_upstream_answered` routing claim
is sound: the existing parametrized tests at
`tests/test_healthclaw_transport.py:404-470` already cover a 4xx the engine
answered, a gateway 5xx, 408, 429, an undecodable 200 and a transport exception
over a real socket, and all pass — every ambiguous case does land on
`HealthClawUnconfirmed`/`null`.

**Not covered.** No Playwright spec is committed — I drove the page through the
MCP browser by hand, so none of the rendering evidence runs in CI, and CI still
has no browser for this template. I did not exercise a true concurrent
double-tap (two submits inside the mint→confirm window); I reached the 409 by the
reload path instead. I did not open a second tab — a fresh load is what I
tested, and I showed that never polls. I did not test the Telegram or dashboard
approve surfaces, and I did not review #550 itself.

**On the human gate.** Spec §2.2 H6 records that `POST /r6/actions/<id>/review`
is gated by `Scope.WRITE`, inside `C_agent`, so an agent holding only its own
credentials can mint the `ActionConfirmation` that this PR's 409 copy reports as
*"You have already approved this request."* I grepped `services/`, `adapters/`,
`skills/`, `hermes/`, `openclaw/` and `api/`: zero hits, and I positive-controlled
the pattern against `careagents/`, where it correctly finds the review POST at
`healthclaw.py:409`. So nothing outside the CareAgents relay reaches that route
today and the sentence holds on the surface that exists. It becomes false the day
an agent surface reaches it, which is R6/H6's own prediction — recording it so
the copy is on the list when #214's closure path moves.

PR #553, opened 2026-09-03
candidate: CareAgents beta batch — hide Telegram tile, CARE_REAL_RECORDS switch, beta banner, canonical-host 308 (#536, #264)

aks129:
## What

Implements the beta batch from the 2026-09-02 council ruling (D3, D6, D7) in `careagents/`, plus the copy that goes with it. One commit; no `r6/` changes; no overlap with #540 (trial `git merge-tree` against its head merges clean).

- **Telegram tile is "coming soon"** (`careagents/templates/home.html`). The `id`s are gone, so `home.js`'s handler has nothing to bind — the tile is a label, not a control. `#tg-surface:hover` CSS removed with it. The `/api/surfaces/telegram` endpoint is untouched.
- **`CARE_REAL_RECORDS = off | allowlist | on`** (`careagents/config.py`, `careagents/connectors.py`, `careagents/app.py`). Default `off`; any other value raises `ConfigError` at boot. `off` renders Fasten / wearable / direct-FHIR as "coming soon" (no consent card, blurb "Not open in this beta — start with the sample records.") and refuses their connect POSTs with 503. `allowlist` consults `CARE_REAL_RECORDS_ALLOWLIST` (comma-separated, trimmed, case-insensitive) via `Config.real_records_open_for(email)`; `on` opens to everyone. Gates **new** connections only — refresh, poll, upload, delete on an existing connection never consult it. `connectors.catalog()` / `start()` take `real_records: bool` with a fail-closed default.
- **Soon-tile click is honest** (`careagents/static/home.js`): a refused POST is shown via `say()` instead of "we'll let you know".
- **Beta banner** (`careagents/templates/_beta_banner.html`, included on `/` and `/home`): "Beta: synthetic records only. Things will break — tell us." (10 words; tests pin < 15).
- **Privacy page** (`templates/privacy.html`, new "CareAgents beta testers" subsection under Data Storage & Retention). Two sentences derived from code: the tenant holds `r6/seed.py`'s one fictional patient (one condition, three observations, one medication, intake questionnaire) plus chat and anything approved, nothing else about the tester; deletion from the hub purges in the same request (`DELETE /api/connections/<id>` → `/internal/purge-tenant`, purge + audit + commit), reports the row count, retains the PHI-free audit trail.
[...]

Issue #536, opened 2026-08-31
Connect Telegram is a dead end: cmd_start discards the pairing code, so the surface stays 'pending' forever


aks129:
CareAgents mints a pairing code and deep-links the user into Telegram. The bot then throws the code away, so pairing can never complete.

## The break, in three lines

**1. CareAgents mints the code and builds the link** — `careagents/app.py:1169`:
```python
deep = (f"https://t.me/{cfg.telegram_bot}?start=care_{code}" ...
```

**2. The bot never reads it.** `grep -n "context.args" openclaw/bot.py` returns exactly one hit, line 615, inside `cmd_curatr`. `cmd_start` does not read `context.args` at all — it calls `_bind_chat_to_tenant(chat_id, username)`, which binds to the **env** `TENANT_ID` (default `desktop-demo`), not to the account that generated the code.

**3. The route that would complete the pairing is dead code.** `POST /api/surfaces/telegram/bind` is defined at `careagents/app.py:1173` and called by **nothing in production** — `grep -rn "surfaces/telegram/bind" openclaw careagents r6 tests` returns the definition plus three calls, all inside `tests/test_careagents.py`.

## What the user experiences

Clicks **Connect Telegram** → Telegram opens → taps **Start** → the bot replies with a message about a shared tenant → the CareAgents tile reads **"pending — finish in Telegram"** for ever. No error anywhere. The tests pass because they call the bind route directly, which is the one thing the real flow never does.

## Second, independent fault

Nothing is currently servicing `@hermes_achilles_1_bot`, the bot the deep link points at:
- Railway `openclaw-bot`: **Offline**, zero SUCCESS deployments, latest REMOVED 2026-06-16
- `getWebhookInfo` → `{"url":"", "pending_update_count":0}` — no webhook, so delivery needs long polling
- `getUpdates?offset=-1&limit=1&timeout=0` → **200**, not 409 — no established poller

So even fixing the payload drop leaves the flow dead until something polls. **Either fault alone is fatal, and only the second is fixable by deploying.**

## Third, design-level

The bot is single-tenant by construction and says so on `/start`: *"This bot reads and writes one tenant, `<TENANT_ID>`, which is shared by everyone using it — not a private space of your own."* Per-account binding needs more than a poller.

## Why this went unnoticed

[...]

Issue #264, opened 2026-08-02
careagents: two live origins, one account store — passkeys break across the DNS cutover (decide before Aug 18)


aks129:
Raised by the CTO during the #258 design review; verified in code.

`careagents/config.py:39-43`:

```python
self.rp_id  = e.get("CARE_RP_ID", "careagents.cloud")
self.origin = (e.get("CARE_ORIGIN") or f"https://{self.rp_id}").rstrip("/")
```

WebAuthn credentials are **origin-bound**. We currently have two live CareAgents deployments sharing one account store:

- `careagents.cloud` — the VPS, where DNS still points, i.e. where real users land
- `careagents-production.up.railway.app` — Railway, the declared target

A passkey registered against one origin **does not work on the other**. So sign-in behaves differently depending on which host a user happens to reach, and any passkey enrolled before the DNS cutover stops working after it — for the same account, with no error the patient can act on.

This matters for the webinar specifically: the passkey path cannot be exercised at all until DNS points at the deployed service, and if we cut over after advocates have enrolled, we silently invalidate their credentials.

**Decide one origin before Aug 18.** The cheapest correct answer is to finish the DNS cutover (already a Phase 2 item per #258), keep `careagents.cloud` as the single `CARE_RP_ID`, and retire the VPS path — rather than run two origins and reconcile them later.

Also worth noting while deciding: the passkey migrated during the Railway move is bound to `careagents.cloud`, so it is testable only after cutover.

Related: #258 (stale deployments, monitoring blind spot), #241 (passkey enrolment reachability).

aks129:
Closed in a lean triage (23 Sep): resolved by the DNS cutover: `careagents.cloud` and the Railway host now serve the same build and store (prod_watch reads both). The abandoned VPS is tracked separately in #624.

PR #569, opened 2026-09-03
test(e2e): a tester's first screen under CARE_REAL_RECORDS=off, in a browser

aks129:
> **Merge order is a hard constraint: #571 must merge before this.**
>
> This branch pins the refusal sentence verbatim, on purpose, because it is the one line a closed-out tester reads. A later commit here updated that pin to the rewritten sentence, which exists only on #571. Measured both ways by the review:
>
> - this branch on its own base: **2 failed, 46 passed** — the two assertions on the refusal string
> - #571's branch with this branch's `e2e/` overlaid, i.e. the post-merge state: **48 passed**
>
> Parts of the description below predate that commit: the sentence quoted under **What**, the run count, and two of the items under **Noticed** which #571 fixes. They are left as written rather than quietly edited, since the review's measurements refer to them.

---

## What

Nine Playwright tests at 390x844 (`e2e/tests/careagents-connect-tiles.spec.ts`), plus a
third `webServer` entry so the suite runs a second CareAgents server in
`CARE_REAL_RECORDS=allowlist` mode. Tests only. No production file changes.

With the switch closed (the fail-closed default):

1. Fasten, wearable and upload tiles render `coming soon` with the beta blurb "Not open
   in this beta — start with the sample records.", `data-soon="1"`, no `data-consent`, a
   dashed border and a CSS-uppercased tag. Each one reads as a label, not as a button
   that promises a connection.
2. Tapping one puts the server's refusal directly under that tile, in the viewport:
   "real-records connect isn't open on this beta deployment yet — start with the sample
   records". Asserted as visible text on a sibling selector, not as a network call. No
   consent card, no provider picker, no popup, no connection row, and the tag keeps
   reading `coming soon` rather than "we'll let you know".
3. The sample tile is still live, carries no tag, and its tap reaches the live branch.
4. The Telegram tile is a `div` with no id, reads `coming soon`, and a tap issues no
   request and changes nothing.
5. Two phone screenshots are attached to the run: the hub as first seen, and the same
   hub after a closed tile is tapped.

With `CARE_REAL_RECORDS=allowlist` on the second server:

6. The allowlisted account sees the three tiles live, each with `data-consent="1"` and
   no closure blurb. A second account on the same deployment sees `coming soon` and gets
   the same refusal on tap.

The beta banner is asserted on the landing page and on the hub, in the viewport, with
its full sentence.

`careagents.spec.ts` keeps every assertion it had. Its four helpers move to
`e2e/tests/careagents-fixtures.ts` (not a `*.spec.ts`, so Playwright does not collect
it) so both specs sign a session in the same way. `codeFromLog` now takes the log path,
because there is more than one CareAgents server to read a code from.

## Why

The review of #553 named this gap in writing:

> **Playwright: not run.** [...] **The gap that leaves:** nothing has confirmed *in a
> browser* that `home.js`'s soon-tile handler renders the 503 text under the tile on a
> phone. The server side is proven (503 body above) and the handler reads correctly, but
> the DOM path is unverified. Worth one manual tap on the deployed build.

Its owner list item 8 asked for one manual tap on a phone. This makes that tap a test
that runs on every PR, and takes the screenshot a human can look at.

The allowlist mode needs its own server because `Config` resolves `CARE_REAL_RECORDS`
once at `create_app()`. One process can only be in one mode, so proving the per-account
difference in a browser needs two. That server sets `FASTEN_PUBLIC_KEY` and
`CARE_WEARABLES_ENABLED`, so the allowlist is the only thing deciding what its two
accounts see. Without those, `catalog()` would answer "soon" for its own reasons and the
test would pass while proving nothing.

## Base note

Stacks on #553. Base is `feat/careagents-beta-batch`, not `main`, and the diff is only
this commit. Merge #553 first. Nothing here changes production code, so it can also be
retargeted at `main` after #553 lands.

## What these tests prove, and what they do not

They prove the browser renders a state: which tiles an account sees, what words are on
them, and what appears under one when a thumb taps it.

They do not prove HealthClaw accepted anything. Both CareAgents servers point
`HEALTHCLAW_BASE` at a dead local port, so no records service exists to accept or reject
a thing. No tenant is created and nothing is ingested. The sample-records tap gets as far
as the call to HealthClaw and stops. That is what makes it a usable discriminator: the
app refuses a closed tile before it calls HealthClaw at all. That distinction is stated in
the spec docstring and in each affected test.

Synthetic `@example.test` addresses only (RFC 2606 reserved). No mail leaves the machine:
the dev-mode stub logs each code to stderr.

## Ran

```
uv run ruff check .
All checks passed!

3178 passed, 13 skipped, 1 xfailed, 2 warnings in 121.14s (0:02:01)

cd e2e && E2E_PORT=5199 npm test
48 passed (25.7s)
```

`E2E_PORT=5199` because :5000 is macOS AirPlay and :5099 held another worktree's server.
CI keeps the default.

Mutation runs, to show the new tests bite. Each mutation was reverted:

| Mutation | Result |
|---|---|
| `home.js` soon-branch ignores a refused POST and sets "we'll let you know" | 3 red: refusal-under-tile, screenshot, unlisted-account |
| harness boots the closed server with `CARE_REAL_RECORDS=on` | 3 red: tiles-coming-soon, refusal-under-tile, screenshot |
| `home.html` restores `id="tg-surface"` / `id="tg-state"` with "connect" | 1 red: Telegram-is-a-label |
| allowlist server booted with an empty `CARE_REAL_RECORDS_ALLOWLIST` | 1 red: allowlisted-sees-live-tiles |

The last one leaves "an unlisted account sees coming soon" green, correctly: an empty
allowlist is closed to everyone, so that account's view does not change.

## What was NOT run

- Nothing ran against a real HealthClaw, a real Fasten key, or the deployed site. See the
  section above.
- No Fasten or wearable flow was started, in either mode. The allowlist test asserts the
  tiles render live and stops there.
- No deploy, no config change on Railway. `CARE_REAL_RECORDS` stays unset there, which is
  closed.
- Chromium only, one viewport (390x844, `isMobile`). No WebKit, no in-app browser, no
  real device. An in-app browser is the stated primary surface and remains untested.
- No passkey path. The spec skips enrolment exactly as `careagents.spec.ts` does.
- I did not run `gh pr checks`. `dependency-audit` is red on every open PR from npm
  advisories whose fix sits in the unmerged #544. Nothing here touches a JS dependency.

## Screenshots

Written by the run to
`e2e/test-results/careagents-connect-tiles-*-chromium/` and attached to the Playwright
report. Both paths are gitignored, so no binary is committed, and CI keeps neither on a
green run: `ci.yml` uploads `playwright-report/` only `if: failure()`. To see the two
images, run the spec locally. Regenerate with:

```
cd e2e && npx playwright test tests/careagents-connect-tiles.spec.ts
```

## Noticed, not touched

- The marketplace footer note still reads "Verified provider & wearable logins happen on
  the provider's own site — we never see your password." It sits directly under three
  tiles that say the beta has not opened those logins. The landing page's step 1 has the
  same problem. Copy for Product to decide, not a switch defect.
- The refusal sentence starts lowercase and has no full stop, so it reads as a fragment
  next to the rest of the hub's copy. Product call.

Refs #553, #536, #264

### DIFF
--- a/e2e/playwright.config.ts
+++ b/e2e/playwright.config.ts
@@ -16,6 +16,22 @@ export const CARE_BASE_URL = `http://localhost:${CARE_PORT}`;
 export const CARE_LOG = '/tmp/e2e-careagents.log';
 const CARE_DB = '/tmp/e2e-careagents.db';
 
+// A THIRD app: the same CareAgents code with CARE_REAL_RECORDS=allowlist
+// (#553, council ruling 2026-09-02 D3). The switch is read once, at
+// create_app(), so one process can only ever be in one mode — proving in a
+// browser that an allowlisted account sees different tiles from a stranger
+// needs its own server, its own SQLite and its own log.
+export const CARE_ALLOW_PORT = process.env.CARE_ALLOW_E2E_PORT || '5102';
+export const CARE_ALLOW_BASE_URL = `http://localhost:${CARE_ALLOW_PORT}`;
+export const CARE_ALLOW_LOG = '/tmp/e2e-careagents-allow.log';
+const CARE_ALLOW_DB = '/tmp/e2e-careagents-allow.db';
+// Fixed, and synthetic. The allowlist is server env, so the address the spec
+// signs in with has to be known before the server boots — it cannot be the
+// per-test unique address the other specs use. `.test` is reserved (RFC 2606)
+// and can never be a real mailbox; nothing is ever sent anyway (the dev mail
+// stub logs to stderr).
+export const CARE_ALLOW_EMAIL = 'e2e-allowlisted@example.test';
+
 export default defineConfig({
   testDir: './tests',
   // Serial execution — shared SQLite instance, tests create resources
@@ -98,6 +114,44 @@ export default defineConfig({
         // Explicitly blank so a developer shell with a real Resend key still
         // gets the stderr mail stub — codes must never leave the machine.
         RESEND_API_KEY: '',
+        // The beta's real-record switch, pinned CLOSED for this server (#553,
+        // D3). Blank resolves to `off` exactly as unset does; the *unset*
+        // default is pinned by pytest (test_careagents.py::
+        // test_real_records_switch_defaults_to_off_and_parses_the_allowlist),
+        // and pinning it here stops a developer shell that exports
+        // CARE_REAL_RECORDS=on from quietly running the browser suite against
+        // the open path while its name still says closed.
+        CARE_REAL_RECORDS: '',
+        CARE_REAL_RECORDS_ALLOWLIST: '',
+      },
+    },
+    {
+      // CareAgents again, in allowlist mode — see CARE_ALLOW_* above.
+      command:
+        `rm -f ${CARE_ALLOW_DB} ${CARE_ALLOW_LOG} && cd .. && ` +
+        `uv run flask --app careagents.wsgi run --port ${CARE_ALLOW_PORT} ` +
+        `2>> ${CARE_ALLOW_LOG}`,
+      url: `${CARE_ALLOW_BASE_URL}/`,
+      reuseExistingServer: !process.env.CI,
+      timeout: 60_000,
+      env: {
+        CARE_ENV: 'development',
+        CARE_DATABASE_URL: `sqlite:///${CARE_ALLOW_DB}`,
+        // Dead port, for the same reason as the entry above.
+        HEALTHCLAW_BASE: 'http://127.0.0.1:9',
+        CARE_ORIGIN: CARE_ALLOW_BASE_URL,
+        CARE_RP_ID: 'localhost',
+        RESEND_API_KEY: '',
+        CARE_REAL_RECORDS: 'allowlist',
+        CARE_REAL_RECORDS_ALLOWLIST: CARE_ALLOW_EMAIL,
+        // Fasten key present and the wearables sidecar declared wired, so the
+        // ONLY thing between an account and a live real-record tile here is
+        // the allowlist. Without these two, `catalog()` would answer "soon"
+        // for its own reasons ("not configured on this deployment") and the
+        // allowlist test would pass while proving nothing. The key is never
+        // used: nothing in the spec starts a Fasten flow.
+        FASTEN_PUBLIC_KEY: 'e2e-not-a-real-fasten-key',
+        CARE_WEARABLES_ENABLED: '1',
       },
     },
   ],
--- a/e2e/tests/careagents-connect-tiles.spec.ts
+++ b/e2e/tests/careagents-connect-tiles.spec.ts
@@ -48,10 +48,10 @@ const PHONE = {
 /** The three sources that take a person's OWN records (connectors.py:74). */
 const REAL_RECORD_TILES = ['fasten', 'wearable', 'direct'] as const;
 
-/** connectors.py:75 — verbatim. This is the sentence a tester reads. */
+/** connectors.py:78-79 — verbatim. This is the sentence a tester reads. */
 const REFUSED =
-  "real-records connect isn't open on this beta deployment yet — " +
-  'start with the sample records';
+  "Connecting your own records isn't open in this beta yet. " +
+  'Start with the sample records.';
 
 /** connectors.py:94 — the blurb that replaces the live one when closed. */
 const CLOSED_BLURB = 'Not open in this beta — start with the sample records.';
--- a/e2e/tests/careagents-connect-tiles.spec.ts
+++ b/e2e/tests/careagents-connect-tiles.spec.ts
@@ -1,0 +1,299 @@
+import { test, expect, Page } from '@playwright/test';
+import {
+  CARE_ALLOW_BASE_URL, CARE_ALLOW_EMAIL, CARE_ALLOW_LOG,
+  CARE_BASE_URL, CARE_LOG,
+} from '../playwright.config';
+import { blockThirdParty, signIn, uniqueEmail } from './careagents-fixtures';
+
+/**
+ * A beta tester's first screen, in a browser (#553, council ruling
+ * 2026-09-02 D3/D6).
+ *
+ * The review of #553 proved the server side of the CARE_REAL_RECORDS switch —
+ * the catalog JSON, the 503 on a closed connect POST — and left one gap in
+ * writing: "nothing has confirmed *in a browser* that home.js's soon-tile
+ * handler renders the 503 text under the tile on a phone." That is what this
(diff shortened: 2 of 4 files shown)


aks129:
## QA review — **FAIL** at `12cf19b`, on merge order alone

The nine tests are honest and they bite. I mutated the product four ways and the
right ones reddened, including one you did not try. Nothing here is a defect in
the test code.

The failure is that **this branch does not pass on the base it declares**, and
the description no longer matches what the branch contains. Measured, not
inferred:

```
$ cd e2e && E2E_PORT=5711 CARE_E2E_PORT=5712 CARE_ALLOW_E2E_PORT=5713 npx playwright test
  2 failed, 46 passed
    [chromium] careagents-connect-tiles.spec.ts:119  tapping a closed tile shows the refusal under that tile
    [chromium] careagents-connect-tiles.spec.ts:282  an account that is not on the list sees "coming soon" on the same deployment

  Expected: "Connecting your own records isn't open in this beta yet. Start with the sample records."
  Received: "real-records connect isn't open on this beta deployment yet — start with the sample records"
```

`12cf19b` ("follow the refusal sentence to its rewritten form") pins the string
that only exists on **#571**. So there is no longer a choice of order.

**Merge order: #553 → #571 → #569.** Verified green in that state — #571's
branch with this PR's `e2e/` overlaid:

```
  48 passed (27.6s)
```

**Not merged. Auto-merge not armed** (`autoMergeRequest: null`, checked before
and after). **I did not run `gh pr checks`.**

---

### MEDIUM — the description would walk a maintainer into that red

Three things in it are now stale, and each on its own is enough:

- **"What", item 2** still quotes the old refusal verbatim as the string the
  test asserts. The branch asserts the new one
  (`careagents-connect-tiles.spec.ts:52-55`).
- **"Ran": `48 passed`** predates `12cf19b`. The branch as it stands is
  `2 failed, 46 passed`.
- **"Noticed, not touched"** lists the marketplace footer note, the landing
  page's step 1 and the lowercase refusal fragment as open. All three are fixed
  in #571, which is now a hard prerequisite rather than a sibling.

**Also:** neither this PR nor #571 has had any CI. The pipeline does not run on
stacked pull requests — six checks, none of them a test, against nineteen for a
change targeting `main` (**#585**). If this merges into
`feat/careagents-beta-batch` ahead of #571, nothing on GitHub will say the batch
branch went red; the first signal would be a confusing failure on #553 against
`main`.

*Fix:* merge #571 first, then correct those three sections. No code change.

---

### Mutations — verified, including one you did not run

Each applied to the post-merge tree and restored afterwards. Baseline for the
spec alone is `9 passed`.

| mutation | reported | measured |
|---|---|---|
| `home.js` soon-branch ignores a refused POST | 3 red | **3 red — the same three** (refusal-under-tile, screenshot, unlisted-account) |
| `home.html` restores `id="tg-surface"` / `id="tg-state"` | 1 red | **1 red** (Telegram-is-a-label), on `toHaveText('coming soon')` receiving `"connect →"` |
| **mine:** drop `el.scrollIntoView()` from `say()` | — | **1 red**, `toBeInViewport` failing with `viewport ratio 0` |
| **mine:** revert `say()` to writing text before unhiding | — | **9 passed** |

The third one matters because it was the assertion I most expected to be
decoration: Playwright's `click()` auto-scrolls the tile into view, so a message
placed directly beneath it could plausibly be on screen for free. It is not.
`toBeInViewport` is doing real work, and it is the assertion that speaks for the
phone.

The fourth is a coverage boundary rather than a defect. Reverting the
accessibility fix #571 ships — `say()` writing into a `display:none` region, the
change no screen reader can hear — leaves all nine of these green. Correct
(this suite predates that fix and asserts rendered state), but the docstring's
"WHAT THEY DO NOT PROVE" does not say it, and it is the kind of thing a later
reader will assume a browser test covers. Worth one line there.

---

### The two-server arrangement — genuinely separate

The premise checks out. `careagents/wsgi.py` calls `create_app()` once at import,
`Config.__init__` reads `CARE_REAL_RECORDS` once (`config.py:37`, `:214`), and
`real_records_open_for()` is the only per-request part. One process, one mode.

Isolation, tested rather than assumed. Separate SQLite, separate log, separate
port, and account ids are random per database
(`models.py:36`, `default=lambda: _uid("acct")`), so they cannot collide across
the two. I minted a session on one server and replayed it against a second built
the same way:

```
cookie from :5701 replayed at :5702/home  -> 302     (signed out)
same cookie at its own :5701/home         -> 200
no cookie at :5702/home                   -> 302
```

Same answer as no cookie at all. No shared authenticated state.

One thing worth knowing, since it is not obvious: the cookie's **signature does**
verify on the other server. `config.py:214` falls back to a fixed
`"dev-careagents-secret"` outside production, and browser cookies are not
port-scoped, so `localhost:5712` and `localhost:5713` share one cookie jar. Each
test gets a fresh Playwright context, so nothing in this spec touches both — but
a future test that visits both servers in one context will be silently signed
out on the second, and that reads as an app bug rather than a harness artifact.

`CARE_ALLOW_EMAIL` being fixed rather than unique is fine: under a CI retry the
30s resend cooldown answers `{"sent": false}`, `auth.js:76` still advances to
the code step, and `codeFromLog` takes the **last** match in the log, which is
the previous code and still live.

### LOW

- **A fourth hardcoded `/tmp` path.** `CARE_ALLOW_DB` / `CARE_ALLOW_LOG` follow
  the existing convention, and the boot command `rm -f`s them. Two agents running
  e2e in different worktrees on this machine unlink each other's SQLite mid-run
  while the ports are happily distinct. Deriving the paths from the port would
  cost one line. Not introduced by you; now four files instead of two.

---

### What I ran

```
$ uv run ruff check .
All checks passed!

3182 passed, 13 skipped, 1 xfailed, 2 warnings in 140.39s

$ uv run python -m pytest tests/test_ratchets.py tests/test_guardrail_conformance.py -q -p no:cacheprovider
50 passed in 2.68s        # Grade A release gate intact
```

Ports 5711 / 5712 / 5713 (5000 is AirPlay). Servers killed between mutation runs
so each booted against the files on disk — with `reuseExistingServer` on, a
survivor serves cached Jinja templates and hands you a false green.

### Not covered

- **Nothing against a running HealthClaw**, exactly as your description says.
  Confirmed the point is made where a reader meets it: the spec's own docstring,
  the sample-tile test's inline comment, and the PR body.
- **Chromium only, 390x844.** No WebKit, no in-app browser, no real device.
- **No screen reader**, so the a11y half of #571 is verified structurally and in
  the accessibility tree, never heard. Details on #571.
- I did not re-audit the #553 batch underneath this.

Issue #269, opened 2026-08-02
careagents: the page scrolls behind an open modal when a background request fails


aks129:
Found by QA re-verifying #224. **Introduced by that fix**, LOW severity, deliberately not blocked on.

`careagents/static/home.js:104` — `say()` now calls `scrollIntoView` to bring the message next to the element it is about. It does so unconditionally, including when a modal is already open over the page.

Reachable sequence: tap a failing tile → while its POST is in flight, tap the wearables tile (the picker opens) → scroll the background (there is no scroll lock, deliberately) → the first POST fails and its message scrolls the page underneath. Measured: `scrollY` jumped **1284 → 628** with the picker open.

Consequences are cosmetic. The fixed picker card does not move (`cardTop` 313 before and after), all rows survive, and it remains fully functional. But the user is ~656px from where they left the page when they dismiss the modal, and an error about a *different* tile is waiting behind it.

Suggested guard, if this code is touched again:

```js
if (!document.querySelector('.modal:not([hidden])')) el.scrollIntoView({block: "nearest"});
```

Repro: hold `**/api/connections/sample` for 3s, tap the sample tile, tap the wearable tile, `window.scrollTo(0, 1284)`, wait for the response, read `scrollY`.

Related, INFO only: inserting the message pushes tiles *below* it down 55px. The tapped tile itself does not move, so the risk of a mis-tap is low, but a second tap within about a second could land on a neighbour that shifted.

PR #571, opened 2026-09-03
fix(careagents): a refused tile now announces itself, and the reassurance stops contradicting the tiles

aks129:
## What

Three fixes to the beta hub, all found by the browser pass over #553.

**1. A refused tile now announces itself.** `#connect-msg` had no `role="status"`
and no `aria-live`, while its sibling `.conn-refresh-msg` (home.html:75) has
carried both since the refresh flow landed. #553 made `#connect-msg` the carrier
of the one sentence a tester gets when a closed tile refuses them, so a blind
tester tapped a tile and was told nothing at all. The element now carries the
sibling's exact attributes.

The attributes alone would not have fixed it. A live region announces only a
change made while it is *in* the accessibility tree, and `say()` (home.js:101)
did two things that take it out — `hidden` is `display:none`, and the element is
MOVED next to the tapped tile, which is a remove-then-insert — in the same task
as the write. `say()` now clears the region, unhides it, lets a frame paint
(nested `requestAnimationFrame`; a single one runs *before* the paint, which is
the same tick as far as the a11y tree is concerned), then writes the text. The
scroll nudge moved into the deferred callback so `block: "nearest"` measures the
final height.

**The sibling likely needs the same treatment.** `.conn-refresh-msg` has the
attributes but its writer (`report()`, home.js:252) sets `textContent` *before*
`hidden = false` — the same ordering defect, on the element this fix was modelled
on. Not touched here; it belongs to the refresh flow, not this batch.

**2. The reassurance no longer contradicts the tiles.** "Verified provider &
wearable logins happen on the provider's own site — we never see your password"
rendered directly beneath three tiles saying those logins are not open in this
beta. The sentence is true and is kept verbatim, now rendered only when the
catalog says the tiles it describes are live.

**3. The refusal reads as a sentence.** It was lowercase, unpunctuated and
em-dashed.

## Base note

**This stacks on #553** (`feat/careagents-beta-batch`), not on `main` — one
commit here; review it, not the diff against `main`. **#569 also
stacks on that same base and needs a one-line string update**; details and merge
order below.

## The new string — **#569 needs a one-line update**

```
Connecting your own records isn't open in this beta yet. Start with the sample records.
```

(was: `real-records connect isn't open on this beta deployment yet — start with
the sample records`)

**#569 stacks on this same base and asserts the old string verbatim** as its
`REFUSED` constant (`e2e/tests/careagents-connect-tiles.spec.ts:52`). That file
was deliberately not edited here. Two of its assertions fail against this branch,
both on that constant and nothing else:

- `tapping a closed tile shows the refusal under that tile` (:127)
- `an account that is not on the list sees "coming soon" on the same deployment` (:295)

Whoever merges second updates `REFUSED` to the string above. **Merge order:** if
#569 goes first, this PR's change makes it red until that one line lands; if this
goes first, #569 needs the line updated before it can go green. The other seven
tests in that spec pass against this branch — including `toBeVisible`,
`toBeInViewport` and the sibling-selector locator, so the `say()` rework does not
disturb them.

No existing Python test pinned the string exactly — the closest
(`test_real_record_connect_posts_are_refused_with_503_when_off`) asserts
`"beta" in error`, which passes either way. A new test pins the new sentence on
the wire for all three closed tiles.

`connectors.py:158/225` carry a *different* message ("isn't configured on this
deployment yet") for an unconfigured Fasten key — deliberately left alone.

## Why

The browser pass over #553. #553 shipped the CARE_REAL_RECORDS switch and proved
its server side; what it could not see was the screen. All three of these are
only visible with a rendered page and, for the first, a screen reader.

## Two judgement calls worth reviewing

- **The gate is `fasten`/`wearable`, not all three real-record tiles.** The
  sentence is about provider and wearable *logins*. `direct` (upload a FHIR
  bundle) is a real-record source but not a login, so a deployment with the
  switch on but no Fasten key and no wearables sidecar would still have shown the
  sentence under two "not configured" tiles. Strict subset of the brief's
  condition, same goal.
- **The landing page's clause is deleted, not made conditional.** `landing()`
  (app.py:240) passes only `me`; the page is public and gets no account, so it
  cannot resolve a switch that is evaluated per account. Step 1 now ends "Then
  connect your records."

## Ran

- `uv run ruff check .` → `All checks passed!`
- `uv run python -m pytest tests/ -q -p no:cacheprovider` →
  `3182 passed, 13 skipped, 1 xfailed, 2 warnings in 104.53s`
- `cd e2e && npm test` → `39 passed (21.6s)`
- #569's spec, overlaid from its branch and then removed → `2 failed, 7 passed`,
  the two named above.

Ports: `E2E_PORT=5297 CARE_E2E_PORT=5301` (5000 is AirPlay; 5099 and 5199 were
held by other agents on this machine — with `reuseExistingServer` on, 5199's
foreign server silently gave one false red before the ports were moved).

## What was NOT run

- No screen reader. The announcement is verified structurally only: the
  attributes on the tag, and the unhide-before-write ordering inside `say()`
  (`tests/test_careagents.py::test_the_connect_refusal_is_announced_to_a_screen_reader`).
  Someone with VoiceOver or NVDA should confirm the refusal is actually spoken.
- Nothing against a running HealthClaw. The e2e harness pins `HEALTHCLAW_BASE` to
  a dead port; no ids crossed the boundary here, and none needed to.
- Nothing deployed.

`dependency-audit` is red on every open PR from npm advisories already fixed in
unmerged PR #544.

### DIFF
--- a/careagents/connectors.py
+++ b/careagents/connectors.py
@@ -72,8 +72,11 @@ _WEARABLE_IDS = {p["id"] for p in WEARABLE_PROVIDERS}
 # unless the deployment's CARE_REAL_RECORDS switch opens them for this
 # account (council ruling 2026-09-02, D3) — the beta runs on synthetic data.
 REAL_RECORD_SOURCES = ("fasten", "wearable", "direct")
-_REAL_RECORDS_CLOSED = ("real-records connect isn't open on this beta "
-                        "deployment yet — start with the sample records")
+# The one sentence a refused tester reads, so it is a sentence: patient-facing
+# copy here is plain, sentence-cased and em-dash-free. Parallel to the closed
+# tile's own blurb ("Not open in this beta — start with the sample records.").
+_REAL_RECORDS_CLOSED = ("Connecting your own records isn't open in this beta "
+                        "yet. Start with the sample records.")
 
 
 def catalog(cfg, real_records: bool = False) -> list[dict]:
--- a/careagents/static/home.js
+++ b/careagents/static/home.js
@@ -102,17 +102,30 @@
     if (anchor && el.previousElementSibling !== anchor) {
       anchor.insertAdjacentElement("afterend", el);
     }
-    el.textContent = text;
+    // #connect-msg is a live region, and a live region announces only a change
+    // made while it is IN the accessibility tree. Both things this function
+    // does take it out: `hidden` is display:none, and the move above is a
+    // remove-then-insert. Setting the text in the same task as either one is a
+    // change nothing is listening for — the refusal a closed tile gives would
+    // be read out by no screen reader at all. So show the empty region, let a
+    // frame paint, then write into it. Nested rAF because a single one runs
+    // BEFORE that paint, which is the same task as far as the a11y tree is
+    // concerned; two frames is ~32ms, under the threshold where an empty box
+    // is perceptible.
+    el.textContent = "";
     el.hidden = false;
-    // Only scrolls if it isn't already fully visible, and only as far as it
-    // has to — no jump when the message is already under the user's thumb.
-    // Never while a dialog is up: the message is behind the overlay, so the
-    // scroll moves nothing the user can see and everything they come back to
-    // (#269). Open state is the absence of `hidden` — that attribute is what
-    // openDialog() and the consent card toggle.
-    if (!document.querySelector(".modal:not([hidden])")) {
-      el.scrollIntoView({ block: "nearest" });
-    }
+    requestAnimationFrame(() => requestAnimationFrame(() => {
+      el.textContent = text;
+      // Only scrolls if it isn't already fully visible, and only as far as it
+      // has to — no jump when the message is already under the user's thumb.
+      // Never while a dialog is up: the message is behind the overlay, so the
+      // scroll moves nothing the user can see and everything they come back to
+      // (#269). Open state is the absence of `hidden` — that attribute is what
+      // openDialog() and the consent card toggle.
+      if (!document.querySelector(".modal:not([hidden])")) {
+        el.scrollIntoView({ block: "nearest" });
+      }
+    }));
   }
 
   // Scroll a section into view and pulse it — the in-page way to point at the
--- a/careagents/templates/home.html
+++ b/careagents/templates/home.html
@@ -92,9 +92,19 @@
       </button>
       {% endfor %}
     </div>
-    <p class="inline-msg" id="connect-msg" hidden></p>
+    {# The only channel for a refusal: tap a closed tile and this paragraph is
+       the whole of what the deployment says back. Same live-region treatment
+       as .conn-refresh-msg above; say() shows it before writing into it. #}
+    <p class="inline-msg" id="connect-msg" role="status" aria-live="polite"
+       hidden></p>
+    {# True, and worth keeping — but only where the logins it describes are
+       open. Under three "not open in this beta" tiles it contradicted the
+       screen it sat on. #}
+    {% if catalog | selectattr('id', 'in', ['fasten', 'wearable'])
+                  | rejectattr('tier', 'equalto', 'soon') | list %}
     <p class="add-note">Verified provider &amp; wearable logins happen on the
       provider’s own site — we never see your password.</p>
+    {% endif %}
   </section>
 
   <!-- SURFACES -->
--- a/careagents/templates/landing.html
+++ b/careagents/templates/landing.html
@@ -35,9 +35,11 @@
     <li>
       <span class="step-n">1</span>
       <h3>Sign in with your face</h3>
+      {# The provider-login clause is gone rather than made conditional: this
+         page is public, gets no account, and so cannot resolve the
+         CARE_REAL_RECORDS switch that decides whether those logins are open. #}
       <p>A passkey — Face ID, Touch ID, or your security key — is your login.
-      No password to phish or forget. Then connect your records through your
-      provider’s own verified login.</p>
+      No password to phish or forget. Then connect your records.</p>
     </li>
     <li>
       <span class="step-n">2</span>
--- a/tests/test_careagents.py
+++ b/tests/test_careagents.py
@@ -5238,3 +5238,88 @@ def test_the_tester_guide_matches_what_the_switch_actually_does():
     assert "Delete the account" in leaving
     assert "no self-serve" in leaving
     assert "support@healthclaw.io" in leaving
+
+
+# --- browser pass over the beta batch (#553) ---------------------------------
+# Three defects a screen reader and a pair of eyes found on the hub that no
+# server-side assertion could see.
+
+def test_the_connect_refusal_is_announced_to_a_screen_reader():
+    """`#connect-msg` is the ONLY channel for a refusal: tap a closed tile and
+    this paragraph is the whole of what the deployment says back. Without the
+    live-region attributes a blind tester taps and is told nothing at all.
+
+    Its sibling `.conn-refresh-msg` has carried `role="status"` +
+    `aria-live="polite"` since the refresh flow landed; this is the same
+    treatment on the same kind of element.
+    """
+    tag = _HOME_HTML[_HOME_HTML.index('<p class="inline-msg" id="connect-msg"'):]
+    tag = tag[:tag.index(">") + 1]
+    assert 'role="status"' in tag, tag
+    assert 'aria-live="polite"' in tag, tag
+
+    # ...and the attributes alone are not the fix. A live region announces a
+    # change made while it is IN the accessibility tree; `hidden` is
+    # display:none, and say() also MOVES the element (a remove-then-insert).
+    # Writing the text in the same task as either one is a change nothing is
+    # listening for. So the region is shown first and written into after.
+    body = re.sub(r"//[^\n]*", "",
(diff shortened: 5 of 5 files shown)


aks129:
## QA review — **PASS** at `2d083ad`, with three test gaps fixed on the branch at `4ddc098`

All three product changes do what the description says, and I confirmed the
announcement one in a real browser rather than only in the source. Two of the
three new tests passed for a reason other than the property they name; those
are fixed here, not sent back. No production file was touched by me.

**Not merged. Auto-merge not armed** (`autoMergeRequest: null` on both PRs,
checked before and after). **I did not run `gh pr checks`.**

---

### First, the thing a maintainer has to act on

**No CI has run on this PR, or on #569.** The pipeline does not run on stacked
pull requests — six checks, none of them a test, against nineteen for a change
targeting `main`. That is filed as **#585**. My local run below is the only test
evidence either branch has.

**Merge order is now hard: this PR first, then #569.** The "if #569 goes first"
half of the Base note is out of date. #569's head (`12cf19b`, *"follow the
refusal sentence to its rewritten form"*) already pins the **new** string, so it
is red against its own base today and goes green only once this lands. Measured
both ways:

```
# 569's branch as it stands  (= base + its test files)
cd e2e && E2E_PORT=5711 CARE_E2E_PORT=5712 CARE_ALLOW_E2E_PORT=5713 npx playwright test
  2 failed, 46 passed
    tapping a closed tile shows the refusal under that tile
    an account that is not on the list sees "coming soon" on the same deployment

# this branch + 569's e2e/ overlaid  (the post-merge state)
  48 passed (27.6s)
```

---

### The announcement — verified in a browser, not just in the diff

Structural first: the tag carries `role="status"` and `aria-live="polite"`, and
`say()` unhides before it writes. Then I drove a real Chromium at 390x844
against a running CareAgents (`CARE_REAL_RECORDS` closed, `HEALTHCLAW_BASE` on a
dead port, synthetic `@example.test` address), with a `MutationObserver` and a
rAF frame counter installed before the tap:

```
kind                  frame  text            hidden  connected  display  role    aria-live  prev sibling  rect
at-load                   0  ""              true    true       none     status  polite     .marketplace  0x0
attributes:hidden       731  ""              false   true       block    status  polite     fasten        351x20
parent:removed-msg      731  ""              false   true       block    status  polite     fasten        351x20
parent:inserted-msg     731  ""              false   true       block    status  polite     fasten        351x20
childList               733  "Connecting…"   false   true       block    status  polite     fasten        351x62
```

The move and the unhide both land on frame 731 with the region empty and
rendered; the text arrives on frame **733**, two frames later, with the element
connected and `display: block`. That is the ordering the fix claims. The
accessibility tree agrees — a snapshot of `#connect-msg` after the tap:

```
- status: Connecting your own records isn't open in this beta yet. Start with the sample records.
```

**What that is not:** it is not a screen reader. Nobody heard this. Two frames
is ~32ms on a 60Hz phone, which is under the delay some ATs (NVDA in
particular) reliably pick up after a region becomes visible, and the double-rAF
does not run at all in a backgrounded tab. The structure is right; whether it is
*spoken* still needs someone with VoiceOver or NVDA. Your "What was NOT run"
section says this, and I am confirming it rather than contradicting it.

**Both of your claims about the sibling hold.** `.conn-refresh-msg`
(home.html:75) has the attributes and its writers do not use them: `report()`
(home.js:252), `sayUpload()` (:322), the pending-card poller (:230), disconnect
(:414) and delete (:431) all set `textContent` before `hidden = false`. Because
the element starts `hidden`, the **first** message on a card is written into a
`display:none` node and announces nothing; later messages on the same card do
announce, since the region is already in the tree by then. Correctly left out of
this batch.

---

### MEDIUM — three mutations survived, all fixed at `4ddc098`

Each was applied to the branch as pushed and restored afterwards.

| mutation | before | after `4ddc098` |
|---|---|---|
| a **single** rAF instead of the nested pair | 223 passed | **red** |
| gate widened to include the `direct` tile | 223 passed | **red** |
| the reassurance paragraph deleted outright | 223 passed | **red** |

**1. `assert "requestAnimationFrame" in body` is satisfied by one rAF** —
`test_the_connect_refusal_is_announced_to_a_screen_reader`, tests/test_careagents.py:4908.
One frame is exactly the case your own comment says does not announce, because a
single frame's callback runs *before* the paint. The check passed for the broken
behaviour. Now pinned on the count and the nesting.

**2. The reassurance test matched a phrase the live Fasten tile also carries.**
`"never see your password"` is the last clause of the live Fasten blurb
(connectors.py:37-39), so the positive assertion was satisfied by the tile, not
by the note. Deleting the `<p class="add-note">` outright left the suite green
at 223 passed. Keyed on the element now.

**3. The judgement call the description defends at length was pinned by
nothing.** You argue for `fasten`/`wearable` over all three real-record sources
because "a deployment with the switch on but no Fasten key and no wearables
sidecar would still have shown the sentence". Correct — and untested. Widening
the gate back to `['fasten', 'wearable', 'direct']` left every test green. New
test `test_the_password_reassurance_needs_a_login_tile_not_just_an_open_switch`
pins that row and the "one configured login is enough" row, and asserts the tier
arrangement so a future default change cannot make it vacuous.

The other five mutations were already pinned and went red as they should:
attributes stripped, `say()` write order reverted, conditional dropped, refusal
string reverted, landing clause restored.

---

### The conditional — every combination, rendered

Server-side, through the test client, reading the real `/home` body and the real
catalog. `add-note` is the reassurance paragraph:

```
off  (switch closed)             fasten=soon   wearable=soon   direct=soon    add-note=hidden
on   fasten+wearables wired      fasten=live   wearable=live   direct=import  add-note=SHOWN
on   fasten only (no sidecar)    fasten=live   wearable=soon   direct=import  add-note=SHOWN
on   wearables only (no key)     fasten=soon   wearable=live   direct=import  add-note=SHOWN
on   neither configured          fasten=soon   wearable=soon   direct=import  add-note=hidden
allowlist  listed account        fasten=live   wearable=live   direct=import  add-note=SHOWN
allowlist  unlisted account      fasten=soon   wearable=soon   direct=soon    add-note=hidden
allowlist  empty list            fasten=soon   wearable=soon   direct=soon    add-note=hidden
```

The gate is right, including the "switch open, neither login configured" row the
description names. The Jinja `in` test resolves as intended on dicts.

---

### LOW

- **The mixed rows still over-scope the sentence** (rows 3 and 4 above). With
  the Fasten key set and no wearables sidecar — which may well be production for
  the one allowlisted clinician — the page says "Verified provider **& wearable**
  logins happen on the provider's own site" directly above a wearable tile
  reading "Open Wearables sidecar not wired here yet". Milder than what you
  fixed, and the sentence is still true of the tile that is live. Product's call
  whether the copy should name only the open one.
- **`#surfaces-msg` (home.html:126) has no live-region attributes at all.** Same
  element class, same page, and `say()` now defers writes into it too. It is a
  live refusal channel whenever `imessage_handle` is configured
  (home.js:561), plus every `flashSection()` message. The description says
  `#connect-msg` "had no `role=status` … while its sibling has carried both",
  which reads as though it were the only one. There are two.
- **A "not configured" Fasten tile keeps the live blurb.** `catalog()` sets
  `tier="soon"` and the note for a missing `FASTEN_PUBLIC_KEY` but does not
  replace `blurb` (connectors.py:110-112), so that tile reads "Log in to your
  clinic or hospital portal. Verified; we never see your password." under a
  "not configured on this deployment" tag. Pre-existing, not this PR.
- **The new comment overstates its own rule.** connectors.py:75-77 says
  patient-facing copy here is "em-dash-free" and cites the closed tile's blurb as
  parallel — that blurb is "Not open in this beta — start with the sample
  records." with an em dash. The rewritten sentence is good; the rule around it
  is not the rule the code follows.
- **`say()` now evaluates the open-modal check two frames late.** The #269 guard
  and its pin still hold, but the decision is made ~32ms after the call, so a
  modal that opens or closes inside that window flips the outcome. No path in
  the current code does that; noting it because the pin cannot see it.

---

### What I ran

```
$ uv run ruff check .
All checks passed!

3182 passed, 13 skipped, 1 xfailed, 2 warnings in 140.39s     # at 2d083ad
3183 passed, 13 skipped, 1 xfailed, 2 warnings in 120.16s     # at 4ddc098

$ uv run python -m pytest tests/test_ratchets.py tests/test_guardrail_conformance.py -q -p no:cacheprovider
50 passed in 2.68s        # Grade A release gate intact

$ cd e2e && E2E_PORT=5711 CARE_E2E_PORT=5712 CARE_ALLOW_E2E_PORT=5713 npx playwright test
48 passed (27.6s)         # this branch + #569's e2e/ overlaid
```

Browser work against a running CareAgents on :5701, `CARE_REAL_RECORDS` closed,
`HEALTHCLAW_BASE=http://127.0.0.1:9`, synthetic addresses, no PHI.

### Not covered

- **No screen reader.** See above. This is the one gap that matters.
- **Chromium only, one viewport.** No WebKit, no in-app browser, no real device
  — and an in-app browser is this product's primary surface.
- **Nothing against a running HealthClaw**, and nothing deployed. No ids crossed
  the boundary here and none needed to.
- **The `.conn-refresh-msg` first-message defect is confirmed by reading, not by
  driving it** — that needs an existing connection card, which needs a records
  service.
- I did not re-audit the #553 batch underneath this; that review stands.

PR #573, opened 2026-09-03
probe: what actually happens to /healthz under concurrent chat turns (#219)

aks129:
Council ruling D2 item 8 asked for a measurement, not a fix: *"#219 probe (10
concurrent). Build only if it fails."* This is the measurement. **No production
code is changed, and no load was sent to production** — `careagents.cloud`
serves a real clinician, so everything here runs against a local gunicorn and a
fake HealthClaw on loopback.

## Verdict: reproduces differently

The starvation half is real and starts earlier than the issue says. The restart
half does not happen at all, and most of the data-loss half was already false.

| # | Claim in #219 | Measured |
|---|---|---|
| 1 | "One worker, 8 threads" | **Wrong.** `--workers 2 --threads 4` (`deploy/careagents/Dockerfile:54`). Same 8 threads, but the split makes saturation start *earlier* and *non-deterministically* |
| 2 | "6 tool rounds × 25s + 90s LLM" holds the thread | **Wrong mechanism, right symptom.** The web tier runs no inference since #257. The hold is the SSE replay loop (`app.py:833`) |
| 3 | "8 concurrent turns block everything including `/healthz`" | **Confirmed at 8; intermittent from 6.** At the code's own worst case `/healthz` waits **118s** |
| 4 | "Railway marks the container unhealthy → restart" | **Does not happen.** Railway's healthcheck is deploy-time only |
| 5 | "All chat histories wiped" | **False.** Transcripts live in HealthClaw, per tenant |
| 6 | "Rate-limit state wiped" | **Half true.** The in-process *burst* limiter resets (by design); the durable daily cap and the login-code attempt counter both survive |

## Does `/healthz` starve? Yes

Shipping config, 8 threads as `2 × 4`, each turn holding for 20s:

| N | turns started | `/healthz` 5s budget | 30s budget | true wait |
|---|---|---|---|---|
| 0 | — | 4/4 ok | 1/1 ok | 0.01s |
| 4 | 4/4 | 4/4 ok | 1/1 ok | 0.00s |
| 6 | 6/6 | 4/4 ok | 1/1 ok | 0.00s |
| 8 | 7/8 | 3/4 ok | 1/1 ok | **18.19s** |
| 10 | 8/10 | 1/4 ok | 1/1 ok | **18.13s** |
| 16 | 8/16 | 0/4 ok | 0/1 ok | **38.32s** |

`/healthz` never refuses a connection — it connects instantly and waits for a
free thread, so the wait tracks the *turn duration*, not N. At the code's own
worst case (`run_deadline_seconds` = 120):

| N | hold | `/healthz` true wait |
|---|---|---|
| 8 | 120s | **118.12s** |

**N=6 is a coin flip.** The single sample above is clean at N=6; repeating it is
what shows the effect. Two gunicorn workers accept from a shared socket, so six
turns land unevenly and saturate one worker while the other idles. Across three
attempts of six runs each — bursty arrivals and arrivals staggered 0.3s apart —
**2, 4 and 4 of 6 stalled ~18s**, the rest ~0.01s. On six trials that spread is
ordinary noise, so read it as "intermittent", not as a rate. What is stable is
the shape: the wait is either ~0s or ~18s and never in between, because it
depends on which worker accepted the probe. Staggering arrivals does not remove
it. Full numbers in the evidence doc.

## Why there is no restart loop

<https://docs.railway.com/guides/healthchecks>:

> The healthcheck endpoint is currently **not used for continuous monitoring**
> as it is only called at the start of the deployment.

The Dockerfile's `HEALTHCHECK` (`--interval=30s --timeout=5s --retries=3`,
i.e. unhealthy after 90s of saturation) is not evaluated by Railway, and Docker
does not restart an unhealthy container by itself. The VPS alternative is
systemd with `Restart=on-failure` and no health check — it restarts on process
*exit*, which saturation does not cause.

**No mechanism, no restart, no wipe, no loop.** What saturation actually costs
on Railway is a *deploy* that fails its health check and does not go live.

## Why the data-loss half is mostly false

Measured by driving the real endpoints, restarting gunicorn against the same
database, and reading the stores back:

| State | Survives a restart? |
|---|---|
| Chat transcripts (HealthClaw `cc_conversation_messages`, per tenant) | **Yes** — not in CareAgents at all |
| Login-code attempts (`EmailToken.attempts`) | **Yes** — read back as 2 |
| Durable daily turn cap (`UsageDay`) | **Yes** — counted 4 across the restart |
| Chat *burst* limiter (in-process deque) | **No** — 429 before, 200 after |

Only the last is lost, and `careagents/config.py` already documents that as
deliberate, backed by the DB-backed daily cap.

## Mitigations, measured one at a time

| Mitigation | Measured effect |
|---|---|
| **Bound the turn** | Turn 20s → 5s dropped the wait **18.19s → 0.01s** at N=8 and **38.32s → 13.73s** at N=16. But the bound already exists (`run_deadline_seconds` = 120); the lever is lowering it, which trades a hung page for a cut-off answer |
| **Read timeout 25s → 10s** | **Nothing while HealthClaw is healthy** — those calls return in milliseconds. It binds only when HealthClaw *hangs*, and then it is the thread hold (measured 23.03s at the 25s default). The 10s comparison sample is not clean and is not quoted as one — see the evidence doc. Does not touch the case #219 describes |
| **Raise threads (2×8)** | **Fixes N=8 outright** (18.19s → **0.00s**). At N=16 it halves the wait (38.32s → **18.23s**) and starts all 16 turns, but `/healthz` still blows the 5s budget 3 times in 4. Buys about one doubling of headroom; the failure moves, it does not go away |

## Two findings not in the issue

1. **`/healthz` fails its own 5s probe budget at zero load.** It calls
   `hc.agent_worker_health()` with the client's 25s timeout, so a slow
   HealthClaw makes `/healthz` take 25s (measured: 25.01s → 503) with no chat
   traffic at all. Triggered by a HealthClaw outage, not by concurrency —
   the opposite of #219's causal story. Worth its own issue.
2. **Each open turn polls HealthClaw 4×/s** — ~31 req/s at N=8, aimed at the
   HealthClaw that also serves the clinician. Flagged, not chased.

Also: `_stream_run` has no `try/except` around `agent_run_events`, so a
`HealthClawError` mid-stream kills the generator and the browser sees the
connection drop with **no `error` event**. Small and separate.

## What is faked, and why

Real: `careagents.wsgi:app` under the Dockerfile's own gunicorn invocation, the
real `HealthClawClient` with its real 25s timeout, real sessions, real rate
limiters, real account store.

Faked: **HealthClaw only**, as a loopback `ThreadingHTTPServer` that
`HEALTHCLAW_BASE` points at — so the client and its timeouts run as shipped,
with nothing monkeypatched inside gunicorn. It holds a run in `running` for a
controlled number of seconds, and that duration is the independent variable.
The LLM is not faked because the web tier never calls it.

`CARE_CHAT_TURNS` / `CARE_CHAT_TURNS_PER_DAY` are raised in the rig so a
rate-limited turn (429, instant, holds no thread) cannot masquerade as a held
thread; every scenario reports how many turns actually got their `accepted`
event.

## What could NOT be measured locally

1. Which `CARE_WEB_WORKERS` Railway sets for CareAgents (both 2×4 and 1×8 were
   measured to cover it).
2. The CareAgents Railway service's own healthcheck settings — the repo's
   `railway.toml` is HealthClaw's; CareAgents deploys from a staged directory.
3. Real HealthClaw latency under the poll storm (the fake answers instantly).
4. The real distribution of turn durations in production — the hold was
   controlled, not observed. That number decides how often N=6 is even reached.
5. The account store is SQLite in the rig, Postgres in production. That
   inflates the "turns started" column (serialised writes at turn start); it
   does not touch the starvation numbers, since the SSE loop never reads it.
6. Railway's edge behaviour on a 150s SSE stream.
7. Whether any of this has happened. This measures a mechanism, not an
   incident rate.

## Deliverables

- `scripts/probe_219_thread_saturation.py` — the rig; a non-author can re-run
  it from `main` with `uv run python scripts/probe_219_thread_saturation.py`.
- `docs/evidence/2026-09-03-probe-219-thread-saturation.md` — full transcript,
  tables, and the recommendation.

Gates:

```
$ uv run ruff check .
All checks passed!

3157 passed, 13 skipped, 1 xfailed, 2 warnings in 112.44s (0:01:52)
```

**Note:** `dependency-audit` is red on every open PR from npm advisories fixed
in unmerged #544, not by anything here.

**Do not merge** — this is a measurement for the council to rule on. Auto-merge
is not enabled.

### DIFF
--- a/docs/evidence/2026-09-03-probe-219-thread-saturation.md
+++ b/docs/evidence/2026-09-03-probe-219-thread-saturation.md
@@ -1,0 +1,763 @@
+# Probe — what actually happens to `/healthz` under concurrent chat turns (#219)
+
+- **Date:** 2026-09-03
+- **Branch / HEAD:** `probe/219-thread-saturation` off `origin/main` @ `89b42fb`
+- **Script:** `scripts/probe_219_thread_saturation.py` (committed; re-runnable
+  from `main` by a non-author with
+  `uv run python scripts/probe_219_thread_saturation.py`)
+- **Production was not touched.** No load was sent to `careagents.cloud` or
+  `app.healthclaw.io`. Every number below comes from a local gunicorn plus a
+  fake HealthClaw on loopback.
+- **One redaction:** the operator's home-directory name in pasted shell output
+  reads `<user>`. Nothing else in any transcript is altered.
+
+## Verdict: **REPRODUCES DIFFERENTLY**
+
+The starvation half is real and worse than stated. The restart half does not
+happen, and most of the data-loss half is already false.
+
+| # | Claim in the issue | Measured |
+|---|---|---|
+| 1 | "One worker, 8 threads" | **Wrong.** `--workers 2 --threads 4`. Still 8 threads, but the split makes saturation start *earlier* and *non-deterministically* |
+| 2 | "Each chat turn holds a thread ... 6 tool rounds × 25s + 90s LLM" | **Wrong mechanism, right symptom.** The web tier runs no inference since #257. The hold is the SSE replay loop, bounded by the 120s run deadline |
+| 3 | "8 concurrent turns block everything including `/healthz`" | **Confirmed at 8, intermittent from 6.** At the code's own worst case `/healthz` waits **118s** |
+| 4 | "Railway marks the container unhealthy → restart" | **Does not happen.** Railway's healthcheck is deploy-time only; it does not monitor a running container |
+| 5 | "All chat histories wiped" | **False.** Transcripts live in HealthClaw, per tenant |
+| 6 | "Rate-limit state wiped" | **Half true.** The in-process *burst* limiter resets (by design); the durable daily cap and the login-code attempt counter both survive |
+| — | *(not in the issue)* | **`/healthz` breaks its own 5s probe budget at zero load**, because it makes a 25s-timeout call to HealthClaw |
+| — | *(not in the issue)* | **Each open chat turn polls HealthClaw 4×/s.** 8 turns ≈ **31 req/s** aimed at the HealthClaw that serves the clinician |
+
+---
+
+## 1. What the code actually does (the premise moved under the issue)
+
+The issue is quoted from the 2026-08-01 architecture audit
+(`docs/2026-08-01-alignment-review.md:44`). Since then #257 moved inference out
+of the web process, which changes the mechanism the issue describes.
+
+A later review already caught this. `docs/2026-08-05-pattern-first-architecture-review.md:161`
+describes the same component as "**8 request slots held by 150s SSE turns**" —
+the correct mechanism, four days after the audit #219 quotes. The issue was
+never updated to match, and has been read since as if the audit's version were
+current. This probe confirms the 08-05 reading and measures it.
+
+`careagents/wsgi.py:3`:
+
+> The WSGI process only authenticates, enqueues, and replays durable run
+> events. Inference and tools execute in ``python -m careagents.worker``.
+
+So `/api/chat` (`careagents/app.py:873`) claims the inbound message, creates a
+durable run, and returns an SSE stream. The gunicorn thread is held by
+`_stream_run` (`careagents/app.py:833`), which polls HealthClaw for run events
+every `run_sse_poll_seconds` until the run is terminal.
+
+The numbers the issue multiplies together are the *worker's* budget, not the
+web tier's:
+
+| Setting | Value | Where |
+|---|---|---|
+| `MAX_TOOL_ROUNDS` | 6 | `careagents/agent.py:24` |
+| HealthClaw client read timeout | 25.0s | `careagents/healthclaw.py:54` |
+| `run_deadline_seconds` | 120 | `careagents/config.py` |
+| `run_sse_timeout_seconds` | 150 | `careagents/config.py` |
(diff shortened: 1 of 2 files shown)


aks129:
**Force-pushed once, and why.** The first push failed `secret-scan`. It was a
real finding, not noise: gitleaks `generic-api-key` on
`scripts/probe_219_thread_saturation.py:64`, which was
`MINT_SECRET = "probe-219-internal-secret"` — a throwaway value the rig hands
to its own loopback fake, but a credential-shaped literal in a committed file
all the same, and the repo's own rule is not to commit one.

Fixed by generating all three rig values (`SESSION_SECRET`, `MINT_SECRET`, and
the fake's step-up stub) with `secrets.token_urlsafe` at start-up. Both
processes already received them through the environment, so nothing about the
measurement changes — the committed file simply no longer contains a literal.

Amended rather than appended, because gitleaks scans the commit: a follow-up
commit would have left the string in the branch's history and the gate red. The
branch is unmerged and mine, and `--force-with-lease` was used. Auto-merge was
not, and is not, enabled.

No measurement was re-run for this change, because it touches no measured path
— but the rig was re-executed after it (`--only hold`) to confirm it still
boots and holds threads identically.

Issue #570, opened 2026-09-03
decide whether questionnaire_populate stays in the model-facing tier now that the read is bounded (D10 bullet 5)


aks129:
Council ruling D10's fifth bullet: "Pull `questionnaire_populate` from the model-facing tier for the beta, unless the demo kit depends on it." PR #562 implemented the other four bullets and left this one, and the review of that PR correctly flagged that it was left without a linked deferral. This is the deferral.

## Where it stands

The tool is still in the model-facing read tier: `adapters/tools.manifest.json:198` and `services/agent-orchestrator/src/tools.ts:423`. There is no "from your records" marker on populated answers.

Nothing in the demo scripts, the examples, or the end-to-end suite calls it, so the ruling's escape clause does not apply.

## Why it was not simply done

The reason the council gave for pulling it was that this path returned unredacted patient data and upstream free text. PR #562 removes that: expressions now see an allowlisted projection, auto-loaded clinical content is redacted before the engine sees it, and deleted rows are not read. The premise the ruling reasoned from is gone.

Cutting a tool from the model-facing tier is a user-visible capability change, so it is not a call to make inside a code review on the strength of "the reason no longer applies".

## What to decide

1. Does the tool stay, now that the read is bounded? If it stays, D10 asks that populated answers carry a "from your records" marker so a person can tell what came from their record and what an agent inferred. That marker does not exist yet.
2. If it goes for the beta, what does a caller get instead, and what says so?
3. Either way, the trade recorded in PR #562 needs to be visible to whoever answers: an auto-loaded value now populates only if the server can label its code.

## Related

PR #562, council ruling 2026-09-02 D10, and the allergen labelling question raised by the same review.

aks129:
Gap found by a parallel session, checked against all five diffs in the stack and against every other open change: **neither half of D10's fifth bullet is done.**

The ruling said pull `questionnaire_populate` from the model-facing tier, or — if it stays — have answers carry a "from your records" marker. PR #562's body correctly deferred the *removal* to this issue, because cutting a capability is a user-visible change. The stated alternative was never built. There is no `meta.tag`, no marker of any kind in the response the engine constructs.

So the tool is live in the model-facing tier today with neither the removal nor the marker.

That is my omission rather than the ruling's ambiguity. Deferring one half of an either-or and not implementing the other leaves the position the ruling explicitly ruled out.

[...]

PR #576, opened 2026-09-03
candidate: report every unpopulated leaf, delete the probe (CTO ruling on #562)

aks129:
## What

Implements ruling 1 and ruling 2 of the CTO ruling on #562.

- **The issue is emitted unconditionally, and the probe is deleted.**
  `resolves_outside_projection`, `_PROBE_PATIENT`, `_PROBE_RESOURCES` and the
  `lru_cache` are gone — 124 lines out of `r6/sdc/expressions.py`.
  `patient_projection` and `build_context` are untouched.
- **One mechanism for both paths.** `_report_unpopulated` is reached from the
  expression path, the `item.code` Observation path, and list-group leaves
  carrying a `definition`, so allergies, conditions and medication doses report
  too.
- **Retext.** One sentence that is true whether the value was withheld or
  simply absent, claiming nothing about the patient, handing over the public
  allowlist. The old constant was named `WITHHELD_BY_PROJECTION` and said so;
  it was true of only some of the leaves it was attached to.
- **Severity `warning` → `information`** (`r6/sdc/routes.py`). The `code` stays
  `incomplete`.
- **Row count equals record count**, pinned. Three unlabelled SNOMED allergies
  produce three repeats and six issues, not fewer rows.
- **The exclusion**, pinned as its own named test:
  `allergies.no-known-allergies` and `medications.no-current-medications` never
  appear in the issue list, on any patient, with any content.

## Why

The probe closed QA's leak but was the wrong control, and the ruling's own live
numbers said so: six of twelve withheld expressions produced no issue at all,
because the classifier fired only when the caller's filter happened to match the
placeholder text. `where($this.startsWith('1'))` — the ordinary way an author
writes a filter, and the exact shape QA used to walk an SSN out — came back
silent, and a caller reads silence as "this patient has no such value". That is
`docs/2026-08-02-retro.md` pattern 1: a control that looks like "we tell you
when we refused" and does "we tell you when your expression matches our
placeholder".

The zero-bits argument is what makes unconditional emission safe rather than
merely safer. `_populate_item` attaches `answer` only when a value resolved, so
**which linkIds have no answer is already in the response**. An issue per
unanswered attempted leaf is a redundant encoding of what the caller is already
holding, and therefore carries no information about the patient *by
construction* — not "we checked it doesn't leak", but "there is nothing to
leak", which nobody has to re-verify the next time the file is touched. It is
also complete, which the probe was not: every withheld path names itself, not
just the ones an inventory of placeholders happened to cover.

## Base note

**This PR stacks on #562** (`fix/sdc-populate-bounded-reads`) and targets that
branch, not `main`. It is the "required before merge" work from the ruling, not
a replacement for #562 — the projection, the redaction, the `is_deleted`
filters and the 7d→24h cut all live in #562 and are unchanged here.

## The live matrix

Against a running app on **port 5311** (5099, 5199 and 5297 were reported taken
by other agents today), synthetic seeded tenant: 2 SNOMED allergies, 1 SNOMED
condition, 1 RxNorm medication, 1 LOINC observation, an SSN-shaped identifier.
All twelve expressions in one request. `probe` is what the deleted classifier
reported, measured by the CTO on `:5199`.

```
issue  answer  probe  expression
------------------------------------------------------------------------------
True   False   True   %patient.identifier.value
True   False   True   %patient.photo.url
True   False   True   %resources.where(resourceType='Observation').code.text
True   False   True   %patient.maritalStatus.text
True   False   True   %patient.telecom.where(system='sms').value
True   False   True   %patient.identifier.first().value
True   False   False  %patient.identifier.value.where($this.startsWith('1'))
True   False   False  %patient.identifier.value.where($this.startsWith('9'))
False  True    False  %patient.identifier.value.combine(%patient.gender)
True   False   False  %patient.contact.gender
True   False   False  %patient.deceasedDateTime
True   False   False  %patient.name.period.start

severity: ['information']   code: ['incomplete']   issues: 11/12
```

**One correction to the ruling's matrix.** Row 9 is not a silence.
`combine` unions a withheld path with an allowlisted one, so on a patient who
has a gender it **resolves — to the gender**, and the item is *answered*. It
reported no issue under the probe and reports none now, correctly, and the
biconditional holds for it. Eleven of the twelve were silences; that one was
not. It is held out of the invariance test for that reason, with the reasoning
written down at `tests/test_populate_issue_property.py`.

Issue text, verbatim:

```
e00: not populated: no value resolved. $populate reads only the %patient
projection (name, birthDate, gender, telecom phone/email, address
line/city/state/postalCode) and coded content this server recognises; anything
else is not read.
```

No expression text is echoed into the OperationOutcome (checked against all
twelve).

**The intake form**, same seeded patient — the addendum's predicted **six**:

```
answered leaves (10): 9 demographics + medications.item.name
issues (6): allergies.item.allergen x2, allergies.item.reaction x2,
            conditions.item.name, medications.item.dose
row counts: allergen 2, condition 1, medication 1   (stored: 2, 1, 1)
no-known-allergies: answered=False   in-issue-list=False
```

**The noise measurement**, both halves:

```
nine intake expressions, full demographics -> 9/9 answered, issues []
postalCode removed                         -> issues ['postal']
restored                                   -> issues []
```

**The SSN walk over HTTP**, 10 guesses in one request:

```
guesses named: 10/10 — all of them, so no guess is distinguishable
RECOVERED: ''
```

**The severity change does not hide the issues from the model.** Checked
because it would have been the quiet way to undo the retext:
`populateQuestionnaire` in `services/agent-orchestrator/src/tools.ts:1487-1510`
returns the whole `Parameters` body verbatim and filters nothing on
`severity` — the `information` issues reach the model exactly as `warning`
ones did. (The `severity`-filtering code at `tools.ts:1317-1319` belongs to
`validate_resource`, a different tool.) Neither the tool description at
`tools.ts:423` nor `adapters/tools.manifest.json:198` claims anything about
withheld data, so no stale copy needed correcting there.

**Audit and PHI sweep:** 7 `$populate` audit rows, details
`populate; issues=0|1|6|10|11`, no PHI marker in any of them; none of the six
withheld markers (identifier, maritalStatus, sms telecom, contact name, photo
URL, `name.text`) appears anywhere in a `$populate` response.

## Ran

```
uv run ruff check .
  All checks passed!

  3187 passed, 13 skipped, 1 xfailed, 2 warnings in 102.44s
  (baseline on da16068: 3180 passed. +8 property tests, +2 list tests,
   -3 resolves_outside_projection tests.)

    -q -p no:cacheprovider
  50 passed          (Grade A green; no ratchet pin was moved)
```

**Mutation-tested — four breaks, all red before restore** (retro rule 1; the
point is that the property test is load-bearing rather than decoration):

| mutation | red |
| --- | --- |
| M1 report only when the expression resolves against the unbounded record (the original oracle) | 18 |
| M2 report only for expressions without a `where(` (the probe's silencing shape) | 17 |
| M3 omit a list repeat whose leaves all resolved nothing | 3 |
| M4 report leaves with no population mechanism (the attestations) | 13 |

Restored, 35 passed. M4 was re-run on its own to confirm by name that
`test_the_attestation_items_never_appear_in_the_issue_list` is one of the
reds — the exclusion pin the addendum asked for is load-bearing, not
incidentally green.

The property test is `answer present ⟺ no issue for that linkId`, per item,
over three patients × the ruling's twelve expressions. Stated as **counts**, so
a repeating group with one labelled and one unlabelled row reads correctly, and
the "was population attempted" predicate is derived from the **Questionnaire**
rather than from the engine — so it stays honest if `populate.py`'s own
predicate changes. The three deleted `resolves_outside_projection` tests are
replaced by it and it catches strictly more: the signature pin could not see a
probe-driven classifier, and M2 is exactly that.

`dependency-audit` is red on this PR. It is red on **every** open PR right now —
npm advisories fixed in unmerged **#544**, nothing from this branch.

## What was NOT run

- **Postgres CI lane.** SQLite only here. Nothing in this change touches a
  query or an id.
- **`$extract` against the changed response.** The ruling establishes that
  nothing new enters the answer set, which is what would round-trip, but I did
  not exercise `$extract` on a form populated by this code.
- **Whether any renderer surfaces the issues.** The CTO flagged this as
  unverified and as a real-patient gate rather than a merge gate; it still is.
  A blank allergen row that is *reported* in the response and still *renders*
  blank on a patient's phone is not fixed where it matters
  (`r6/sdc/documents.py`, the intake form UI).
- **The MCP `questionnaire_populate` caller** against the new issue count and
  the `information` severity.
- **Load.** Emission is now one list append per unanswered leaf; the probe's
  ~800× amplification is gone with it, but I did not re-measure.

## Deliberately not touched — the ruling's other three required items

These are in the ruling and **not** in this PR, so they do not get lost:

3. **`r6/sdc/routes.py:205`** — the unredacted subject append. Still there.
   Nothing reads it, and after this change there is no `%resources` consumer at
   all, so it is dead weight rather than a live door — but it is the ruling's
   item 3 and wants its own diff.
4. **D10 bullet 5** — `questionnaire_populate` in the model-facing read tier
   (`adapters/tools.manifest.json:198`,
   `services/agent-orchestrator/src/tools.ts:423`). No "from your records"
   marker was added. **The written deferral is #570**, which records the
   decision to be made and why the reason the council gave for pulling the
   tool no longer holds. An earlier version of this line said no deferral
   was written; that was wrong, and the ruling's requirement is met by the
   link. Note this change *strengthens* the case for leaving the tool in:
   the text a model now sees claims nothing about the patient.
5. **The PR-body correction on #562** about the scope of the labelling trade.

Also unfiled, from the ruling's "not blocking, please file" list: `$extract`'s
duplicate-Patient and wrong-element write; the SNOMED label additions needing
clinician initials; `r6/sdc/expressions.py` logging caller-supplied expression
text at WARNING (now line 165 — log injection, newlines included, pre-existing);
`_coerce` returning dicts.

### DIFF
--- a/r6/sdc/expressions.py
+++ b/r6/sdc/expressions.py
@@ -30,14 +30,23 @@ own structure.
 
 AND NOTHING IN THIS MODULE MAY EVALUATE A CALLER'S EXPRESSION AGAINST THE
 REAL RECORD. Not to answer a yes/no question, not to decide whether to warn.
-`resolves_outside_projection` reads a constant probe for exactly that reason
-— its docstring has the reproduction. A bool derived from a patient's data
-and handed to the caller is a read of that data; repeat it once per
-questionnaire item and it is a fast one.
+A bool derived from a patient's data and handed to the caller is a read of
+that data; repeat it once per questionnaire item and it is a fast one. That
+is not a hypothetical: a `resolves_outside_projection` classifier lived here
+and decided whether to report an unpopulated item by evaluating the caller's
+own expression against the unbounded record, and eleven HTTP requests walked
+a stored identifier out through the issue list one character at a time (QA
+review of PR #562).
+
+There is no classifier here now, in any shape. `r6/sdc/populate.py` reports
+every leaf that resolved no value, which depends on nothing but whether an
+answer was produced — a fact the caller already holds, because the response
+carries `answer` only on the leaves that got one. Nothing to leak, by
+construction, and nobody has to re-verify that the next time this file is
+touched.
 """
 
 import logging
-from functools import lru_cache
 
 import fhirpathpy
 
@@ -140,139 +149,6 @@ def build_context(subject=None):
     if projection is None:
         return {}
     return {'patient': projection, 'subject': projection}
-
-
-#: The record `resolves_outside_projection` probes. It is a CONSTANT, and it
-#: has to stay one — see that function's docstring. Every element the
-#: projection withholds is populated here with a placeholder, so an
-#: expression reaching for one resolves against this and not against a
-#: patient. Nothing here is anybody's data.
-_PROBE_PATIENT = {
-    'resourceType': 'Patient',
-    'id': 'projection-probe',
-    'name': [{'given': ['Probe'], 'family': 'Probe', 'text': 'Probe Probe',
-              'prefix': ['Probe'], 'suffix': ['Probe'], 'use': 'official'}],
-    'birthDate': '2000-01-01',
-    'gender': 'unknown',
-    'deceasedBoolean': False,
-    'multipleBirthInteger': 1,
-    'active': True,
-    'telecom': [{'system': s, 'value': 'probe', 'use': 'home'}
-                for s in ('phone', 'email', 'sms', 'fax', 'pager', 'url',
-                          'other')],
-    'address': [{'line': ['probe'], 'city': 'probe', 'state': 'probe',
-                 'postalCode': 'probe', 'district': 'probe',
-                 'country': 'probe', 'text': 'probe', 'use': 'home'}],
-    'identifier': [{'system': 'urn:probe', 'value': 'probe', 'use': 'usual',
-                    'type': {'text': 'probe'}}],
-    'photo': [{'url': 'probe', 'contentType': 'image/png', 'title': 'probe',
-               'data': 'cHJvYmU='}],
-    'contact': [{'name': {'given': ['Probe'], 'family': 'Probe'},
-                 'telecom': [{'system': 'phone', 'value': 'probe'}],
-                 'address': {'line': ['probe'], 'city': 'probe'},
-                 'relationship': [{'text': 'probe'}]}],
-    'communication': [{'language': {'text': 'probe'}, 'preferred': True}],
-    'maritalStatus': {'text': 'probe', 'coding': [{'code': 'probe'}]},
-    'generalPractitioner': [{'reference': 'Practitioner/probe',
-                             'display': 'probe'}],
-    'managingOrganization': {'reference': 'Organization/probe',
-                             'display': 'probe'},
-    'link': [{'other': {'reference': 'Patient/probe'}, 'type': 'seealso'}],
-    'extension': [{'url': 'urn:probe', 'valueString': 'probe'}],
-    'modifierExtension': [{'url': 'urn:probe', 'valueString': 'probe'}],
-    'meta': {'tag': [{'code': 'probe'}], 'security': [{'code': 'probe'}],
-             'profile': ['urn:probe'], 'source': 'probe'},
-    'text': {'status': 'generated', 'div': 'probe'},
-    'implicitRules': 'urn:probe',
-    'language': 'en',
-}
-
-#: One placeholder per resource type $populate auto-loads, so an expression
-#: reaching for `%resources` is reported rather than silently empty. Constant,
-#: for the same reason as _PROBE_PATIENT.
-_PROBE_RESOURCES = [
-    {'resourceType': 'Observation', 'id': 'probe', 'status': 'final',
-     'subject': {'reference': 'Patient/probe', 'display': 'probe'},
-     'code': {'text': 'probe', 'coding': [{'code': 'probe',
-                                           'display': 'probe'}]},
-     'valueString': 'probe',
-     'valueCodeableConcept': {'text': 'probe'},
-     'note': [{'text': 'probe'}],
-     'performer': [{'reference': 'Practitioner/probe', 'display': 'probe'}]},
-    {'resourceType': 'MedicationRequest', 'id': 'probe', 'status': 'active',
-     'subject': {'reference': 'Patient/probe', 'display': 'probe'},
-     'medicationCodeableConcept': {'text': 'probe'},
-     'dosageInstruction': [{'text': 'probe'}]},
-    {'resourceType': 'AllergyIntolerance', 'id': 'probe',
-     'patient': {'reference': 'Patient/probe', 'display': 'probe'},
-     'code': {'text': 'probe'},
-     'reaction': [{'manifestation': [{'text': 'probe'}]}]},
-    {'resourceType': 'Condition', 'id': 'probe',
-     'subject': {'reference': 'Patient/probe', 'display': 'probe'},
-     'code': {'text': 'probe'}},
-]
-
-
-@lru_cache(maxsize=512)
-def resolves_outside_projection(expression):
-    """Does `expression` reach for something the projection does not carry?
-
-    THE ONE PROPERTY: **this is a pure function of the expression text.** It
-    never sees a patient. It evaluates `expression` twice against the
-    CONSTANT `_PROBE_PATIENT` — once through the projection and once around
-    it — and reports "outside" when the unbounded form resolves and the
-    bounded form does not.
-
-    It reads a constant because the obvious implementation does not, and the
-    obvious implementation is a PHI exfiltration channel. Evaluating the
-    caller's expression against the REAL record and returning whether it was
-    non-empty makes the issue list a one-bit oracle over exactly the data the
-    projection withholds: a caller sends
-
-        %patient.identifier.value.where($this.startsWith('123'))
-
-    on one item, reads whether an issue named that linkId, and walks the
-    withheld SSN out one character at a time — measured at eleven HTTP
-    requests, with the value never once appearing in an answer (QA review of
-    PR #562, council ruling D10). A bool is a value. Any function of the
-    record whose output reaches the caller is a read of the record, however
-    narrow the output looks.
-
-    Answering from a constant keeps what the ruling asked for — an item that
-    was refused says so, naming its linkId, instead of looking like a patient
-    with no phone number — while telling the caller nothing whatsoever about
-    this patient. `%patient.identifier` is outside the projection for
-    everyone; that is a property of the allowlist, not of a record.
-
-    The bounded half of the comparison is what keeps an ALLOWLISTED path that
-    the real patient simply lacks (`%patient.telecom.where(system='email')`
-    on a patient with no email) from being reported as withheld: it resolves
-    against the probe on both sides, so it is not "outside".
-
-    WHAT THIS IS NOT: exhaustive. A withheld path `_PROBE_PATIENT` happens
-    not to populate — `%patient.deceasedDateTime`, `%patient.name.period`,
-    `%patient.telecom.rank` today — answers False, so that item comes back
-    silently empty instead of naming itself. That is a gap in the REPORTING,
-    never in the bound: the projection still withholds it, and nothing about
-    the patient reaches the caller either way. Widen the probe to close one;
-    do not widen it with anything but placeholders.
-
-    Cached because the answer depends only on `expression`, which also caps
-    the work a caller can buy per request.
-    """
-    if not expression:
-        return False
-    projected = patient_projection(_PROBE_PATIENT)
-    bounded = {'patient': projected, 'subject': projected}
-    unbounded = {
-        'patient': _PROBE_PATIENT,
-        'subject': _PROBE_PATIENT,
-        # The name the projection withholds entirely.
-        'resources': _PROBE_RESOURCES,
-    }
-    if evaluate(expression, _PROBE_PATIENT, unbounded) is None:
-        return False
-    return evaluate(expression, projected, bounded) is None
 
 
 def evaluate(expression, resource, context=None):
--- a/r6/sdc/populate.py
+++ b/r6/sdc/populate.py
@@ -27,16 +27,29 @@ test_zero_allergies_never_infers_no_known_allergies (the load-bearing one)
 and test_zero_medications_never_infers_no_current_medications.
 """
 
-from r6.sdc.expressions import (build_context, evaluate,
-                                resolves_outside_projection)
+from r6.sdc.expressions import build_context, evaluate
 
-#: What an item's issue says when the %patient projection is the reason it has
-#: no answer. Fixed text: a Questionnaire author controls the expression, and
-#: echoing it back would put an author-supplied literal (`where(value='...')`)
-#: into an OperationOutcome that leaves the boundary.
-WITHHELD_BY_PROJECTION = (
-    'not populated: the expression resolves outside the %patient projection '
-    'this operation permits (name, birthDate, gender, telecom, address)'
+#: What an attempted-but-unresolved leaf reports. ONE sentence for both
+#: causes, because the server does not tell the two apart and must not appear
+#: to: "the projection withheld it" and "the patient has no email address"
+#: read identically here on purpose. `questionnaire_populate` is in the
+#: model-facing read tier, and text naming a refusal would have a model tell
+#: a patient their record was held back when it is simply empty.
+#:
+#: What the caller gets instead is the allowlist, which is public — they can
+#: compare it against their own expression without the server answering a
+#: question about a patient to do it. (The constant this replaced was called
+#: WITHHELD_BY_PROJECTION and said so; it was true for only some of the
+#: leaves it was attached to, which is retro pattern 1 living in a name.)
+#:
+#: The expression is deliberately NOT echoed: a Questionnaire author controls
+#: it, and a `where(value='...')` clause is a place to park a literal that
+#: would then ride out on an OperationOutcome.
+NOT_POPULATED = (
+    'not populated: no value resolved. $populate reads only the %patient '
+    'projection (name, birthDate, gender, telecom phone/email, address '
+    'line/city/state/postalCode) and coded content this server recognises; '
+    'anything else is not read.'
 )
 
 INITIAL_EXPRESSION_URL = (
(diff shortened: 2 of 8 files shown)


aks129:
## QA: FAIL on #576 — two MEDIUMs, neither a leak. The zero-bits claim survived every attack I could build.

**This is not a GitHub approval and must not be read as one.** `autoMergeRequest` was
`null` on #576 and #578 when I started, and `null` after my push. Merge order is
**#562 → #576 → #578**. `dependency-audit` is red here and on every open PR: npm
advisories fixed in unmerged **#544**, nothing from this branch.

---

### The zero-bits argument holds. I could not falsify it.

I did not accept the argument; I tried to break `answer present ⟺ no issue for that
linkId`, per item, and failed.

- **Fuzz, in-process.** 300 random patients (each of 12 optional elements present or
  absent) × 37 expressions, including `.exists()`, `.empty()`, `.count()`, `.length()`,
  `iif(%patient.identifier.exists(),'a','b')`, `%patient.identifier.value.length() > 5`,
  `|` and `combine()` unions of a withheld path with an allowlisted one,
  `where($this.startsWith('1'))`, `descendants()`, `select($this)`, `%resources`,
  `%context` / `%resource` / `%rootResource`, an undefined `%var`, a parse error, and
  constants. **0 falsifications.**
- **The invariance half.** On a questionnaire of 23 withheld/erroring expressions over
  200 random patients the issue list took **two** signatures, and the single row that
  moves is `%patient.identifier.value | %patient.gender` — which *answers*, from the
  allowlisted half. It varies exactly as far as the answer does. Every genuinely withheld
  path names itself on every patient.
- The type/length shapes do not leak either, and the reason is the projection rather than
  the reporting: `%patient.identifier.exists()` is `false` and `.count()` is `0` for
  *every* patient, so they answer, constantly.
- **Live, over HTTP** (seeded synthetic tenant, port 5433): the eleven-guess SSN walk
  named 10/10 guesses on the first request. `RECOVERED: ''`.
- No issue ever names a linkId absent from the response — I checked that explicitly, since
  an issue for a leaf the caller cannot see would be a bit they do not already hold. A
  group is dropped only when its whole subtree contains no leaf, and such a subtree emits
  no issues.

**Row count.** 0/1/3/7 allergies × {unlabelled SNOMED, `code: {}`, labelled, no `code`
key} — 16 shapes, rows == records in all 16. Live intake: 2 allergies / 1 condition /
1 medication stored → 2 / 1 / 1 rows.

**The exclusions.** Never named, on any patient, with any content — but the exclusion is
**by mechanism, not by name**, and the PR is right to do it that way. A renamed
`nka-renamed`, a nested `outer.nka`, and an NKA-shaped boolean *inside* a repeating
allergy group are all silent because none carries an expression, a code, or a
`definition`. The same boolean *with* a `definition` is reported — correctly, since a
definition is an attempted population.

**Live intake, the addendum's predicted six:**

```
answered leaves (10): 9 demographics + medications.item.name
issues (6): allergies.item.allergen x2, allergies.item.reaction x2,
            conditions.item.name, medications.item.dose
severity: ['information']   code: ['incomplete']
rows: allergies 2, conditions 1, medications 1   (stored 2/1/1)
attestations: unanswered, and named in 0 issues
6 issues vs 8 unanswered leaves — the 2 are the attestations
PHI markers in the response: none of 6
audit: 18 populate rows, details 'populate; issues=N', 0 PHI markers
```

**Nothing downstream treats issues as failure** — because nothing reads them at all.
`grep` across `r6/`, `careagents/`, `services/`, `static/`, `templates/` finds exactly one
producer (`r6/sdc/routes.py:90`) and no consumer. `tools.ts:1487` returns the whole
`Parameters` body verbatim, `severity`-blind. That also settles the CTO's open question
the other way: `r6/actions/review.py:250` discards the return with `qr, _issues = ...`, so
**the action rail cannot surface a blank allergen row** — the issue channel is dropped at
the call site. Real-patient gate, as the addendum said, now measured rather than assumed.

---

### MEDIUM — "every attempted leaf reports" has a hole, and the pin cannot see it

`r6/sdc/populate.py:229-232` (the `_list_group_resource_type` gate) and
`r6/sdc/populate.py:372-382` (the "NOTHING WAS ATTEMPTED" branch).

The addendum's predicate is "an `initialExpression`, an `item.code` Observation match, or
**a list-group leaf with a `definition`**". A `definition` leaf inside a `repeats: true`
group satisfies that wording whatever type it names. The engine reads it narrower:
`_list_group_resource_type` recognises only MedicationRequest / AllergyIntolerance /
Condition, so a repeating group naming anything else is not a list group, falls through to
ordinary-group recursion, and its leaves reach `_resolve_answer` with no expression and no
code — nothing attempted, nothing reported.

Live, port 5433, seeded tenant:

```
POST /r6/fhir/Questionnaire/$populate
  repeats:true group, leaves definition=...#Immunization.vaccineCode.text
                                        ...#Immunization.occurrenceDateTime
-> 200
   {"item":[{"item":[{"linkId":"imm.vaccine"},{"linkId":"imm.date"}],"linkId":"imm"}]}
   issues parameter present: False
```

Two unanswered leaves, **zero issues, no `issues` parameter at all.** That is the state
the ruling set out to remove — "a caller reads silence and concludes the patient has no
such value" — and it makes the PR body's "complete: every withheld path names itself" not
true. No live impact: the intake form uses only the three supported types. Not a leak: the
silence is a function of questionnaire structure, identical on every patient.

**The pin is the other half.** `tests/test_populate_issue_property.py`'s
`_attempted_link_ids` encoded the *ruling's* reading, so its own predicate disagreed with
the engine — and the file passed only because no fixture exercised the difference, which
is exactly the "implementation-independent" claim its docstring makes. Reproduce:

```
uv run python - <<'PY'
import sys; sys.path[:0] = ['.', 'tests']
from r6.sdc.populate import populate_questionnaire
from test_populate_issue_property import _assert_zero_bits
q={"resourceType":"Questionnaire","status":"active","item":[{"linkId":"imm",
   "type":"group","repeats":True,"item":[{"linkId":"imm.v","type":"string",
   "definition":"http://x#Immunization.vaccineCode.text"}]}]}
p={"resourceType":"Patient","id":"p1"}
qr,iss=populate_questionnaire(q,p,[])
_assert_zero_bits(q,qr,iss,"unsupported")   # AssertionError, pre-49f3149
PY
```

**What I did about it, and what I did not.** I fixed the test only — I write no production
code. `49f3149` makes the helper read "attempted" the way the engine does, states the
three supported types as a literal (the repo's two-file convention, cf.
`_UNREDACTED_EXITS`) so a fourth type goes red until the pin is updated, adds the shape to
the biconditional matrix, and pins today's silence in
`test_a_repeating_group_of_an_unsupported_type_is_silent` with the divergence written out.
Plain assertion, not a strict xfail — CLAUDE.md records a strict-xfail row going red the
day someone fixed what it pinned.

**The call is the CTO's, not mine:** widen the engine to report any `definition` leaf
under a repeating group (the mutation is three lines and I confirmed it reddens exactly
the two tests that should go red), or narrow the ruling's wording. Either way both places
move together. **This is why I am marking FAIL** — not because the behaviour is dangerous,
but because it is retro pattern 1 in a PR whose whole argument is that it removed one, and
the check that should catch it could not.

### MEDIUM — the issue list is a ~15× worsening of an existing response-size amplification

`r6/sdc/populate.py:48-53` (`NOT_POPULATED`, 230 chars) × `r6/sdc/routes.py:342-345`.

The caller supplies both halves: a `content` Bundle (passed through unredacted by design)
and an inline Questionnaire. K definition leaves in one repeating group × R records =
K·R issues, each carrying the same 230-character sentence.

```
K x R        request     QR      issues   total/req   pre-#576
 10 x   10     2.7KB    2.4KB    29.9KB       12x        0.9x
 50 x  200    28.3KB  208.5KB     2.9MB      113x        7.4x
200 x  500    81.1KB    2.1MB    29.3MB      396x         26x
500 x 2000   282.8KB   20.8MB   293.5MB     1138x         75x
```

Live over HTTP, same shape: 35.0KB request → **3157.4KB response**, status 200, 0.02s.

`MAX_CONTENT_LENGTH` is 16 MB (`main.py:379`), which does not bound this — the issue list
is built whole in memory before serialisation. Arithmetic, not a measurement I ran: the
same shape at the 16 MB cap is ~4.5e9 cells and the process would not survive it. The
underlying K·R shape is pre-existing; #576 multiplies the bytes per cell by ~14, and the
prior review's DoS finding on this exact path is why I am not filing it as a LOW.

I have deliberately **not** pinned a number — choosing a cap is your call. One fix shape
that preserves the biconditional exactly: the sentence is a constant, so emit it once as a
summary issue and give the per-linkId issues a short diagnostic.

### LOW / noted, no action asked

- The linkId is echoed verbatim into `diagnostics` (`r6/sdc/routes.py:343`), and now on
  every unanswered leaf rather than only on probe hits. Zero new bits — the QR already
  echoes every linkId — but the model-facing tier receives caller-authored text far more
  often than before. Pre-existing shape.
- `r6/actions/review.py:208` still does `content = [patient]` with an unredacted Patient,
  the append #578 deletes on the `$populate` route; and that path calls no
  `apply_redaction` at all, so D10's bound is a property of the `$populate` route rather
  than of the engine. Pre-existing, outside both PRs, #578's body already names it —
  worth filing.
- The six identical 230-char sentences on an ordinary seeded patient (~25 on a
  MEDENT-shaped import) all land in a model's context verbatim. Product observation.

### Required-before-merge items 4 and 5 are still open

Both PR bodies name them honestly, which is the opposite of the silent omission the ruling
was guarding against — but the ruling asked for the marker **or a written deferral**, and
#576's body says "no deferral is written". Under the ruling as written the stack is not
merge-ready until item 4 is closed one way or the other; item 5 is a #562 body edit.

### What I ran

```
  3193 passed, 13 skipped, 1 xfailed        (3189 before my two test additions)
uv run ruff check .
  All checks passed!
  50 passed                                  (Grade A green, no pin moved)
```

**Test honesty.** `git checkout origin/fix/sdc-populate-bounded-reads -- r6/sdc/` →
**12 failed, 38 passed**, plus 2 collection errors (`test_populate_issue_property.py` and
`test_sdc_populate.py` cannot import `NOT_POPULATED`). The new tests are load-bearing. One
exception worth naming: `test_the_attestation_items_never_appear_in_the_issue_list` passes
on the old code too — it is a regression pin, not a change-proving test, and its teeth come
from your M4 mutation.

**My own additions, mutation-tested** (restored green at 56):

| mutation | red |
| --- | --- |
| M1 `item.code` branch stops reporting | 2 |
| M2 `item.code` reports only when no Observation was stored (a record-driven classifier) | 2, **both new — the shipped suite is green on M2** |
| M3 unsupported-type definition leaves report (the shape of the fix) | 2 |

M2 is the one that matters. The `item.code` path had no coverage anywhere, because the
intake form has no `item.code` items, so a classifier re-introduced on that branch alone
would have passed the entire suite. It is now covered over eight record shapes, including
the two that only appear after `apply_redaction` runs — a `valueCodeableConcept` stripped
of its `text` whose code has no label, and a `value[x]` type the resolver does not read.
Both resolve to nothing on a record that *exists*, and all five unanswerable shapes
produce one byte-identical issue payload.

### What I did NOT cover

- **`$extract` against a form populated by this code.** Nothing new enters the answer set,
  which is what would round-trip, but I did not exercise it — and the duplicate-Patient
  write the ruling describes is still unfiled.
- **Any renderer.** I established that nothing reads the issues; I did not look at the PDF
  or the intake UI, and the action rail provably cannot show them.
- The Postgres CI lane (SQLite only), the MCP `questionnaire_populate` caller against the
  new issue count, and any browser/mobile rendering.
- Concurrency: the amplification numbers are single-request.

---

**Verdict: FAIL**, on the unsupported-type silence plus the amplification. Neither is a PHI
leak, neither is an oracle, and the projection, the redaction and the unconditional-issue
design are all correct and I want them. It flips to PASS when the CTO rules on the
`definition`-leaf reading (either direction, both places updated together) and either
accepts or bounds the response-size multiplier.

Do not arm auto-merge.

aks129:
### QA addendum — a second silence with the same root cause, and two wording corrections

Still not an approval; `autoMergeRequest` still `null` on both PRs.

**A nested group inside a repeating list group loses its leaves entirely.**
`_populate_list_group` (`r6/sdc/populate.py:313-329`) iterates the group's **direct**
children and tests `child.get("definition")`. A child that is itself a group has none, so
it is emitted as a bare item and its own leaves are never visited:

```
group allergies.item (repeats, AllergyIntolerance)
  leaf  a.allergen            definition=...#AllergyIntolerance.code.text
  group a.reactions
    leaf a.reactions.text     definition=...#AllergyIntolerance.reaction.manifestation.text

-> [{"linkId":"allergies.item","item":[
      {"linkId":"a.allergen","answer":[{"valueString":"Penicillin"}]},
      {"linkId":"a.reactions"}]}]
   issues: []
   'a.reactions.text' anywhere in the response: False
```

The stored `reaction.manifestation.text` is "Hives" and the leaf that asks for it does not
appear at all — no answer, no issue, no item. Same family as the MEDIUM above (the rule
for which list-group leaves count as attempted), same non-impact today (the intake form
has no nesting inside `allergies.item`), and again not a leak: it is a function of
questionnaire structure only. Worth folding into whichever way the CTO rules on that
predicate, rather than a separate finding.

The property test passes here because neither `_attempted_link_ids` nor
`_leaf_occurrences` walks into the sub-group either — so this is a third shape the pin
cannot see, which is the same point as the MEDIUM.

**Two corrections to my comment above, both wording:**

1. In the amplification table, only the **`50 x 200` row was also measured live over HTTP**
   (35.0KB → 3157.4KB, 0.02s). The `200 x 500` and `500 x 2000` rows are in-process
   measurements of the engine plus `_issues_outcome`, not HTTP round trips. I said so once
   in the prose; the table should have carried it.
2. "the shipped suite is green on M2" is an inference, not something I ran in isolation.
   What I observed: under M2 exactly two tests reddened and **both are ones I added**;
   nothing shipped reddened. Same fact, stated as the measurement.

**One coverage note against the brief I was given**, since it asked specifically about
result-type dependence: `iif()`, `.exists()`, `.empty()` and `.count()` return *constants*
against the projection — `false`, `true`, `0` — because the withheld element is absent
rather than present-with-a-type, so they answer identically for every patient. On the
`item.code` path an unsupported `value[x]` (I used `valueBoolean`) resolves to nothing and
is byte-for-byte indistinguishable from "no Observation stored". Neither carries a bit.

PR #578, opened 2026-09-03
candidate: delete the unredacted subject append and stop logging caller expressions (CTO items on #562)

aks129:
## What

Two cleanups, both from the CTO ruling on #562. Neither changes behaviour, and
the live comparison below is the evidence for that rather than the claim.

- **The unredacted subject append is gone** (`r6/sdc/routes.py`, required item
  3). `_gather_content` appended the stored Patient — unredacted — to the
  content list handed to the populate engine. The subject still reaches the
  engine whole, through the `subject` argument, which is where reference
  matching and the `%patient` projection both read it from.
- **`evaluate()` no longer logs the caller's expression** (`r6/sdc/expressions.py`,
  the ruling's "not blocking, please file" list). The WARNING now names the
  item's linkId and the exception's class.
- One comment and two docstrings corrected where they described the deleted
  behaviour. The `_gather_content` comment explaining why the subject is *not*
  redacted stays — that is still true of where the subject travels — and the
  half describing the append is gone.

### Nothing read the append — here is the proof, four reads

`content_resources` is read in exactly two places, and neither can see a
Patient:

| read | filter | Patient |
| --- | --- | --- |
| `populate.py:200` `observations = [...]` | `resourceType == "Observation"` | excluded |
| `populate.py:306` `_populate_list_group` | `resourceType == resource_type`, over the three `_LIST_RESOURCE_CONFIG` keys (MedicationRequest / AllergyIntolerance / Condition) | excluded |

The other two paths the CTO named do not reach the list at all:

- **Code matching.** `_observation_answer` (`populate.py:410`) is handed
  `observations`, already filtered above — it never sees `content_resources`.
- **The reference-matching helper.** `_references_subject`
  (`populate.py:333`) only ever runs on resources that survived the list-group
  filter, and the thing it matches *against* is `subject_ref = _reference(subject)`
  — derived from the `subject` argument, not from the list. Removing the append
  cannot change what it compares.
- **Expressions.** `_resolve_answer` evaluates against `context["patient"]`, the
  projection. `build_context` puts nothing else in the environment, which
  `tests/test_sdc_expressions.py::test_the_environment_holds_the_projection_and_nothing_else`
  asserts as an exact set — there is no `%resources` for an expression to walk
  the list with.

Dynamic half: full suite green, and the live matrix below byte-identical.

It was still worth deleting rather than leaving. The append was the second door
to the identifier oracle in QA's review — `%resources.where(resourceType='Patient').identifier.value`
distinguished a right guess from a wrong one — and it stays reachable by
anything that walks `content`, so its safety is a property of today's consumers
rather than of the code. The sentence that would put it back was in
`populate_questionnaire`'s own parameter docstring: *"should include the subject
and any Observations"*. That is corrected here, as part of the same change; it
is the contract being changed, not scope creep.

### The expression text

A Questionnaire is request body, so the expression is the caller's: it can carry
newlines that forge a log line, and literals a caller chose to park in a
`where()`. `str(exc)` is out for the same reason — fhirpathpy puts the offending
token in the message (`Not implemented: notAFunction`), which is the same text
arriving by the back door. What is logged is the linkId and the exception class.

Nothing is bounded or sanitised beyond that, and one honest note: **the linkId is
caller-supplied too.** It is questionnaire structure rather than patient data,
which is the CTO's reason for accepting it, and `%r` is what escapes control
characters in it — that is the whole mitigation, and it is the same `%r` the old
line used. No new code.

Read off the two server logs, and stated exactly as measured — the same matrix
requests went to both, the forged request went only to the "after" server:

```
before (branch point, same request):
  r6.sdc.expressions - WARNING - FHIRPath evaluation failed for
  "%resources.where(resourceType='Observation').code.text": ValueError

after (same request):
  r6.sdc.expressions - WARNING - FHIRPath evaluation failed for item 'e02':
  ValueError

after, POSTed Questionnaire whose expression is
"notAFunction()\n2026-09-03 18:00:00 - r6.access - WARNING - step-up token
 accepted for tenant admin":
  r6.sdc.expressions - WARNING - FHIRPath evaluation failed for item
  'demographics.phone': Exception
  (grep for `notAFunction` and `step-up token accepted` in the log: no hits)
```

The "before" side is #576's head, not `main`; the forged shape was not sent to
it, so the first line is the echo demonstrated live and the forged line is an
"after"-only measurement.

## Base note

**This PR stacks on #576, which stacks on #562.** Merge order: **#562 → #576 →
this**. It targets `fix/populate-issue-unconditionally` (#576), not `main`. The
projection, the redaction, the `is_deleted` filters and the 7d→24h cut are
#562's; the unconditional issue and the probe deletion are #576's; neither is
touched here.

## Before and after, live

Against a running app on **port 5417** (5099, 5199, 5297 and 5311 were reported
taken today), synthetic seeded tenant matching #576's: 2 SNOMED allergies, 1
SNOMED condition, 1 RxNorm medication, 1 LOINC observation, an SSN-shaped
identifier, and `maritalStatus` / `contact` / `photo` / an `sms` telecom /
`name.text` / `name.period.start` for the withheld paths.

Method: one script, run against the branch point (byte-identical to #576's head)
and again against this commit, on the same seeded database. **`diff` of the two
outputs is empty.**

<details><summary>The run — condensed for reading, not the raw capture</summary>

Section 1 keeps its table verbatim; sections 2–5 are summarised (the answered
leaves are listed by group rather than one per line, and the parentheticals are
mine, not the script's). The load-bearing fact is the one above it: the two raw
captures `diff` to nothing.

```
=== 1. ruling matrix (status 200) ===
issue  answer  expression
True   False   %patient.identifier.value
True   False   %patient.photo.url
True   False   %resources.where(resourceType='Observation').code.text
True   False   %patient.maritalStatus.text
True   False   %patient.telecom.where(system='sms').value
True   False   %patient.identifier.first().value
True   False   %patient.identifier.value.where($this.startsWith('1'))
True   False   %patient.identifier.value.where($this.startsWith('9'))
False  True    %patient.identifier.value.combine(%patient.gender)
True   False   %patient.contact.gender
True   False   %patient.deceasedDateTime
True   False   %patient.name.period.start
severity: ['information']
code: ['incomplete']
issues: 11/12
issue text e00: e00: not populated: no value resolved. $populate reads only
the %patient projection (name, birthDate, gender, telecom phone/email,
address line/city/state/postalCode) and coded content this server
recognises; anything else is not read.      [wrapped for width]

=== 2. escape sweep (status 200) ===
answer  expression
True    %patient.descendants()
True    %patient.children()
False   %patient.repeat(item)
True    %patient.select($this)
False   %patient.ofType(Patient)
True    %patient.toString()
False   Patient.identifier.value
False   %resources
False   %context.identifier.value
False   %resource.identifier.value
False   %rootResource.identifier.value
False   %patient.extension.valueString
False   %patient.contact.telecom.value
False   %patient.name.text
False   %patient.address.country
False   %patient.photo.contentType
markers present in response: []
   (the four structural walkers that answer resolve to the PROJECTION; none
    of the six markers — SSN, maritalStatus, sms telecom, contact name,
    photo URL, name.text — appears anywhere in the response)

=== 3. intake form (status 200) ===
answered leaves (10): 9 demographics + medications.item.name
issues (6): allergies.item.allergen x2, allergies.item.reaction x2,
            conditions.item.name, medications.item.dose
row counts: allergen 2, condition 1, medication 1     (stored: 2, 1, 1)
no-known-allergies:     answered=False in-issue-list=False
no-current-medications: answered=False in-issue-list=False

=== 4. SSN walk (status 200) ===
guesses named: 10/10 — all of them, so no guess is distinguishable
RECOVERED: ''

=== 5. noise (status 200) ===
nine demographics -> answered 9, issues []
plus one absent element -> issues ['absent']
```

</details>

Every number matches #576's live matrix, including the `combine(...)` row that
correctly answers rather than being a silence.

## Tests

Both were written first and confirmed red against the branch point.

- `tests/test_sdc_populate_bounded.py::test_the_content_list_handed_to_the_engine_carries_no_subject`
  — spies on `populate_questionnaire` over the real HTTP route and asserts the
  **list**, not the response: no entry is a Patient, none carries the subject's
  id, and the withheld identifier is not in it. Asserted where it can be seen,
  because a leak through the append is invisible in the response — which is
  exactly why it survived this long. Paired with positives on the same request
  (the Observation *is* there and *is* redacted; the subject still arrives whole
  as the argument; the allowlisted items still populate) so an empty list cannot
  pass it. Re-adding `content.append(subject)` → red on the first assertion,
  measured.
- `tests/test_sdc_expressions.py::test_a_failing_expression_is_not_echoed_into_the_log`
  — drives a crafted expression through `populate_questionnaire` and asserts the
  **LogRecord**, not `caplog.text`. That distinction is load-bearing: `%r`
  escapes the newline when the record is formatted, so a text assertion passes
  while the raw expression still travels in `record.args` to every handler —
  and this app installs a `JSONFormatter` in production (`main.py`). It also
  pins `exc_type.isidentifier()`, which is what separates the class name from
  `str(exc)`, and the linkId, which is what makes the log useful. (This app's
  own `JSONFormatter`, `main.py:36`, uses `getMessage()` and would not have
  leaked the raw string — the second commit corrects a docstring that said
  otherwise. The record is still where the untrusted text is.)
  Mutation: drop `link_id=link_id` in `populate.py` → red
  (`assert None == 'demographics.phone'`); restored, green.

## Ran

```
uv run ruff check .
  All checks passed!

  3189 passed, 13 skipped, 1 xfailed, 2 warnings in 112.55s
  (baseline on #576's head: 3187 passed. +2, the two tests above.)

    -q -p no:cacheprovider
  50 passed          (Grade A green; no ratchet pin moved — nothing here
                      changes an R6Resource.query site)
```

Plus, against the running app on `:5417`: the five sections above, twice, and a
POSTed Questionnaire carrying the forged log line, with the server log read back
both times.

The second commit is a docstring correction only. `ruff` and
`tests/test_sdc_expressions.py tests/test_sdc_populate_bounded.py` re-run on it
(`All checks passed!` / `24 passed in 0.95s`); the full suite and the live run
were not repeated, because no assertion and no production line changed.

`dependency-audit` will be red here. It is red on **every** open PR right now —
npm advisories fixed in unmerged **#544**, nothing from this branch.

## What was NOT run

- **Postgres CI lane.** SQLite only. Nothing here touches a query or an id.
- **`$extract`.** Unchanged by this diff, and nothing new enters the answer set,
  but not exercised.
- **`test_evaluate_returns_none_on_bad_expression`.** Noticed while crafting the
  new test: `"this is not fhirpath ((("` returns `[]` from fhirpathpy without
  raising, so that test has been exercising the empty-result path, not the
  `except` branch it reads as covering. The new test is currently the only one
  that reaches the log line at all. Left alone — it is not in the ruling and
  it does still pin something true.
- **The other `_gather_content`.** `r6/actions/review.py:208` does the same
  thing (`content = [patient]`) on the action rail's form-fill path. The same
  four-read proof says it is dead weight there too, but it is a different route
  and not in the ruling — **noticed, deliberately not touched.** Worth its own
  issue.
- **Load.** Deleting a list entry can only shorten two comprehensions; not
  measured.
- **Whether any renderer surfaces the issues** — still the real-patient gate the
  CTO flagged on #576, unchanged here.
- **The MCP `questionnaire_populate` caller.** Unchanged response shape, not
  re-exercised.

## Still open from the ruling, not in this PR

Required items 4 (D10 bullet 5 — the "from your records" marker or a written
deferral) and 5 (the PR-body correction on #562's labelling trade) are
**not** here. Neither is a code change to these files, and #576's body already
carries them forward; naming them again so they do not get lost between three
stacked PRs.

### DIFF
--- a/r6/sdc/expressions.py
+++ b/r6/sdc/expressions.py
@@ -151,19 +151,32 @@ def build_context(subject=None):
     return {'patient': projection, 'subject': projection}
 
 
-def evaluate(expression, resource, context=None):
+def evaluate(expression, resource, context=None, link_id=None):
     """Evaluate a FHIRPath expression, returning a scalar, list, or None.
 
     Returns the single value when the result has one element, the list when
     it has several, and None when empty or on any evaluation error.
+
+    `link_id` names the questionnaire item, for the failure log only — see
+    below for why it is what gets logged and the expression is not.
     """
     if not expression:
         return None
     try:
         result = fhirpathpy.evaluate(resource or {}, expression, context or {})
     except Exception as exc:  # noqa: BLE001 — never let one expr kill the run
-        logger.warning('FHIRPath evaluation failed for %r: %s',
-                       expression, type(exc).__name__)
+        # THE EXPRESSION IS NOT LOGGED. A Questionnaire is request body, so
+        # its expression text is the caller's: newlines that forge a log
+        # line, and literals the caller chose to park in a `where()`. Nor is
+        # `str(exc)` — fhirpathpy puts the offending token in the message
+        # ("Not implemented: notAFunction"), which is the same text arriving
+        # by the back door. The linkId identifies the item well enough to
+        # debug and is questionnaire structure rather than patient data; it
+        # is caller-supplied too, and %r is what escapes control characters
+        # in it. Pinned by tests/test_sdc_expressions.py::
+        # test_a_failing_expression_is_not_echoed_into_the_log.
+        logger.warning('FHIRPath evaluation failed for item %r: %s',
+                       link_id, type(exc).__name__)
         return None
     if not result:
         return None
--- a/r6/sdc/populate.py
+++ b/r6/sdc/populate.py
@@ -182,8 +182,11 @@ def populate_questionnaire(questionnaire, subject, content_resources):
 
     questionnaire: Questionnaire dict.
     subject: Patient dict (or None).
-    content_resources: list of resource dicts available for population
-        (should include the subject and any Observations).
+    content_resources: list of resource dicts available for population —
+        Observations, and the MedicationRequest / AllergyIntolerance /
+        Condition rows the list groups match. NOT the subject: that arrives
+        as `subject` above, and putting it here as well left an unredacted
+        Patient reachable by anything walking this list (PR #562 review).
     issues: list of {'linkId', 'detail'} for items that errored (not for
         items that simply had no data).
     """
@@ -347,7 +350,8 @@ def _resolve_answer(item, item_type, context, observations, issues, link_id):
         # The resource root is `context["patient"]` — the bounded projection,
         # not the stored Patient — so the allowlist applies to the
         # resource-root form (`Patient.name.family`) as well as to `%patient`.
-        value = evaluate(expr, context.get("patient"), context)
+        value = evaluate(expr, context.get("patient"), context,
+                         link_id=link_id)
         if value is not None:
             return _coerce(value, item_type), value_key
         _report_unpopulated(issues, link_id)
--- a/r6/sdc/routes.py
+++ b/r6/sdc/routes.py
@@ -195,14 +195,23 @@ def register_sdc_routes(blueprint, deps):
         return None
 
     def _gather_content(params, subject, tenant_id):
+        # THE SUBJECT IS NOT IN THIS LIST. It reaches the engine as
+        # populate_questionnaire's `subject` argument, and it is NOT redacted
+        # there: an intake form exists to carry the patient's own name, DOB
+        # and address, and apply_redaction would truncate all three. What
+        # bounds it is the %patient projection in r6/sdc/expressions.py — an
+        # allowlist of exactly those elements — plus reference matching,
+        # which reads only the id.
+        #
+        # It used to be appended here as well. Nothing read it: the engine
+        # filters this list to Observations and to the three list-resource
+        # types, so a Patient passed no filter. But it was the second door to
+        # the identifier oracle QA walked an SSN through (PR #562 review) —
+        # `%resources.where(resourceType='Patient').identifier.value` — and
+        # it stayed reachable by anything that walks content. Pinned by
+        # tests/test_sdc_populate_bounded.py::
+        # test_the_content_list_handed_to_the_engine_carries_no_subject.
         content = []
-        if subject:
-            # The subject is NOT redacted: an intake form exists to carry the
-            # patient's own name, DOB and address, and apply_redaction would
-            # truncate all three. What bounds the subject is the %patient
-            # projection in r6/sdc/expressions.py, which is an allowlist of
-            # exactly those elements.
-            content.append(subject)
         if subject and subject.get("id"):
             for resource_type, subject_field in _AUTO_LOADED_RESOURCE_TYPES:
                 content.extend(_redacted_for_populate(
--- a/tests/test_populate_issue_property.py
+++ b/tests/test_populate_issue_property.py
@@ -172,7 +172,46 @@ def _expression_questionnaire(expressions):
 # The property, and the two halves it is built from
 # ---------------------------------------------------------------------------
 
-def _attempted_link_ids(questionnaire, parent_repeats=False):
+#: The resource types list-group population actually implements. This is a
+#: LITERAL, deliberately, not an import of r6/sdc/populate.py's
+#: `_LIST_RESOURCE_CONFIG` — the repo's two-file convention (cf.
+#: `_UNREDACTED_EXITS` and tests/test_unredacted_exits.py). Adding a fourth
+#: type to the engine goes red here until someone updates this set, which is
+#: the point of writing it twice.
+#:
+#: It is here because "a list-group leaf with a `definition`" turned out to
+#: mean two different things. The engine reads it as "a leaf of a group whose
+#: leaves name a type in that table" — `_list_group_resource_type` returns
+#: None for anything else, the group falls through to ordinary-group
+#: recursion, and its leaves are then not attempted at all. The first version
+#: of this helper read it as "any `definition` leaf under any repeating
+#: group", which is the CTO addendum's literal wording. The two disagree on
+#: exactly one shape, pinned below by
+#: test_a_repeating_group_of_an_unsupported_type_is_silent, and no fixture
+#: here exercised it, so the disagreement was invisible. QA review of #576.
+LIST_POPULATED_TYPES = {"MedicationRequest", "AllergyIntolerance", "Condition"}
+
+
+def _list_group_type(item):
+    """The resource type this repeating group's leaves populate from, or None.
+
+    Mirrors r6/sdc/populate.py:_list_group_resource_type — the FIRST child
+    definition naming a supported type wins, and a group naming none is not a
+    list group at all.
+    """
+    if not item.get("repeats"):
+        return None
+    for child in item.get("item", []):
+        definition = child.get("definition") or ""
+        if "#" not in definition:
+            continue
+        rtype = definition.split("#", 1)[1].split(".", 1)[0]
+        if rtype in LIST_POPULATED_TYPES:
+            return rtype
+    return None
+
+
+def _attempted_link_ids(questionnaire):
     """linkIds where the questionnaire asks populate to resolve something.
 
     Read from the Questionnaire, not from populate.py, so this test still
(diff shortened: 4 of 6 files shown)


aks129:
## QA: PASS on #578 — the four-read proof re-derived, and the "identical output" claim is a SHA match

**This is not a GitHub approval and must not be read as one.** `autoMergeRequest` was
`null` on #576 and #578 when I started, and `null` after my push. Stack order is
**#562 → #576 → #578**; this PR's verdict is conditional on its base — I marked **#576
FAIL** for two MEDIUMs (neither a leak), written up there. `dependency-audit` is red here
and on every open PR: npm advisories fixed in unmerged **#544**, nothing from this branch.

---

### 1. The subject append — I re-derived the proof rather than checking your arithmetic

`content_resources` is read in exactly two places, and the other two paths the CTO named
never touch the list:

| site | filter | can a Patient survive it? |
| --- | --- | --- |
| `populate.py:200-201` `observations = [...]` | `resourceType == "Observation"` | no |
| `populate.py:305-310` `_populate_list_group` | `resourceType == resource_type`, over the three `_LIST_RESOURCE_CONFIG` keys | no |
| `_observation_answer` (`populate.py:410`) | handed `observations`, already filtered | never sees the list |
| `_references_subject` (`populate.py:333`) | runs only on rows that survived the type filter; compares against `_reference(subject)`, derived from the **argument** | cannot be reached, and could not change if it were |

`build_context` is the fifth path and takes no content at all. The environment is exactly
`{patient, subject}` — no `%resources` for an expression to walk, and
`test_the_environment_holds_the_projection_and_nothing_else` pins it as an exact set.
Your table is right.

**Dynamic half, at the engine.** Same questionnaire of 37 adversarial expressions,
`content=[]` vs `content=[markerPatient]` vs `content=[markerPatient]*5`: identical every
time — 25 issues, 12 answers, and zero of six planted markers (`name.text`, an SSN-shaped
identifier, `photo.url`, `maritalStatus.text`, a contact name, an `sms` telecom) anywhere
in the output.

**Dynamic half, at the route.** I did the before/after as a byte comparison rather than
taking it on the PR body's word: same seeded SQLite database, same running app on the same
port, `git checkout e5b2d3c -- r6/` for "before" and `HEAD` for "after", four request
shapes — the real intake form, a 32-expression escape/withheld matrix, a request whose
inline `content` Bundle carries a marker-bearing **Patient**, and a no-subject request.

```
7143ac8d0005cf4a3415e03d4abf87d3cd75ac9a98c3d76ffb743ff2493d83ed  before.json
7143ac8d0005cf4a3415e03d4abf87d3cd75ac9a98c3d76ffb743ff2493d83ed  after.json
diff exit 0, 35402 bytes each
```

Byte-identical. The claim is true, and the third shape is the one that matters: a caller
can still put a Patient into `content` through the inline Bundle, and it is still read by
nothing.

### 2. The log change — checked through `record.args`, not just the formatted message

Driven through `populate_questionnaire` with a crafted expression
`notAFunction()\n2026-09-03 18:00:00 - r6.access - WARNING - step-up token accepted for
tenant admin`, over seven linkId shapes (plain, newline-forge, ANSI escapes, dict, list,
int, `None`), inspecting the `LogRecord` itself:

- **No caller expression text reaches any handler.** `notAFunction` and `step-up token
  accepted` appear in neither `getMessage()` nor `record.args` — only
  `type(exc).__name__`. That was the whole point of the change and it holds.
- **`%r` does escape the linkId.** A linkId containing a real newline renders as `\n` in
  `getMessage()`, and the standard `basicConfig` line contains no embedded newline, so a
  forged second log line is not producible through the linkId either.
- **`record.args[0]` does carry the raw linkId**, newline unescaped — as your body says.
  This app's `JSONFormatter` (`main.py:33-45`) uses `getMessage()`, so nothing leaks here;
  an external handler that serialises `args` would see it. Disclosed accurately, and
  `6f3de46`'s correction of the docstring is the right call.
- **Can a caller put patient data in a linkId?** Yes — they author the Questionnaire. It
  is their own text, and the QuestionnaireResponse already echoes every linkId back, so
  the log adds no channel that did not exist. Noted on #576 that under #576 the linkId now
  also rides into `diagnostics` on every unanswered leaf rather than only on probe hits.

`test_a_failing_expression_is_not_echoed_into_the_log` asserts the `LogRecord` rather than
`caplog.text`, which is the distinction that makes it worth having. Confirmed red against
`e5b2d3c`.

### 3. Test honesty

Against the branch point (`git checkout origin/fix/sdc-populate-bounded-reads -- r6/sdc/`)
both of this PR's tests go red:

```
FAILED tests/test_sdc_expressions.py::test_a_failing_expression_is_not_echoed_into_the_log
FAILED tests/test_sdc_populate_bounded.py::test_the_content_list_handed_to_the_engine_carries_no_subject
```

`..._carries_no_subject` asserts the **list** rather than the response, which is the only
place a leak through the append would have been visible — that is why it survived this
long, and the test is written where the property lives.

### What I ran

```
  3193 passed, 13 skipped, 1 xfailed        (3189 before my two test additions on #576)
uv run ruff check .
  All checks passed!
  50 passed                                  (Grade A green, no pin moved)
```

Plus, against a running app on `:5433` with a synthetic seeded tenant (2 SNOMED allergies,
1 SNOMED condition, 1 RxNorm medication, 1 LOINC observation, an SSN-shaped identifier,
`maritalStatus` / `contact` / `photo` / `sms` / `name.text`): the intake form end to end
(10 answered, 6 issues, `information` / `incomplete`, rows 2/1/1, attestations silent,
zero markers), the twelve-expression matrix, the SSN walk, and the before/after capture
above on both commits.

### What I did NOT cover

- **`$extract`.** Unchanged by this diff and not exercised.
- **The Postgres CI lane** — SQLite only.
- **Any renderer.** Separately, I found that nothing anywhere reads the `issues` parameter:
  one producer (`r6/sdc/routes.py:90`), no consumer across `r6/`, `careagents/`,
  `services/`, `static/`, `templates/`.
- **The other `_gather_content`.** `r6/actions/review.py:208` still does
  `content = [patient]` with an unredacted Patient — and that path calls no
  `apply_redaction` at all, so D10's bound is a property of the `$populate` route rather
  than of the engine. Your body names it as deliberately not touched; I agree it is out of
  scope here and it should get an issue.

---

**Verdict: PASS on #578.** Both changes are correct, both are proved rather than argued,
and the "no behaviour change" claim is the strongest kind — a byte-for-byte comparison on
a live app, including the shape that would have exercised the removed append. It merges
behind #576, whose FAIL I have written up there.

Do not arm auto-merge.

PR #549, opened 2026-09-02
candidate: prod-watch watches careagents.cloud and the Telegram tile; the one flaky test no longer reaches the network (#537, #433)

aks129:
## What

- `scripts/prod_watch.py`: `CAREAGENTS` is now `https://careagents.cloud`, the origin users reach. The Railway hostname keeps a readiness check under its own name, `careagents (railway host): ready (db reachable)`, because `/healthz` there is what Railway's own health check hits and the one path that keeps answering once everything else 308s; its build marker rides along in the detail so a split deployment is visible.
- One new check, `careagents: telegram not advertised as live`. Fails on any `t.me/`/`tg://` link, or a Telegram call to action ("Open in Telegram", "Telegram — connect →") on any page an unauthenticated visitor can read: the landing page, plus `/home` only if it answers 200 without a session (fetched without following its redirect). A tile marked "coming soon" passes. A landing that could not be read is a failure, not a pass. The detail names which pages were read. No `getMe` probe — no bot token in this script's environment.
- `tests/test_prod_watch_build.py`: the autouse fixture fakes `prod_watch.post` beside `get` and tripwires `requests.get`/`requests.post` beneath both, so any test in the file that reaches the network fails loudly. Docstring corrected. New tests for #433, the host retarget, and the Telegram check — including one that runs the detector on the real `careagents/templates/home.html` and asserts agreement with the template's state (so the in-flight "coming soon" change does not turn it red on `main`).
- `tests/test_build_marker_stamp.py`: pinned totals move 12 → 14 and 11 → 13 (MOVED PIN comments, per the file's convention); the gap of exactly one is the property pinned, and it holds.

## Why

- **Council D2 item 4 (#537):** the monitor watched a hostname nobody visits and could not see Telegram. Twelve checks were green while the surface had been dead since June (#536). Also the fix #289 asked for (not auto-closed here; maintainer's call).
- **Council D13 (#433), unanimous "fix today":** the file said "No network" while `test_a_current_build_passes` made a real HTTPS POST to the public demo MCP server on every run, with a 1-second timeout. CI green depended on a production service. Delivery-process amendment 4 (D1): no test in this file reaches the network now, and the docstring says so truthfully.

## Measured, not assumed: the Telegram check is green on production today

The brief expected red. It is green, and the detail says why:

```
 OK  careagents: telegram not advertised as live — no live Telegram link or call to action on /; /home not scanned (302 without a session)
```

[...]

Issue #410, opened 2026-08-05
PR #409 relocates the lie: five CareAgents callers report an outage as 'not yours', 'unknown run', or 'not found'


aks129:
Found reviewing PR #409 (the #403 seam fix). **The seam fix is correct and should ship.** This is the matching caller pass it needs, and one of the five is a behaviour change introduced *by* #409 rather than a pre-existing wart.

## The one that changes behaviour

`careagents/app.py:889-898`:

```python
def _agent_owns_action(agent_id, action_id):
    ...
    try:
        hc.action_status(ctx["tenant"], action_id)
        return ctx["tenant"]
    except HealthClawError:
        return None
```

Callers render `"That form isn't yours."` with a 404 (`:900-914` GET review, `:916-920` POST submit).

Before #409, `action_status` did not wrap `requests.RequestException` (verified on `origin/main:careagents/healthclaw.py:227-233`) — a refused connection escaped raw and Flask answered **500**. After #409 it is a typed `HealthClawError`, so it lands in this catch and the patient is told **the form isn't theirs**.

A 500 is ugly but honest: something broke. "That form isn't yours" is a confident false statement about ownership, on the human-approval path — the guardrail this product sells. The fix moved the defect rather than removing it, in exactly the shape the fix exists to eliminate:

> A guardrail produced **nothing**, and the caller read that as **an answer**.

## The other four

Same class, pre-existing, and #409 widens how many transport conditions reach them:

| Site | Outage is reported as |
|---|---|
| `app.py:828` | `{"error": "unknown run"}` 404 |
| `app.py:849` | `{"status": "unknown"}` 404 |
| `app.py:1084` | `{"error": "unknown run"}` 404 |
| `app.py:716` | `return False` — an availability flag reads false |

"Unknown run" and "not found" are claims about what exists. During an incident none of them is true.

## The pattern is already right elsewhere in the same file

`:298`, `:527`, `:807`, `:1061` answer `503 "records service unavailable"`; `:878` answers `502 "labs unavailable"`; `:416` maps 4xx through and 5xx to 502. #409 itself added the correct shape at `:544-558` (`503 {"status": "unavailable"}`) and rendered it in `home.js`. So the house style exists — these five predate it or were missed.

## Fix

Distinguish "we asked and the answer is no" from "we could not ask" at each of the five. `HealthClawError` already carries `.status`: a 404 from the engine is a real not-found; a transport failure or a 5xx is not, and must surface as unavailable.

[...]

Issue #272, opened 2026-08-02
prod-watch: an unreachable /healthz is reported as a stale build, prescribing a redeploy for an outage


aks129:
Found by QA in the final review round of #258. **LOW**, pinned as a strict xfail (`test_an_unreachable_endpoint_is_not_reported_as_a_stale_build`) so it cannot be quietly lost.

`scripts/prod_watch.py:182` — when `/healthz` is unreachable, `get()` returns an exception name, `body` stays `{}`, and `deployed` falls back to `"unknown"`, which is indistinguishable from a genuinely unmarked build. The build check then asserts a verdict about a field it never read:

```
FAIL careagents: running the current build — deployed build unknown is not one of the 1
commit(s) this run accepts (tip 4f2a91c). CareAgents does not auto-deploy — redeploy per
RELEASING.md §4.
```

`build_info` records `asserted: true, ok: false`, so the stale-build alarm fires and the issue body tells whoever reads it at 03:00 to redeploy — when the actual problem is that CareAgents is **down**.

This is #258's own principle inverted: the branch exists to stop a monitor reporting something it did not verify, and here it reports a build verdict without having read a build.

**Why it is LOW rather than blocking:** the outage alarm fires in the same run, carrying the real failing checks in its body, and the stale issue closes on the next healthy run. It misdirects; it does not silence.

**Fix:** assert the build only when `/healthz` was actually read — otherwise `report()` it as informational, the same way a run with no `--expect-sha` does. The distinction already exists in the code; this path just does not use it.

Related: #258, #270.

PR #580, opened 2026-09-03
candidate: one doubling of thread headroom, and a health check that answers for itself (#219)

aks129:
Candidate, not a merge request. Three changes, each sized by what the #219
probe measured (`docs/evidence/2026-09-03-probe-219-thread-saturation.md` on
`probe/219-thread-saturation`, PR #573) rather than by the issue's estimate.
Nothing was run against production.

## What

1. **`deploy/careagents/Dockerfile:67`** — gunicorn `--threads 4` → `--threads 8`.
2. **`careagents/app.py:1374` (`/healthz`)** — the worker-health call gets its
   own bounded budget, and the endpoint stops failing on a dependency it
   cannot reach. `careagents/healthclaw.py` gains a per-call `timeout` so it
   can.
3. **`careagents/app.py:851` (`_stream_run`)** — a mid-stream failure now emits
   a terminal `error` frame instead of dropping the connection silently.
4. **`scripts/prod_watch.py`** — a second named check that reads the run-worker
   state, because change 2 stopped the status code carrying it and the
   readiness check never read the field. Without this, change 2 would have
   traded a loud failure for a quiet one.

### 1. One doubling of thread headroom

Every open chat turn holds one web thread for the whole life of its run — the
SSE replay loop, not inference, which has run outside this process since #257.
The pool is therefore the concurrency ceiling for **every** route, `/healthz`
included. Measured against this exact gunicorn invocation, 20s turns
(evidence §3, §7):

| config | N=8 | N=16 |
| --- | --- | --- |
| 2 workers × 4 threads (shipping) | `/healthz` waited **18.19s** | **38.32s** |
| 2 workers × 8 threads | **0.00s** | **18.23s** |

At the code's own worst case — `run_deadline_seconds` = 120 — 8 concurrent
turns made `/healthz` wait **118.12s** on the shipping config (§4). The 2×4
split also starves *intermittently from 6*, because two pre-fork workers share
an accept socket and six turns land unevenly (§3).

**One doubling is all the evidence supports.** It moves the failure from 8
concurrent turns to 16; it does not remove it, and at N=16 the 5s budget is
still blown 3 times in 4. Each thread also polls HealthClaw ~4×/s on a
browser's behalf (§9), so more threads is more load on the engine that also
serves clinicians. The comment at the setting and the new pin test both say so.

**Can the platform override this?** Half of it, and this is the trap the repo
already hit with `PORT`:

- `--workers` **is** overridable — `${CARE_WEB_WORKERS:-2}`, a Railway
  dashboard variable. Its actual production value was not readable from this
  repo (evidence §10.1); the probe covered both 2×4 and 1×8 and they behave
  the same at 8 threads total.
- `--threads` is **not** env-settable. The Dockerfile value is authoritative
  unless a service overrides the whole start command.
- So a dashboard `CARE_WEB_WORKERS` other than 2 multiplies a thread count
  nobody measured. I did **not** add a `CARE_WEB_THREADS` flag — unrequested —
  but someone should read the CareAgents service's variables before deploying.

### 2. A health check that answers for itself

The probe measured `/healthz` taking **25.01s** and returning **503 at zero
load**, because its worker-health call inherited the client's 25s chat timeout
(§8). That blows every budget it has: the container probe allows 4s
(`careagents/healthcheck.py`), the Dockerfile's `HEALTHCHECK` 5s.

The call now takes a `(connect, read)` pair of 1s + 1s — 2.0s of network wait,
worst case — and the three worker states stop being collapsed into one bool:

| `run_workers_state` | `/healthz` | Why |
| --- | --- | --- |
| `ready` | 200 | Fresh worker presence. |
| `not_ready` | **503** | HealthClaw answered: nothing is draining the queue. |
| `unknown` | 200 | We could not ask inside the budget. |

**The deploy-gating question, ruled.** Railway probes only at the start of a
deploy, so a 503 here means the deploy does not go live and traffic stays on
the previous version. My ruling is that this is wanted for `not_ready` and
wrong for `unknown`:

- **`not_ready` keeps gating the deploy — deliberately.** HealthClaw answered
  and there is no worker. That is a CareAgents deployment which serves pages
  and cannot finish a chat turn — exactly the web-only deployment
  `docs/runbooks/careagents-durable-worker.md` documents a 503 for. The fault
  is in the thing being deployed, so blocking it is the right outcome, and
  "never relax the check" is honoured: that scenario produces the same 503 it
  always did, and a test pins it.
- **`unknown` no longer gates it.** Being unable to reach HealthClaw is a fact
  about HealthClaw, not about this container. Failing there blocks a CareAgents
  deploy on someone else's outage — including the deploy that would fix it —
  while traffic stays on a previous version with exactly the same dependency.
  It is also the mislabeling #410 named: "we could not ask" filed as "the
  workers are down".

This changes one previously-pinned behaviour, deliberately and in the open:
`test_unreachable_worker_control_plane_degrades_readiness_and_admission`
asserted 503 on the unreachable path. It is rewritten (and renamed) with the
reasoning above. **Admission is untouched** — `POST /api/chat` still refuses a
turn in both failure states, so nobody is routed into a turn no worker can run.
That is the gate that protects people; `/healthz` is the gate that gates
deploys.

`run_workers` stays a bool (two runbooks and a test read it as one); the new
`run_workers_state` field carries the claim the bool cannot make. The runbook
gained the table above.

### 2b. The monitor that would have gone quiet

Change 2 moved a signal, and `scripts/prod_watch.py` was reading the old one.
Its "careagents: ready" check asserts `status_code == 200 and accounts is True`
and never reads `run_workers` — free coverage while `/healthz` failed closed on
both worker failures, and a blind spot the moment one of them became a 200.

That blind spot is not hypothetical: it is the 2026-08-06 shape exactly. The
edge to HealthClaw was blocked *from CareAgents* while HealthClaw itself
answered prod_watch perfectly, so `healthclaw: alive` passed and the only red
line was CareAgents readiness. After change 2 alone, nothing would have been
red. So this PR adds the check rather than filing it:

```
 OK  careagents: a run worker is draining the queue — ready — a fresh worker is claiming runs
FAIL careagents: a run worker is draining the queue — not_ready — HealthClaw answered: nothing is draining the queue — the careagents-worker service is down or was never deployed
FAIL careagents: a run worker is draining the queue — unknown — CareAgents could not reach HealthClaw's worker health inside its budget — the engine or the network to it, not this app; chat is refusing turns
```

Both failures exit 1. It is a **second named check**, not another clause on the
readiness line, because that line means one thing — the app reached its own
database — and a control that quietly does two is this codebase's own defect
shape (`docs/2026-08-02-retro.md`). The detail names the state because the two
failures have different owners: `not_ready` sends an operator to the worker
service, `unknown` sends them to the engine. When `/healthz` is not readable at
all the check does not run — a verdict about a field the run never read is the
#272 defect, and the readiness check already reports that outage with the right
remedy.

Check-count pins in the build-marker tests move 12 → 13 and 11 → 12; the gap
between them, which is the honesty property those two tests exist for, is
unchanged.

> **Merge-order warning — this collides with PR #549.** Both PRs branch from
> `89b42fb`, where the pins are **12** and **11**, and both move them. #549
> adds two checks (careagents.cloud and the Telegram tile) and sets **14** and
> **13**; this PR adds one and sets **13** and **12**. Whichever merges second
> gets a conflict on those `assert` lines, and **the answer is neither side —
> it is 15 and 14.** Taking one branch's number wholesale silently under-counts
> the other's checks, which is the failure these two tests exist to catch. The
> `MOVED PIN` comments above each assertion state each delta so the resolution
> is arithmetic rather than a guess.

### 3. A stream error that said nothing

`_stream_run` had no `except` around `agent_run_events`, so a HealthClawError
killed the generator inside an open response: the connection dropped with no
`error` frame and the page just stopped (evidence §7b, "one defect visible here
and not in the issue"). It now emits `{"type": "error", "text": …}` carrying
the same generic sentence every other failure surface uses — never the
exception, which in production carries connection strings and correlation ids —
and closes. The run itself is durable and unaffected.

**Known edge, disclosed rather than fixed:** after the error frame the stream
ends *cleanly*, so `chat.js` resets its failure counter and reconnects. If
`get_agent_run` succeeds while `agent_run_events` keeps failing, the browser
will loop and print a new ⚠️ each time. On a full HealthClaw outage the GET
503s and the existing 5-failure cap stops it. Fixing `chat.js` is out of scope
here; flagging for a reviewer to rule on.

## Ran

- `uv run python -m pytest tests/ -q -p no:cacheprovider` →
  **3162 passed, 13 skipped, 1 xfailed** (ratchets and conformance included).
- `uv run ruff check .` → **All checks passed!**
- Mutation-checked each new pin by reverting the production change and watching
  the test go red: the readiness gate, the timeout budget, the stream `except`,
  the Dockerfile thread count, and the new prod_watch assertion.

New/changed tests: the 2×8 pin with its measurement
(`tests/test_careagents_container_roles.py`), `/healthz` answering 200 with
`run_workers_state: "unknown"` against a dead downstream and asking for a
bounded timeout, the client actually forwarding that timeout to the wire (a
fake would not have caught it being dropped — CareAgents fakes prove a call is
made, not accepted), `not_ready` still gating the deploy, a stream that raises
mid-run emitting a terminal error frame with nothing after it, and both worker
failure states exiting 1 from prod_watch with the state named in the line.

## Deploy note

**This changes a container setting, so it takes effect only on a deploy — and
deploying CareAgents is owner-only.** It is a manual `railway up` from a staged
directory for *both* the web and worker services
(`docs/runbooks/careagents-durable-worker.md`); pushing `main` does not deploy
CareAgents. Merging this changes nothing in production until someone does that.

**Read `CARE_WEB_WORKERS` on the deployed CareAgents service BEFORE deploying,
not after.** Nobody has read it. It is a Railway dashboard variable that is not
in this repo, and it was not readable from the probe either (evidence §10.1).
The Dockerfile defaults it to 2 and every measurement in this PR assumed 2, so
2 × 8 = 16 threads. If the dashboard sets anything else, this change ships a
thread count that was never measured at that width: 4 workers would be 32
threads, each of which holds a chat turn and polls HealthClaw ~4×/s at the
engine that also serves clinicians. The `--threads` half is not env-settable,
so the two cannot be reconciled after the fact from the dashboard alone — it
would take another image. This is the same shape as the `PORT` trap this repo
already hit: a Dockerfile value the platform overrides. If the variable is not
2, do not deploy this until the number has been re-measured or the variable
has been set to 2.

## What was NOT run

- **No load against production.** Nothing was sent to `careagents.cloud` or
  `app.healthclaw.io`. The probe's rig is committed on
  `probe/219-thread-saturation` for anyone re-measuring locally.
- **The 2×8 config was not re-measured here.** These numbers are the probe's,
  taken on its branch.
- **No container was built or run**, so the new thread count is proven only
  through the Dockerfile's own `CMD` dispatch, as the existing container-roles
  tests do.
- **The `/healthz` timeout was asserted, not waited on** — no test hangs a real
  socket for 25s.
- Nothing was measured about how often 6 concurrent turns actually happen; that
  needs production telemetry the probe also could not get (evidence §10.4).

`dependency-audit` is red on every open PR from npm advisories already fixed in
unmerged PR #544; it is unrelated to this change.

### DIFF
--- a/careagents/app.py
+++ b/careagents/app.py
@@ -45,6 +45,16 @@ logger = logging.getLogger(__name__)
 # strongest privacy control in the one place people read carefully.
 CONSENT_VERSION = "2026-08-01"
 
+# `/healthz` asks HealthClaw whether a run worker is present. That call gets
+# its own budget rather than the client's 25s chat timeout: as a
+# `(connect, read)` pair the worst case is 2.0s of network wait, which fits
+# inside the 4s the container probe allows (`careagents/healthcheck.py`) and
+# the Dockerfile's `HEALTHCHECK --timeout=5s`. The probe measured the old
+# shape taking 25.01s to answer at zero load, purely because it inherited the
+# chat timeout (docs/evidence/2026-09-03-probe-219-thread-saturation.md §8,
+# PR #573).
+_HEALTHZ_WORKER_TIMEOUT = (1.0, 1.0)
+
 # Local hard cap for the file-upload path (#227). The engine's
 # `internal/ingest-bundle` is the source of truth for the value; this
 # ceiling stops an oversized (or chunked / no-Content-Length) request
@@ -811,9 +821,10 @@ def create_app(config: Config | None = None,
     WORKERS_READY, WORKERS_NOT_READY, WORKERS_UNKNOWN = (
         "ready", "not_ready", "unknown")
 
-    def _worker_state() -> str:
+    def _worker_state(timeout=None) -> str:
         try:
-            status = hc.agent_worker_health(cfg.run_worker_stale_seconds)
+            status = hc.agent_worker_health(cfg.run_worker_stale_seconds,
+                                            timeout=timeout)
         except HealthClawError:
             return WORKERS_UNKNOWN
         return WORKERS_READY if status.get("available") else WORKERS_NOT_READY
@@ -838,7 +849,21 @@ def create_app(config: Config | None = None,
             "type": "accepted", "run_id": run_id,
             "next_cursor": cursor}) + "\n\n"
         while time.monotonic() - started < cfg.run_sse_timeout_seconds:
-            page = hc.agent_run_events(tenant, run_id, after=cursor, limit=100)
+            try:
+                page = hc.agent_run_events(tenant, run_id, after=cursor,
+                                           limit=100)
+            except HealthClawError:
+                # Without this the generator dies inside an open response and
+                # the browser is left with a stream that stopped for no stated
+                # reason (#219). Only this projection ends: the run is durable
+                # and keeps going, and a reconnect resumes at `cursor`. The
+                # text is the same generic sentence every other failure uses —
+                # an exception message here would be the first place engine
+                # internals reached a patient's screen.
+                logger.warning("run event stream failed for tenant %s", tenant)
+                yield "data: " + json.dumps({
+                    "type": "error", "text": GENERIC_FAILURE_TEXT}) + "\n\n"
+                return
             events = page.get("events") or []
             for event in events:
                 cursor = max(cursor, int(event.get("id") or 0))
@@ -1356,11 +1381,29 @@ def create_app(config: Config | None = None,
         load balancer would route real sign-ins straight into failure. It now
         round-trips a trivial query so the answer reflects reality.
 
-        Readiness has two inputs, and both gate the status code: the account
-        store must answer, and a fresh agent-run worker must be present. A
-        deployment with no worker can serve pages but cannot complete a chat
-        turn, so advertising it as ready would route people into failure the
-        same way the hard-coded accounts=True once did.
+        Readiness has two inputs and they are NOT symmetric (#219). The
+        account store is ours, so it gates the status code outright. The run
+        workers are HealthClaw's, so what gates the status code is their
+        ANSWER, not our ability to get one:
+
+        - Confirmed absent — HealthClaw answered, and no worker presence is
+          fresh. That deployment serves pages and cannot finish a chat turn.
+          It is the web-only deployment the durable-worker runbook tells
+          operators to expect a 503 for, and it stays a 503 deliberately: on
+          Railway the health check runs at the start of a deploy, and this is
+          a deploy that must not go live.
+        - Could not ask, inside `_HEALTHZ_WORKER_TIMEOUT` — that is a fact
+          about HealthClaw, not about this container. Reported as
+          `run_workers_state: "unknown"`; it does not fail the endpoint.
+          Failing it would block a CareAgents deploy on someone else's
+          outage, including the deploy that fixes it, while traffic stays on
+          a previous version with exactly the same dependency. The probe
+          measured the old shape taking 25.01s to answer 503 with nothing
+          running at all (evidence §8, PR #573).
+
+        Admission is unchanged, and it is the gate that protects people:
+        `POST /api/chat` still refuses a turn in BOTH states, so nothing
+        routes anyone into a turn no worker can run.
 
         It also reports which build is running (#258). That is telemetry, not
         a gate: an absent marker reports "unknown" and changes nothing about
@@ -1371,14 +1414,18 @@ def create_app(config: Config | None = None,
         including dependencies it does not control.
         """
         accounts_ok = svc.ping()
+        workers_state = _worker_state(timeout=_HEALTHZ_WORKER_TIMEOUT)
         # `run_workers` stays a boolean readiness VERDICT, not a claim about
         # the workers: false is correct when we could not confirm them, and
         # two runbooks plus the container-roles test read it as a bool.
-        workers_ok = _worker_state() == WORKERS_READY
-        ready = accounts_ok and workers_ok
+        # `run_workers_state` carries the claim the bool cannot make (#410) —
+        # and now also the difference the status code turns on.
+        workers_ok = workers_state == WORKERS_READY
+        ready = accounts_ok and workers_state != WORKERS_NOT_READY
         body = {"status": "ok" if ready else "degraded",
                 "provider": cfg.provider, "accounts": accounts_ok,
                 "run_workers": workers_ok,
+                "run_workers_state": workers_state,
                 "build": cfg.build_sha, "built_at": cfg.build_time}
         return jsonify(body), (200 if ready else 503)
 
--- a/careagents/app.py
+++ b/careagents/app.py
@@ -55,6 +55,23 @@ CONSENT_VERSION = "2026-08-01"
 # PR #573).
 _HEALTHZ_WORKER_TIMEOUT = (1.0, 1.0)
 
+# The same call on the ADMISSION path (`POST /api/chat`, and the iMessage
+# relay ingress). It gets its own budget rather than sharing the readiness
+# probe's, because the stakes are opposite: `/healthz` must answer inside a
+# platform probe window and a wrong answer costs a deploy, while a turn
+# refused too eagerly costs a patient their question. A worker-health call
+# does two database round-trips on the engine that also serves clinicians, so
+# 1s of read is too tight to refuse a real user on.
+#
+# Left unbounded, this inherits the client's 25s CHAT timeout — measured at
+# 25.00s against a HealthClaw that accepts the connection and never answers.
+# That is a gunicorn thread held for 25 seconds *to refuse a turn*, in exactly
+# the dependency outage `/healthz` now reports as `unknown` and lets deploy,
+# and it eats the thread headroom this same change bought. 4s is a ceiling
+# chosen to end the 25s hold while leaving room for a loaded engine; it is not
+# a measured optimum, and a p99 from production should replace it.
+_ADMISSION_WORKER_TIMEOUT = (1.0, 3.0)
+
 # Local hard cap for the file-upload path (#227). The engine's
 # `internal/ingest-bundle` is the source of truth for the value; this
 # ceiling stops an oversized (or chunked / no-Content-Length) request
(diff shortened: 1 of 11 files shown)


aks129:
## QA review — **FAIL as submitted**, one HIGH fixed on the branch

Reviewed against the issue, the probe evidence, and a running pair (a real CareAgents web process under gunicorn against a real HealthClaw, not the unit fakes). Three defects found; all three are fixed in `3b66e7c` and `2578cee`, now on this branch. Nothing merged, no auto-merge armed.

### HIGH — the split gate admits a misconfigured deployment

`careagents/app.py:841` routed **every** `HealthClawError` to `unknown`, and `agent_worker_health` raises one for any status outside `(200, 503)` as well as for a dead socket. So "HealthClaw answered 403 because this container's mint secret is wrong" was filed as "we could not ask", and `/healthz` answered 200 — which on Railway means the deploy goes live.

Measured, before the fix:

| | wiring | `/healthz` |
|---|---|---|
| A | correct, no run worker | 503 `not_ready` — right |
| B | wrong `HEALTHCLAW_MINT_SECRET` (403 from HealthClaw) | **200 `"status":"ok"`** |
| C | wrong `HEALTHCLAW_BASE`, nothing listening | **200 `"status":"ok"`** |
| D | wrong `HEALTHCLAW_BASE` path root (404 from HealthClaw) | **200 `"status":"ok"`** |

B and D are faults in the artifact being deployed — exactly the case this PR's own ruling says must stay blocked. Both ship a CareAgents that serves pages and cannot finish a single chat turn.

The fix carves on **4xx**, not "any HTTP status": a 4xx means HealthClaw understood the request and refused something about *it*, so the fault is ours. Transport failure, 5xx and a 200 carrying a proxy interstitial all stay `unknown`, because gating on those would re-block the deploy that fixes someone else's outage. After: B and D answer 503 `run_workers_state: "rejected"`; A and C unchanged.

**C is not fixed, and I do not think it can be.** From inside the container a typo'd host is indistinguishable from a host that is simply down. It is caught within one prod_watch cycle (`unknown` reddens the new check), but it is a residual that belongs to the CTO, not to me.

`rejected` is a fourth state rather than a collapse into `not_ready`, because the two send an operator to different dashboards. Admission gets `run_workers_rejected` for the same reason — reusing `run_workers_unavailable` would file "our secret is wrong" as "the workers are down", which is #410 one layer along.

### MEDIUM — admission still holds a thread for 25 seconds

`/healthz` got a bounded budget; `careagents/app.py:958` and `:1322` did not, so they inherit the client's 25s **chat** timeout. Measured against a HealthClaw that accepts the connection and never answers:

```
  /healthz's bounded call    ->  1.01s
  admission's unbounded call -> 25.00s
```

A gunicorn thread held for 25 seconds *to refuse a turn*, in exactly the outage `/healthz` now reports as `unknown` and lets deploy — and it spends the headroom the thread change buys. Now bounded at `(1.0, 3.0)`. That ceiling ends the 25s hold; it is not a measured optimum, and a production p99 should replace it.

### MEDIUM — the terminal error frame starts an unbounded retry loop

The PR discloses this and asks a reviewer to rule. **Ruling: in scope**, because this change is what makes the loop unbounded. The new frame ends the SSE response *cleanly*, so `chat.js` resets `reconnectFailures` and reconnects every 400ms indefinitely, printing a fresh ⚠️ each pass — roughly 2.5 req/s per open tab aimed at the engine that also serves clinicians, on the phone this product is used from. Before the change the generator died and the client saw a truncated stream, which the 5-failure cap would stop *if* the truncation surfaced to `fetch`. The new behaviour is unbounded by construction, where the old one was bounded by accident.

Fixed in `chat.js` by treating `error` as terminal, which every producer already does (`agent.py:427`, `:469`, `run_turn_to_message`, and the new stream handler all return after it). Verified by reading the producers and the client loop — there is no e2e coverage of the SSE reconnect path to extend.

### Verified good

- **The #549 arithmetic in the merge warning is correct.** Counted: `main` has 11 `check(` sites and pins 12/11; #549 adds 2 and pins 14/13; this PR adds 1 and pins 13/12. Combined is 14 sites, so **15 and 14** is right. One thing the warning does not mention: #549 keeps `HEALTHY = {"status": "ok", "provider": "openai", "accounts": True}` without `run_workers`, so whoever resolves must take this branch's `HEALTHY` as well, or the new check goes red inside #549's own tests.
- **The monitor reddens for both failure states, names the owner of each, and does not run when `/healthz` is unreadable.** That last branch had no test; I added one.
- **The thread change is honestly documented.** The PR says plainly that `CARE_WEB_WORKERS` was never read, that `--threads` is not env-settable, and that the two cannot be reconciled from the dashboard afterwards. I could not read it either: the Railway CLI here is linked to a different project, and re-linking is a state change I will not make during a review. **The deploy note stands — read that variable before deploying.**
- **`GENERIC_FAILURE_TEXT` is `"Something went wrong on our side."`** No internals, no PHI, and the test asserting the exception text never reaches the body is real.
- **Every new pin is honest.** I mutation-tested all of them: the readiness gate, the per-call timeout reaching the wire, the stream `except`, the Dockerfile thread count (13 red), and the prod_watch check (3 red). Plus the five I added.

### Not covered

- No load test of 2×8. Those numbers are the probe's, on its branch, as the PR says.
- No container built or run; the thread count is proven through the `CMD` dispatch only.
- Admission's refusal was exercised through the unit path, not a browser session against the running pair.
- Nothing about how often 6 concurrent turns actually happen.

### Housekeeping

- `dependency-audit` is red on every open PR from npm advisories already fixed in unmerged **PR #544**. Unrelated to this branch.
- **This takes effect only on deploy**, and deploying CareAgents is a manual `railway up` from a staged directory for *both* services. Merging changes nothing in production.

Counts on this branch after the fixes: `uv run ruff check .` → `All checks passed!`; `uv run python -m pytest tests/ -q -p no:cacheprovider` → **3173 passed, 13 skipped, 1 xfailed** (3162 as submitted; +11 new tests).

Issue #574, opened 2026-09-03
the privacy-policy link inside every FHIR resource and both discovery documents returns 400 to its own reader


aks129:
Found while reviewing the MCP metadata work, then confirmed against production on 2026-09-03.

```
curl -o /dev/null -w '%{http_code}' https://app.healthclaw.io/r6/fhir/docs/privacy-policy
400   {"code":"security","diagnostics":"X-Tenant-Id header is required"}

curl -o /dev/null -w '%{http_code}' https://app.healthclaw.io/r6/fhir/docs/privacy-policy -H 'X-Tenant-Id: desktop-demo'
200
```

The route is behind the tenant gate. Three different documents point at it, and in each case its intended reader has no tenant.

| Where | Field | Who follows it |
|---|---|---|
| Every FHIR resource we return | `_disclaimer.url` (`r6/health_compliance.py:67`) | anyone reading a resource outside our own client |
| OAuth discovery | `service_documentation` (`r6/oauth.py:188`) | any client doing discovery |
| MCP protected-resource metadata (PR #556) | `resource_documentation` | a client that has not authenticated yet, by construction |

So the disclaimer we attach to every clinical resource, to say where our terms live, points at a URL that errors for the person it is aimed at. A partner evaluating us reads the discovery document first and finds a broken link to our policy.

The public marketing policy page at `/privacy` returns 200, so this is the API-flavoured copy specifically.

## Fix

Exempt this one route from the tenant gate. A policy document has nothing tenant-specific in it, needs no patient data to render, and is the one page that must be readable by someone who has not connected yet.

Check the gate's shape before assuming where the exemption goes: it is applied by a `before_request` hook on the blueprint, so the exemption belongs there rather than in the view.

If for some reason the route must stay gated, the alternative is to repoint all three fields at `/privacy` and delete this one, but that leaves the API terms unreachable from the API, which is worse.

## Tests

- The route answers 200 with no `X-Tenant-Id` header.
- It still emits no audit event and reads no tenant data.
- A pinning test that every URL published in discovery documents and in `_disclaimer` resolves without a tenant header, so a future field cannot quietly join the broken set.

## Related

PR #556, RFC 9728 §2 (`resource_documentation` is optional, and a broken one is worse than none).

aks129:
Fixed on main by #583 (merged 2026-09-05 during the queue walk), which publishes URLs a partner can follow: the privacy-policy link resolves and the discovery documents carry the served scheme. Closing as shipped.

Issue #567, opened 2026-09-03
SMART configuration advertises http:// endpoints on an https-only deployment


aks129:
Measured against production on 2026-09-03, not inferred.

```
curl -s https://app.healthclaw.io/r6/fhir/.well-known/smart-configuration
{"authorization_endpoint":"http://app.healthclaw.io/r6/fhir/oauth/authorize",
 "registration_endpoint":"http://app.healthclaw.io/r6/fhir/oauth/register",
 "revocation_endpoint":"http://app.healthclaw.io/r6/fhir/oauth/...
```

Every URL in the SMART configuration is **`http://`**, on a deployment that is only reachable over `https://`.

## Cause

The app builds these from `request.host_url` (`r6/oauth.py:176`, `:196`, `r6/routes.py:340`, `:368`, `:1128`, `:1139`, `:1421`, `:1447`). The platform terminates TLS and forwards the request over plain HTTP, so Flask sees `http` unless `X-Forwarded-Proto` is honoured. There is no `ProxyFix` anywhere in the app, so it is not honoured.

The Host header **is** preserved — the same measurement shows `app.healthclaw.io`, the custom domain, not an internal name — so this is the scheme alone.

## Why it matters

The SMART configuration is the first document an integrating partner reads. SMART App Launch and OAuth 2.1 require https endpoints; a conformant client rejects these or, worse, follows them and downgrades. Anyone who tries our authorization endpoint from this document is sent to a URL we do not serve.

It also lands in FHIR `Bundle.entry.fullUrl` and the `self` links, so every search result we return carries an http identifier for a resource that only exists over https.

## Fix

Honour the forwarded scheme once, centrally, rather than at each call site:

- Add `ProxyFix` with `x_proto=1, x_host=1` (and only those) at app creation, so `request.host_url` is right everywhere at once, **or**
- introduce one `public_base_url()` helper built from configuration and use it at every site listed above.

Prefer whichever the team can pin with a test. The trap to avoid is fixing the SMART document alone and leaving `fullUrl` wrong, since the same helper feeds both.

## Tests

- A request carrying `X-Forwarded-Proto: https` produces `https://` in the SMART configuration, the capability statement, and `Bundle.entry.fullUrl`.
- A request without it still works locally over http.
- A pinning test over every site that builds a URL from the request, so a new one cannot quietly join the list.

## Related

[...]

PR #583, opened 2026-09-04
candidate: publish URLs a partner can actually follow (#567, #574)

aks129:
Refs #567. Refs #574. Deliberately not `Closes` — merging this deploys it, and neither issue can be closed until the two production curls quoted below return `https://` and `200`. Close them on that measurement, not on this merge.

## What

Two defects in what we publish to partners, one cause: the URL we hand out is not one the reader can use.

**#567 — the protocol.** `ProxyFix(x_for=0, x_proto=1, x_host=0, x_port=0, x_prefix=0)` at app creation (`main.py:395-430`), so `request.host_url` is right everywhere at once. Protocol only — every other forwarded header stays untrusted, for the reason below.

**#574 — the policy link.** `/r6/fhir/docs/privacy-policy` added to `_EXEMPT_EXACT_PATHS` (`r6/routes.py:175`), the frozenset the blueprint's `before_request` tenant hook consults. Exact path, not a prefix or a suffix: `/r6/fhir/Patient/privacy-policy` stays gated (verified by request, below).

`tests/test_published_urls.py` — ten tests, two of them pins over every published URL rather than the two fields the issues named.

## This cannot be shown fixed until it deploys

Both defects only appear behind the platform's proxy. Locally the edge has to be simulated by sending `X-Forwarded-Proto` by hand, which proves the middleware reads the header — not that the platform sends it in the form assumed here. **The confirming measurement is re-running the two production requests quoted below, after this deploys.** Until then the evidence in Ran shows the mechanism works, not that the defect is gone. That is why this PR references the issues rather than auto-closing them.

## Why

Measured against production on 2026-09-03, quoted from the issues:

```
curl -s https://app.healthclaw.io/r6/fhir/.well-known/smart-configuration
{"authorization_endpoint":"http://app.healthclaw.io/r6/fhir/oauth/authorize",
 "registration_endpoint":"http://app.healthclaw.io/r6/fhir/oauth/register",
 "revocation_endpoint":"http://app.healthclaw.io/r6/fhir/oauth/...
```

```
curl -o /dev/null -w '%{http_code}' https://app.healthclaw.io/r6/fhir/docs/privacy-policy
400   {"code":"security","diagnostics":"X-Tenant-Id header is required"}

curl -o /dev/null -w '%{http_code}' https://app.healthclaw.io/r6/fhir/docs/privacy-policy -H 'X-Tenant-Id: desktop-demo'
200
```

The hostname is preserved in that first measurement — `app.healthclaw.io`, the custom domain — so #567 is the protocol alone. SMART App Launch and OAuth 2.1 require https endpoints; a conformant client rejects these or follows them and downgrades. The same wrong scheme lands in `Bundle.entry.fullUrl` and every search `self` link, so each search result carries an http identifier for a resource that only exists over https.

For #574, the reader of all three pointers has no tenant by construction: the `_disclaimer` on every clinical resource, `service_documentation` in the discovery document, and the `resource_documentation` in the MCP protected-resource metadata from #556 (a client that has not authenticated yet).

### Why the middleware, and why it is in `main.py`

The issue offered two shapes. The middleware wins on the thing that caused this: a helper is only as complete as the last person who remembered to call it, and #567's whole failure mode is "the discovery document was fixed and `fullUrl` was not". `ProxyFix` also covers the three sites outside the issue's list — `r6/wearables/routes.py:138`, `r6/command_center/routes.py:629`, `app.py:485` — and the next site nobody has written yet.

It lives in `main.py` because a WSGI middleware belongs at application creation and nowhere else. `r6/routes.py` is at 3924 of its 3927-line ratchet, and it would have been the wrong home for this even with room to spare; the file took one line, the exemption.

### Trusting `X-Forwarded-Proto`, and nothing else

That is a proxy-trust decision, and the reasoning is in a comment beside the call. Safe here because the platform's edge is the only route to the container (the container publishes no port of its own); the header is trusted for exactly one hop, so the value the edge appends beats anything a client sends; and nothing in this app keys a security decision on the request scheme (`SESSION_COOKIE_SECURE` is static config, not derived from it).

**`x_host` is `0`, and the reason is evidence rather than taste.** The first draft of this change trusted the forwarded host too. It should not: both settings fail the same way when they are wrong — by publishing something a partner cannot use — so the tiebreaker is which one has been measured.

- **Measured:** the protocol is wrong without the fix. Production answers `http://` on an https-only deployment.
- **Measured:** the hostname is already correct. The same production request returns `app.healthclaw.io`, the custom domain, not an internal name.
- **Never measured:** what this platform puts in `X-Forwarded-Host`, or whether it sends one at all.

Trusting an unmeasured header in preference to a measured one is the wrong direction. The protocol trust earns its risk because the protocol is demonstrably broken without it; the host trust buys nothing we do not already have, at the risk of publishing an internal hostname. `test_the_forwarded_host_is_not_trusted` pins it, so re-widening is a decision someone makes on purpose rather than a default nobody notices.

`x_for` stays `0` for a different reason — `r6/rate_limit.py` does its own hop-counted `X-Forwarded-For` parse, and letting `ProxyFix` rewrite `REMOTE_ADDR` would give the limiter a second, differently-trusted answer to the same question.

### What the pins cover

`test_every_published_url_carries_the_proxy_protocol` and `test_every_published_pointer_resolves_without_a_tenant` walk both discovery documents, the CapabilityStatement and the `_disclaimer`, and assert every URL is https behind the proxy and resolves without a tenant header. Two documented exclusion sets, so a new exclusion is a decision rather than an oversight: canonical FHIR identifier hosts (`hl7.org` and friends — `http://hl7.org/fhir/...` is a name, not an endpoint), and two same-host identifiers that are not followable pointers (`implementation.url`, and the `$ingest-context` OperationDefinition).

One note the issue text overstates: `_disclaimer.url` is a **relative** path (`r6/health_compliance.py:67`), so it was only ever a #574 problem, never a #567 one. It is left relative — `add_disclaimer` is a pure function called outside a request context in existing tests. The pin resolves it against the request base and asserts on the result.

## Ran

`uv run ruff check .`

```
All checks passed!
```

`uv run python -m pytest tests/ -q -p no:cacheprovider`

```
3167 passed, 13 skipped, 1 xfailed, 2 warnings in 125.87s (0:02:05)
```

Ratchets and conformance are in that run and green; `r6/routes.py` is 3925 lines against a pin of 3927, so `_GOD_MODULE_LINES` is untouched. The two suites that send `X-Forwarded-*` headers of their own — `tests/test_rate_limit_client_ip.py` and `tests/test_security_hardening.py` — were also run on their own and pass, which is the check that `x_for=0` left the rate limiter's hop counting alone.

Mutation-tested against the shipped setting: with `ProxyFix` removed, 6 of 10 fail; with the exemption line removed, 3 of 10 fail; with `x_host` re-widened to `1`, 1 of 10 fails.

Then against a running app (`flask --app main init-db`, `seed-demo`, port 5443):

```
$ curl -s -H 'X-Forwarded-Proto: https' .../r6/fhir/.well-known/smart-configuration
{"authorization_endpoint":"https://127.0.0.1:5443/r6/fhir/oauth/authorize", …

$ curl -s .../r6/fhir/.well-known/smart-configuration          # no header, local dev
{"authorization_endpoint":"http://127.0.0.1:5443/r6/fhir/oauth/authorize", …

$ curl -s -H 'X-Forwarded-Proto: https' .../r6/fhir/metadata
implementation.url  https://127.0.0.1:5443/r6/fhir
authorize           https://127.0.0.1:5443/r6/fhir/oauth/authorize
token               https://127.0.0.1:5443/r6/fhir/oauth/token
register            https://127.0.0.1:5443/r6/fhir/oauth/register

$ curl -s -H 'X-Forwarded-Proto: https' -H 'X-Tenant-Id: desktop-demo' '.../r6/fhir/Observation?_count=1'
self     https://127.0.0.1:5443/r6/fhir/Observation?_count=1
fullUrl  https://127.0.0.1:5443/r6/fhir/Observation/demo-obs-bp

$ curl -s -o /dev/null -w '%{http_code}' .../r6/fhir/docs/privacy-policy        # no headers
200
$ curl -s -o /dev/null -w '%{http_code}' .../r6/fhir/Observation/demo-obs-bp    # no headers
400
$ curl -s -o /dev/null -w '%{http_code}' .../r6/fhir/Patient/privacy-policy     # no headers
400
```

The forwarded host is ignored, checked by request on the same running app:

```
$ curl -s -H 'X-Forwarded-Proto: https' .../r6/fhir/.well-known/smart-configuration
https://127.0.0.1:5457/r6/fhir/oauth/authorize

$ curl -s -H 'X-Forwarded-Proto: https' -H 'X-Forwarded-Host: container.internal:8080' ...
https://127.0.0.1:5457/r6/fhir/oauth/authorize     # unchanged
```

The last two are the check that the exemption did not widen the gate: a real resource read, and a resource whose id looks like the exempt path, both still demand a tenant.

Following the published pointer end to end, with no tenant header:

```
$ curl -s .../r6/fhir/.well-known/oauth-authorization-server | jq -r .service_documentation
https://127.0.0.1:5443/r6/fhir/docs/privacy-policy
$ curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:5443/r6/fhir/docs/privacy-policy
200
```

## What was NOT run

- **The actual defect** — see the section above. The two production requests are the confirming measurement, and they can only be run after deploy.
- The e2e Playwright suite (`e2e/tests/api.spec.ts`) — it needs a running stack and browsers. Its privacy-policy case sends a tenant header, so it is unaffected either way.
- Postgres CI lane, MCP server, CareAgents — untouched by this change.
- `dependency-audit` is red on every open PR from npm advisories already fixed in unmerged PR #544. Not caused by this branch.

## Deliberately not touched

- `_disclaimer.url` stays relative (see above).
- `app.py:485`, `r6/wearables/routes.py:138`, `r6/command_center/routes.py:629` build URLs from the request too. `ProxyFix` fixes all three for free; none of them is asserted here, because each belongs to a flow with its own tests and its own reviewer.
- `OAUTH_ISSUER` still wins over the request-derived base when set. Unchanged, and correct.

### DIFF
--- a/main.py
+++ b/main.py
@@ -358,6 +358,40 @@ def _register_request_hooks(flask_app: Flask) -> None:
         return response
 
 
+#: The only schemes this application is ever reachable over, and therefore
+#: the only values `X-Forwarded-Proto` may carry into `wsgi.url_scheme`.
+#: Anything else is either an injection or a scheme we do not serve, and both
+#: produce a published URL a partner cannot follow — which is the whole point
+#: of honouring the header in the first place.
+_SERVED_URL_SCHEMES = frozenset({"http", "https"})
+
+
+def _drop_untrusted_forwarded_proto(wsgi_app):
+    """Remove `X-Forwarded-Proto` unless it names a scheme we serve.
+
+    Runs OUTSIDE ProxyFix (so before it), because ProxyFix modifies the
+    environ in place and then calls down — there is no "after" to inspect
+    from a wrapper. Sanitising the input is also the smaller change: ProxyFix
+    stays stock, and the rule is one comparison anyone can read.
+
+    See the comment at the ProxyFix call in `create_app` for the measurements
+    this exists to stop.
+    """
+    def middleware(environ, start_response):
+        raw = environ.get("HTTP_X_FORWARDED_PROTO")
+        if raw is not None:
+            # ProxyFix's trust depth of 1 reads the rightmost entry, so that
+            # is the one that has to be legitimate.
+            scheme = raw.rsplit(",", 1)[-1].strip().lower()
+            if scheme in _SERVED_URL_SCHEMES:
+                environ["HTTP_X_FORWARDED_PROTO"] = scheme
+            else:
+                del environ["HTTP_X_FORWARDED_PROTO"]
+        return wsgi_app(environ, start_response)
+
+    return middleware
+
+
 def create_app(settings: Mapping[str, Any] | None = None) -> Flask:
     """Create and configure an independent Flask application instance."""
     supplied_settings = dict(settings or {})
@@ -429,6 +463,38 @@ def create_app(settings: Mapping[str, Any] | None = None) -> Flask:
     flask_app.wsgi_app = ProxyFix(
         flask_app.wsgi_app, x_for=0, x_proto=1, x_host=0, x_port=0, x_prefix=0
     )
+    # ...but only when the value is a scheme we actually serve. ProxyFix
+    # copies the header into `wsgi.url_scheme` VERBATIM, with no validation,
+    # and every published URL is `scheme + "://" + host`. An arbitrary string
+    # there is not a scheme, it is a URL prefix. Measured on a running app:
+    #
+    #   X-Forwarded-Proto: https://evil.example/?
+    #     -> authorization_endpoint:
+    #        https://evil.example/?://127.0.0.1:5511/r6/fhir/oauth/authorize
+    #   X-Forwarded-Proto: javascript
+    #     -> javascript://127.0.0.1:5511/r6/fhir/oauth/authorize
+    #   X-Forwarded-Proto: ws
+    #     -> every route in the app answers 405, because werkzeug's router
+    #        treats a ws/wss scheme as a websocket request and no HTTP rule
+    #        matches
+    #
+    # That value reaches more than a document the caller is already reading:
+    # `app.py:485` builds the quickstart link in the welcome email from
+    # `request.url_root`, so a poisoned scheme is a link WE send to a third
+    # party from our own sender, and `r6/wearables/routes.py:138` builds the
+    # `redirect_uri` handed to an external OAuth provider.
+    #
+    # Whether a client can set this at all depends on the edge overwriting or
+    # appending the header, which has never been measured on this platform —
+    # the same "unmeasured, so do not trust it" standard that keeps x_host at
+    # 0 above. The allowlist makes the answer not matter.
+    #
+    # Judged on the RIGHTMOST value because that is the one ProxyFix takes
+    # with a trust depth of 1 (`X-Forwarded-Proto: https,http` yields `http`).
+    # Invalid means the header is dropped, not corrected: the app then falls
+    # back to the scheme the connection actually arrived on, which is exactly
+    # the pre-#567 behaviour and always safe.
+    flask_app.wsgi_app = _drop_untrusted_forwarded_proto(flask_app.wsgi_app)
 
     db_uri = flask_app.config["SQLALCHEMY_DATABASE_URI"]
     if (
--- a/main.py
+++ b/main.py
@@ -19,6 +19,7 @@ from typing import Any
 
 import click
 from flask import Flask, g, request as flask_request
+from werkzeug.middleware.proxy_fix import ProxyFix
 from models import db
 from r6.database_migrations import upgrade_database
 from r6.runtime_config import validate_runtime_environment
@@ -390,6 +391,33 @@ def create_app(settings: Mapping[str, Any] | None = None) -> Flask:
             SESSION_COOKIE_HTTPONLY=True,
             SESSION_COOKIE_SAMESITE="Lax",
         )
+
+    # Every URL this app publishes is built from ``request.host_url``: the
+    # SMART configuration, the OAuth discovery document, the
+    # CapabilityStatement, ``Bundle.entry.fullUrl`` and every search self
+    # link. The hosting platform terminates TLS and forwards to the container
+    # over plain HTTP, so without this Flask sees scheme "http" and we
+    # advertise http:// endpoints on a deployment reachable only over https
+    # (#567). One middleware fixes every call site at once, including ones
+    # nobody has written yet — the alternative, a helper applied per site,
+    # is only ever as complete as the last person to remember it.
+    #
+    # Trusting X-Forwarded-* is a proxy-trust decision. It is safe here
+    # because the platform's edge is the only route to the container (the
+    # container publishes no port of its own), because each header is
+    # trusted for exactly one hop so the value the edge appends is the one
+    # that wins over anything a client sends, and because nothing in this
+    # app keys a security decision on the request scheme —
+    # SESSION_COOKIE_SECURE above is static config, not derived from it. A
+    # deployment that ever exposes this container directly must drop this.
+    #
+    # x_for stays 0 deliberately: r6/rate_limit.py does its own hop-counted
+    # X-Forwarded-For parse, and letting ProxyFix rewrite REMOTE_ADDR would
+    # give the limiter a second, differently-trusted answer for the same
+    # question.
+    flask_app.wsgi_app = ProxyFix(
+        flask_app.wsgi_app, x_for=0, x_proto=1, x_host=1, x_port=0, x_prefix=0
+    )
 
     db_uri = flask_app.config["SQLALCHEMY_DATABASE_URI"]
     if (
--- a/main.py
+++ b/main.py
@@ -402,21 +402,32 @@ def create_app(settings: Mapping[str, Any] | None = None) -> Flask:
     # nobody has written yet — the alternative, a helper applied per site,
     # is only ever as complete as the last person to remember it.
     #
-    # Trusting X-Forwarded-* is a proxy-trust decision. It is safe here
+    # Trusting X-Forwarded-Proto is a proxy-trust decision. It is safe here
     # because the platform's edge is the only route to the container (the
-    # container publishes no port of its own), because each header is
-    # trusted for exactly one hop so the value the edge appends is the one
-    # that wins over anything a client sends, and because nothing in this
-    # app keys a security decision on the request scheme —
-    # SESSION_COOKIE_SECURE above is static config, not derived from it. A
-    # deployment that ever exposes this container directly must drop this.
+    # container publishes no port of its own), because the header is trusted
+    # for exactly one hop so the value the edge appends is the one that wins
+    # over anything a client sends, and because nothing in this app keys a
+    # security decision on the request scheme — SESSION_COOKIE_SECURE above
+    # is static config, not derived from it. A deployment that ever exposes
+    # this container directly must drop this.
     #
-    # x_for stays 0 deliberately: r6/rate_limit.py does its own hop-counted
-    # X-Forwarded-For parse, and letting ProxyFix rewrite REMOTE_ADDR would
-    # give the limiter a second, differently-trusted answer for the same
-    # question.
+    # EVERY OTHER FORWARDED HEADER IS UNTRUSTED, and the reason is evidence
+    # rather than taste. Both x_proto and x_host fail the same way when they
+    # are wrong — by publishing something a partner cannot use — so the
+    # tiebreaker is which one we have measured. The protocol is measurably
+    # wrong without the fix; the hostname was measured as already correct
+    # (production returns the custom domain, not an internal name), and what
+    # X-Forwarded-Host contains here has never been measured at all.
+    # Trusting an unmeasured header in preference to a measured one is the
+    # wrong direction, so x_host stays 0: it would buy nothing we do not
+    # already have, at the risk of publishing an internal hostname.
+    #
+    # x_for stays 0 for a different reason: r6/rate_limit.py does its own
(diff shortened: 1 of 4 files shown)


aks129:
## QA review — **FAIL as submitted**, one MEDIUM fixed on the branch

Reviewed against #567, #574 and a running app (`init-db`, `seed-demo`, port 5511, then 5519 to re-verify). One defect found and fixed in `cd03122`, now on this branch. The exemption survived everything I threw at it. Nothing merged, no auto-merge armed.

### MEDIUM — the forwarded protocol is trusted without being validated

`ProxyFix` copies `X-Forwarded-Proto` into `wsgi.url_scheme` verbatim, and every URL this app publishes is `scheme + "://" + host`. A value that is not a scheme is therefore a URL prefix. Measured on the running app:

```
X-Forwarded-Proto: https://evil.example/?
  "authorization_endpoint":
    "https://evil.example/?://127.0.0.1:5511/r6/fhir/oauth/authorize"

X-Forwarded-Proto: javascript
  "javascript://127.0.0.1:5511/r6/fhir/oauth/authorize"

X-Forwarded-Proto: ws     (and wss)
  /r6/fhir/health              405
  /r6/fhir/metadata            405
  /r6/fhir/docs/privacy-policy 405
  /r6/fhir/Observation         405
```

The `ws` case is the same defect with a different blast radius: werkzeug's router treats a `ws`/`wss` scheme as a websocket request, no HTTP rule matches, and the whole app becomes unreachable for that caller. One header, one request.

The value also does not stay inside the document the caller is already reading. `app.py:485` builds the quickstart link in the welcome email from `request.url_root`, so a poisoned scheme becomes a link **we send to a third party from our own sender**; `r6/wearables/routes.py:138` builds the `redirect_uri` handed to an external OAuth provider. Both are sites this PR correctly claims ProxyFix fixes for free — it does, and it exposes them here by the same mechanism. (The email path is shown by inspection: Resend is not configured locally, so no email was sent.)

**On exploitability, honestly:** whether a caller can set this at all depends on whether the platform edge overwrites, appends, or passes the header through. I measured that werkzeug takes the rightmost value at trust depth 1 (`https,http` → `http`), so an edge that overwrites *or* appends is safe and only pass-through is not. What Railway does is not measurable from here. That is the same gap the PR names for `x_host` — and the PR applies the "never measured, so do not trust it" standard to the *header* while assuming the answer for its *value*: "the value the edge appends is the one that wins over anything a client sends" is asserted, not measured. An allowlist is cheaper than the measurement.

Fixed with a nine-line middleware outside ProxyFix that drops the header unless the rightmost value is `http` or `https`. It sanitises the input rather than correcting the output because ProxyFix mutates the environ in place and then calls down — there is no "after" for a wrapper to inspect. Re-verified live: injected values fall back to the real connection scheme, `ws`/`wss` answer 200 again, and #567 is untouched (`https` → https, `HTTPS` → https, `https,http` → http, no header → http, `fullUrl` and the `self` link still https).

### The exemption does not widen the gate

Every case from the brief, run against the app with **no `X-Tenant-Id`**:

```
GET  /r6/fhir/docs/privacy-policy                               200
GET  /r6/fhir/docs/privacy-policy/                              404
GET  //r6/fhir/docs/privacy-policy                              200   (same document)
GET  /r6/fhir/docs/privacy-policy%20                            400
GET  /r6/fhir/docs/privacy%2Dpolicy                             200   (same route after decode)
GET  /r6/fhir/docs/Privacy-Policy                               400
GET  /r6/fhir/DOCS/privacy-policy                               400
GET  /r6/fhir/Patient/privacy-policy                            400
GET  /r6/fhir/Observation/privacy-policy                        400
GET  /r6/fhir/docs/privacy-policy/../../Patient/demo-patient-1  400
GET  /r6/fhir/docs/./privacy-policy                             404
GET  /r6/fhir/docs/privacy-policy?_tenant=desktop-demo          200
POST /PATCH /DELETE                                             405
OPTIONS / HEAD                                                  200
//r6/fhir/Patient/demo-patient-1                                400
//r6/fhir/Observation                                           400
```

Same bytes for anonymous, `desktop-demo` and another tenant (identical SHA), and no `AuditEventRecord` written. Nothing reaches tenant data. The structural argument holds too: `_is_exempt_discovery_path` has exactly two call sites (the tenant hook and the read-auth hook), and `privacy_policy()` is a `jsonify` of constants.

### LOW — the exemption is method-agnostic, and this path has a write route

`PUT /r6/fhir/docs/privacy-policy` matches `r6.update_resource` with `resource_type="docs"`, `resource_id="privacy-policy"`, so it now reaches the handler with no tenant instead of being refused at the door:

```
PUT /r6/fhir/docs/privacy-policy      (no tenant, no step-up, no X-Human-Confirmed)
  400 {"code":"not-supported","diagnostics":"Resource type docs is not supported"}

PUT /r6/fhir/Observation/zzz          (control)
  400 {"code":"security","diagnostics":"X-Tenant-Id header is required"}
```

Not fixed, and I do not think it should be here. It is a pre-existing class rather than something this PR introduced — the url_map shows `POST` on `/r6/fhir/metadata`, `/health` and `/$conformance` all reaching `r6.create_resource` the same way — and there are two backstops behind it: the supported-type allowlist, then `require_grant(scope=Scope.WRITE)` inside `update_resource`. Method-scoping `_is_exempt_discovery_path` would also touch the `/oauth/` and `/.well-known/` prefixes where POST is legitimate, which is wider than a QA commit should be. Worth an issue: this is the first two-segment entry in `_EXEMPT_EXACT_PATHS`, and two segments is what makes it collide with `/<resource_type>/<id>`.

### The pins are real

I mutation-tested them rather than trusting the counts:

| mutation | result |
|---|---|
| ProxyFix removed | 6 of 10 red — matches the PR |
| exemption line removed | 3 of 10 red — matches the PR |
| a **new** hardcoded `http://` endpoint added to the SMART config | 2 red |
| a **new** pointer aimed at a tenant-gated route | 1 red |
| a new pointer at a POST-only route (`/oauth/token`) | green |

So the answer to "would it catch a newly added URL that is wrong" is **yes** — `_walk_urls` is generic within the four documents, not a fixed list. The last row is the documented `MethodNotAllowed` escape hatch working as intended, but it marks the boundary: the pin proves "not a 404, and not tenant-gated", not "a reader can actually use this". The document *list* is also fixed, so a fifth discovery document would join uncovered.

### Verified good

- **`x_for=0` left the rate limiter alone.** `tests/test_rate_limit_client_ip.py` + `tests/test_security_hardening.py` → **32 passed**. `_get_real_value` returns `None` at trust depth 0, so `REMOTE_ADDR` is never rewritten and `r6/rate_limit.py` keeps its own hop-counted parse.
- **The forwarded host is ignored**, by request, before and after my change.
- **Completeness.** I swept every `host_url` / `url_root` / `_external=True` site: 15 of them, all request-derived, all covered by the middleware. No remaining site builds a published URL from a hardcoded scheme. The three extra sites the PR names are real and are fixed by it.
- **`_disclaimer.url` is relative** (`"/r6/fhir/docs/privacy-policy"`), as the PR says and contrary to the issue text.

### Not covered

- The actual production defect. As the PR says, the confirming measurement is re-running the two production curls after deploy.
- The e2e Playwright suite.
- The Postgres CI lane, MCP server, CareAgents.
- `r6/wearables/routes.py:138` and `r6/command_center/routes.py:629` were reached only by inspection — the wearables path 503s without `OPEN_WEARABLES_URL`, and `desktop-demo` is not a public tenant on a fresh local instance, so `generate-link` demanded a step-up token.

### Housekeeping

- `dependency-audit` is red on every open PR from npm advisories already fixed in unmerged **PR #544**. Unrelated to this branch.
- **This takes effect only on deploy.** Everything above was measured against a locally simulated edge, which proves the middleware reads the header — not that the platform sends it in the form assumed here.

Counts on this branch after the fix: `uv run ruff check .` → `All checks passed!`; `uv run python -m pytest tests/ -q -p no:cacheprovider` → **3178 passed, 13 skipped, 1 xfailed** (3167 as submitted; +11 new tests). Ratchets and conformance are in that run and green; I touched neither `r6/routes.py`'s line count nor `tests/test_guardrail_conformance.py`.

aks129:
Independently measured before-state for the "close on the production curl" line, so the baseline is on record rather than reconstructed after the fix.

Contributed by a parallel session and **re-verified here on 2026-09-04** before posting:

```
GET /r6/fhir/.well-known/oauth-authorization-server
  issuer:                http://app.healthclaw.io
  authorization_endpoint: http://app.healthclaw.io/r6/fhir/oauth/authorize
  token_endpoint:         http://app.healthclaw.io/r6/fhir/oauth/token
```

**The issuer is the one that matters most, and #567 did not name it.** My own earlier sweep tested `/.well-known/oauth-authorization-server` at the root, got a 404, and moved on — the document is served under `/r6/fhir`. So the authorization-server metadata was never in the issue's site list.

That raises the severity of what this PR fixes. RFC 8414 requires the issuer to be an https URL, and an issuer is not merely a link: clients compare it for identity. A conformant client fetching this document today can reject us outright, and the value it would be comparing against is one we cannot serve.

`grep -rn "request.host_url" r6/ app.py main.py | wc -l` → **14**. One cause, and this PR corrects all of them at once, which is the argument its author made for the middleware over a helper: a helper is only ever as complete as the last person who remembered to call it, and this endpoint is the proof — it was missed by the issue, by my sweep, and would have been missed by a helper.

The confirming measurement after deploy is the same three requests, expecting `https://` in all three.

PR #584, opened 2026-09-04
candidate: every unanswered leaf reports, and the explanation is said once (review of #576)

aks129:
Fixes the two MEDIUMs from the QA review of #576. Neither was a leak, and the
review's zero-bits argument still holds after both fixes.

## What

**Finding 1: leaves that reported nothing at all.** Two shapes came back with
unanswered leaves and zero issues. Over HTTP the response carried no `issues`
parameter.

1. A `repeats: true` group whose leaves name a resource type with no resolver
   was not a list group. It fell through to ordinary recursion, and its leaves
   reached the "nothing was attempted" branch of `_resolve_answer`.
2. `_populate_list_group` walked a row's direct children only and tested
   `child.get("definition")`. A child that is itself a group carries none, so
   it was emitted bare and its own leaves were never visited: no answer, no
   issue, no item in the response.

"Attempted" is now one rule about questionnaire structure and nothing else: an
`initialExpression`, an `item.code`, or a `definition` on a leaf anywhere under
a repeating group. List rows resolve at every depth. Adding a fourth resource
type changes which leaves answer, never which leaves report.

**Finding 2: the explanation was repeated per leaf.** The 230-character
sentence rode on every unanswered leaf, so the response grew with the leaf
count times a fixed paragraph. Per the ruling, no cap. The reason is stated
once as an `informational` issue and each `incomplete` issue carries only the
linkId. Per-leaf entries stay greppable by linkId, which is what a caller
branches on.

The sentence is now 285 characters, because it gained a clause pointing at the
per-leaf issues. Said once, that costs 55 bytes; said per leaf it would have
cost more than the old one did.

## Why

The cause of finding 1 is the reason it matters. `_attempted_link_ids` in the
property test encoded the *ruling's prose* while the engine encoded a
*resource-type table*, so the pin agreed with the description and disagreed
with the code. It passed only because no fixture exercised the difference.
That is `docs/2026-08-02-retro.md` pattern 1 sitting inside the change whose
whole argument is that it removed one.

The test now derives "attempted" the way the engine does, and neither side
keeps a copy of the type table, so there is nothing left to diverge. The
literal `LIST_POPULATED_TYPES` that QA added in `49f3149` is deleted for that
reason, not ignored: the engine no longer has a type-dependent predicate to
mirror.

`_assert_zero_bits` also gains a structural mirror, because the biconditional
alone cannot see finding 1's second shape. A leaf absent from the response is
neither answered nor unanswered, so dropping one satisfies the property while
defeating its purpose. Every emitted repeat of a repeating group must now
carry exactly the leaves the questionnaire puts in it, at every depth.

The NKA exclusion is untouched. Both of its independent reasons are now
written beside it: the attestation booleans carry no `definition`, and they
are siblings of the repeating group rather than children of it.

## Base note

Stacks on #578, which stacks on #576, which stacks on #562. **Merge in that
order: #562, #576, #578, then this.** The base branch here is
`fix/populate-drop-subject-append` (#578).

On `dependency-audit`, which the review of #576 reported red on every open PR
from npm advisories fixed in unmerged #544: it does not appear in `gh pr
checks` for #576, #578, or this PR. None of the three touches an npm file, so
the job looks path-filtered and simply did not run. The advisories are real
and #544 is still unmerged; the "red here" part does not reproduce on this
stack. Every check that did run on this PR passes.

## The size measurements

Same K x R shape the review measured: a caller supplies K definition leaves in
one repeating group and R records, so K x R leaf occurrences report. Measured
with one script, run on the base commit and on this branch.

In process (issue bytes, and total response over request):

| K x R | request | issues before | before | issues after | after |
| --- | --- | --- | --- | --- | --- |
| 10 x 10 | 3.0KB | 31.9KB | 12.1x | 9.6KB | 4.7x |
| 50 x 200 | 29.3KB | 3191.5KB | 122.8x | 926.2KB | 45.5x |
| 200 x 500 | 85.7KB | 31977.6KB | 420.4x | 9321.7KB | 156.2x |
| 500 x 2000 | 293.0KB | 320097.7KB | 1232.1x | 93535.6KB | 458.8x |

Live over HTTP, 50 x 200 against a running app on port 5423:

```
before: 29.3KB request -> 3519.6KB response, 120.1x, sentence repeated 10000x
after:  29.3KB request -> 1254.3KB response,  42.8x, sentence appears once
```

The residual is the pre-existing K x R shape of the QuestionnaireResponse
itself, which this PR does not touch. At 500 x 2000 the QR alone is 40.9MB
against a 293.0KB request, or 139.5x. The issue list is now 2.3x the QR rather
than 7.8x, against the review's pre-#576 figure of 75x for this shape.

The issue channel is therefore no longer the dominant term, but the operation
is still worth a look. It is listed as an observation below rather than fixed
here, because bounding the QR is a different change with different callers.

## Live run

Real intake form, seeded synthetic tenant, port 5423. Full demographics, two
allergies with unlabelled SNOMED codes, one ICD-10 condition, one RxNorm
medication with no `dosageInstruction`.

```
answered leaves (11): 10 demographics + medications.item.name
                      + conditions.item.name
  (QA's live run reported 10 answered / 6 issues. The difference is the seed,
   not the behaviour: my condition carries ICD-10 E11.9, which
   r6/terminology.py labels, so conditions.item.name resolves.)
unanswered leaves (7): allergies.item.allergen x2, allergies.item.reaction x2,
                       medications.item.dose, and the 2 attestations
per-leaf issues (5): the 5 unanswered leaves that were attempted
                     (the 2 attestations are named in 0 issues)
issue codes: informational x1 (the reason), incomplete x5 (one linkId each)
severities: information only
'%patient projection' in the whole response: 1 occurrence (was 5)
rows: allergies 2, conditions 1, medications 1, against 2/1/1 stored
PHI markers in the issues parameter: none of 3
```

Both fixed shapes, same running app:

```
shape 1, repeats:true group of Immunization leaves
  before: {"item":[{"linkId":"imm.vaccine"},{"linkId":"imm.date"}], ...}
          issues parameter present: False
  after:  same response items, per-leaf issues: ['imm.vaccine', 'imm.date']

shape 2, nested group inside a repeating AllergyIntolerance group
  before: a.reactions emitted bare, 'a.reactions.text' absent from the response
  after:  a.reactions carries a.reactions.text; 2 rows x 2 leaves = 4 issues
```

## Mutation checks

Applied to the committed tree, run, reverted. Restored green at 56.

| mutation | red |
| --- | --- |
| M1 finding 1 shape 1 reverted | 2, both unsupported-type tests |
| M2 finding 1 shape 2 reverted | 3, all nested-group tests |
| M3 the sentence copied back onto every per-leaf issue | 6 |
| M4 the reason never stated at all | 3 |
| M5 a list-row leaf dropped instead of emitted | 4 |
| M6 the exclusion removed, so an item with no mechanism reports | 13, including both NKA guards |

The finding-2 bound is stated against the explanation itself rather than an
absolute byte count: one more unanswered leaf must cost less than one more
copy of the reason. Today that is 90.9 bytes against 285. Under M3 it is
377.9, which is what makes the test red.

The two finding-2 tests are complementary on purpose, and the seam is worth
naming. A short constant prefix per leaf, say `"not populated: " + linkId`,
would pass the byte bound. It is caught by the other test, which asserts each
per-leaf `diagnostics` equals the linkId exactly. One pins "only the linkId",
the other pins "no paragraph". Neither alone is enough.

M2 is the one that matters, and only the structural mirror catches it. Under
M2 the counted biconditional passes, because the dropped leaf is not in the
response to be counted. The mirror fails with the right diagnosis:

```
AssertionError: [nested-list / labelled, with a reaction] a repeat of
allergies.item does not mirror the questionnaire:
missing=['a.reactions.text'], unexpected=['a.reactions']
```

## Tests changed rather than added

Both are shape changes the ruling required, not reviewer tests broken.

- `test_a_repeating_group_of_an_unsupported_type_is_silent` is inverted and
  renamed to `..._reports_every_leaf`. QA's own docstring names this file as
  one of the two places to update if the reading is widened; the other was
  `LIST_POPULATED_TYPES`, now deleted.
- `_issue_link_ids` in `tests/test_sdc_populate_bounded.py` reads the
  `incomplete` issues and no longer splits `diagnostics` on a colon. A second
  helper, `_issue_explanation`, asserts there is exactly one `informational`
  issue and returns it, so the two HTTP tests that check the wording check it
  where it now lives.

## Ran

```
uv run ruff check .
  All checks passed!
  3198 passed, 13 skipped, 1 xfailed          (3193 before, 5 tests added)
  50 passed                                   (Grade A green, no pin moved)
```

Live against a running app on port 5423, seeded synthetic tenant: the intake
run, both fixed shapes, and the amplification case, all quoted above. Before
and after numbers come from the same two scripts, run on the base commit and
on this branch.

## What was NOT run

- Postgres CI lane. SQLite only.
- `$extract` against a form populated by this code. Nothing new enters the
  answer set, but the round trip is not exercised.
- Any renderer. QA established that nothing reads the issues; the PDF path and
  the intake UI were not looked at.
- The MCP `questionnaire_populate` caller against the new issue shape. It
  returns the whole `Parameters` body verbatim and is severity-blind, so the
  extra `informational` issue reaches a model's context as one more line. That
  is a decrease from one 250-character sentence per unanswered leaf, but it is
  reasoned rather than measured.
- Concurrency. Every size number here is a single request.
- The QuestionnaireResponse's own K x R growth, which is now the dominant term
  in the amplification and is pre-existing on this route. Worth filing.

Also left alone, from the review's own noted list: `r6/actions/review.py:208`
still does `content = [patient]` with an unredacted Patient and calls no
`apply_redaction`, so D10's bound is a property of the `$populate` route
rather than of the engine. `$extract`'s duplicate-Patient write is still
unfiled. Both are outside this PR and #578's body already names the first.

## Ruling item 4

#576's body says no deferral is written for the ruling's fourth item. One is:
#570 records the decision about whether `questionnaire_populate` stays in the
model-facing tier, and why the reason for removing it no longer holds. #576's
body is updated to cite it.

### DIFF
--- a/r6/sdc/expressions.py
+++ b/r6/sdc/expressions.py
@@ -151,7 +151,7 @@ def build_context(subject=None):
     return {'patient': projection, 'subject': projection}
 
 
-def evaluate(expression, resource, context=None, link_id=None):
+def evaluate(expression, resource, context=None, link_id=None, warned=None):
     """Evaluate a FHIRPath expression, returning a scalar, list, or None.
 
     Returns the single value when the result has one element, the list when
@@ -159,6 +159,16 @@ def evaluate(expression, resource, context=None, link_id=None):
 
     `link_id` names the questionnaire item, for the failure log only — see
     below for why it is what gets logged and the expression is not.
+
+    `warned` is an optional set of linkIds already logged, owned by the
+    caller and scoped to ONE REQUEST — populate_questionnaire builds it per
+    call. It exists because the same leaf is evaluated once per repeat
+    inside a list group, and a malformed expression would otherwise log a
+    line per row: the caller supplies the records through the inline
+    `content` Bundle, so log volume would scale with attacker-controlled
+    input. The evaluation itself is unchanged and still happens per row —
+    this is a logging concern only, and every line after the first says the
+    same thing about the same linkId.
     """
     if not expression:
         return None
@@ -175,6 +185,13 @@ def evaluate(expression, resource, context=None, link_id=None):
         # is caller-supplied too, and %r is what escapes control characters
         # in it. Pinned by tests/test_sdc_expressions.py::
         # test_a_failing_expression_is_not_echoed_into_the_log.
+        #
+        # ONCE PER REQUEST PER LINKID. `warned` is None for callers with no
+        # request scope (the direct unit tests), and those log every time.
+        if warned is not None:
+            if link_id in warned:
+                return None
+            warned.add(link_id)
         logger.warning('FHIRPath evaluation failed for item %r: %s',
                        link_id, type(exc).__name__)
         return None
--- a/r6/sdc/populate.py
+++ b/r6/sdc/populate.py
@@ -10,7 +10,16 @@ Pure function (no DB, no Flask). Supports three SDC population mechanisms:
   - List-resource-based: a `repeats: true` group whose leaves carry a
     `definition` naming a supported resource type (MedicationRequest /
     AllergyIntolerance / Condition) emits one repeat of the group per
-    matching resource for the subject (see _populate_list_group).
+    matching resource for the subject, resolving leaves at every nesting
+    depth inside the group (see _populate_list_group).
+
+WHAT COUNTS AS ATTEMPTED, since it is one rule and it is read in two places:
+an initialExpression, an `item.code`, or a `definition` on a leaf anywhere
+under a `repeats: true` group. The third clause does NOT depend on the type
+the definition names — a repeating group of Immunizations is an attempted
+population this file has no resolver for, so it answers nothing and reports
+every leaf, rather than going silent and letting a caller read the silence
+as "this patient has no immunizations".
 
 Out of scope (v1): StructureMap-based and CQL populate.
 
@@ -29,12 +38,23 @@ and test_zero_medications_never_infers_no_current_medications.
 
 from r6.sdc.expressions import build_context, evaluate
 
-#: What an attempted-but-unresolved leaf reports. ONE sentence for both
-#: causes, because the server does not tell the two apart and must not appear
-#: to: "the projection withheld it" and "the patient has no email address"
-#: read identically here on purpose. `questionnaire_populate` is in the
-#: model-facing read tier, and text naming a refusal would have a model tell
-#: a patient their record was held back when it is simply empty.
+#: Why attempted-but-unresolved leaves report, said ONCE PER RESPONSE. ONE
+#: sentence for both causes, because the server does not tell the two apart
+#: and must not appear to: "the projection withheld it" and "the patient has
+#: no email address" read identically here on purpose.
+#: `questionnaire_populate` is in the model-facing read tier, and text naming
+#: a refusal would have a model tell a patient their record was held back
+#: when it is simply empty.
+#:
+#: ONCE, not once per leaf. This constant is identical for every leaf of
+#: every response, so a copy on each issue was pure repetition: it made the
+#: response grow with the number of unanswered leaves TIMES a fixed
+#: paragraph, and a 29.3KB request came back 3519.6KB over HTTP. The issues
+#: now carry the only thing that varies — the linkId — and this sentence is
+#: emitted once beside them (r6/sdc/routes.py:_issues_outcome). The CTO
+#: ruling on the QA review of #576 chose this over a cap: a cap is a second
+#: control to reason about and would silently truncate a legitimate long
+#: form, while the amplification was entirely the repetition of a constant.
 #:
 #: What the caller gets instead is the allowlist, which is public — they can
 #: compare it against their own expression without the server answering a
@@ -46,8 +66,9 @@ from r6.sdc.expressions import build_context, evaluate
 #: it, and a `where(value='...')` clause is a place to park a literal that
 #: would then ride out on an OperationOutcome.
 NOT_POPULATED = (
-    'not populated: no value resolved. $populate reads only the %patient '
-    'projection (name, birthDate, gender, telecom phone/email, address '
+    'not populated: no value resolved. Each accompanying issue names one '
+    'such item by linkId. $populate reads only the %patient projection '
+    '(name, birthDate, gender, telecom phone/email, address '
     'line/city/state/postalCode) and coded content this server recognises; '
     'anything else is not read.'
 )
@@ -187,8 +208,11 @@ def populate_questionnaire(questionnaire, subject, content_resources):
         Condition rows the list groups match. NOT the subject: that arrives
         as `subject` above, and putting it here as well left an unredacted
         Patient reachable by anything walking this list (PR #562 review).
-    issues: list of {'linkId', 'detail'} for items that errored (not for
-        items that simply had no data).
+    issues: list of {'linkId'} — one entry per leaf OCCURRENCE where
+        population was attempted and no value resolved. The reason is the
+        same for all of them and is the module constant NOT_POPULATED, said
+        once per response by r6/sdc/routes.py:_issues_outcome rather than
+        copied onto each entry.
     """
     issues = []
     # The context carries a BOUNDED projection of `subject`, never the stored
@@ -203,7 +227,8 @@ def populate_questionnaire(questionnaire, subject, content_resources):
     answer_items = []
     for item in questionnaire.get("item", []):
         answer_items.extend(_populate_item(
-            item, subject, context, observations, issues, content_resources))
+            item, subject, context, observations, issues, content_resources,
+            in_repeating_group=False))
 
     qr = {
         "resourceType": "QuestionnaireResponse",
@@ -217,10 +242,16 @@ def populate_questionnaire(questionnaire, subject, content_resources):
     return qr, issues
 
 
-def _populate_item(item, subject, context, observations, issues, content_resources):
+def _populate_item(item, subject, context, observations, issues,
+                   content_resources, in_repeating_group):
     """Populate one questionnaire item. Always returns a list of zero or more
     QuestionnaireResponse items (zero or one for ordinary items; zero or many
     for a repeating list-resource group — one per matching resource).
+
+    `in_repeating_group` is True once ANY ancestor group carries
+    `repeats: true`. It is what makes a `definition` leaf an attempted
+    population — see _resolve_answer's last branch, which is where the rule
+    is written down.
     """
     link_id = item.get("linkId")
     item_type = item.get("type")
@@ -231,18 +262,22 @@ def _populate_item(item, subject, context, observations, issues, content_resourc
             return _populate_list_group(item, list_resource_type, subject,
                                         content_resources, issues)
         # Ordinary group: recurse, keep the group only if it produced
-        # child answers.
+        # child answers. A `repeats: true` group whose leaves name a type
+        # with no resolver lands here too — that is what the flag carries
+        # down, so its leaves still count as attempted.
+        nested = in_repeating_group or bool(item.get("repeats"))
         children = []
         for child in item.get("item", []):
             children.extend(_populate_item(
                 child, subject, context, observations, issues,
-                content_resources))
+                content_resources, nested))
         if not children:
             return []
         return [{"linkId": link_id, "item": children}]
 
     answer_value, value_key = _resolve_answer(
-        item, item_type, context, observations, issues, link_id)
+        item, item_type, context, observations, issues, link_id,
+        in_repeating_group)
     # Leaf items are always emitted so the response mirrors the questionnaire's
     # structure; the answer array is attached only when a value resolved.
     answer_item = {"linkId": link_id}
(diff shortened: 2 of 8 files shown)


aks129:
## QA: FAIL on #584 — one MEDIUM. Both claimed fixes are real and I verified them over HTTP; the rule they generalise still has a third shape.

**This is not a GitHub approval and must not be read as one.** `autoMergeRequest`
was `null` on #584, #578 and #588 before I started and `null` after my push. Do not
arm auto-merge. Merge order is unchanged: #562 → #576 → #578 → this.

**Read this first: this branch has never been tested by CI.** Six checks, none of
them a test — I confirmed it from the API rather than from the pull request page:
`GET /commits/6fe3ec0/check-runs` returns five (`claude-standards-review`,
`auto-merge-when-satisfied`, `scan / scan changed files` ×2, `Vercel Preview
Comments`) and `GET /commits/6fe3ec0/status` returns the sixth (`Vercel`). Zero
test jobs. My local run below is the only test evidence this change has, which is
what #588 fixes.

---

### Both fixes work, and I could not break them

**Finding 1, shape 1 — a repeating group of an unsupported type.** Live, seeded
synthetic tenant, `$populate` over HTTP:

```
repeats:true group, leaves definition=...#Immunization.vaccineCode.text
                                      ...#Immunization.occurrenceDateTime
-> 200
   unanswered leaves (2): imm.vaccine, imm.date
   informational issues: 1     incomplete issues (2): imm.vaccine, imm.date
```

Was two unanswered leaves, zero issues, no `issues` parameter at all.

**Finding 1, shape 2 — a nested group inside a repeating list group.** Two records,
the second unlabelled:

```
rows emitted: 2   records supplied: 2
answered  : a.allergen
unanswered: a.reactions.text, a.allergen, a.reactions.text
incomplete: a.reactions.text, a.allergen, a.reactions.text
unanswered leaves NOT named by an incomplete issue: []
```

**Finding 2 — said once.** At one issue and at one hundred, `%patient projection`
appears exactly once in the whole OperationOutcome. The 10×10 shape the review
measured came back 10,035 bytes; the old shape carried 100 copies of a 230-character
sentence. `severity` is `information` on every issue; the two kinds are told apart by
`code`, not by position, and each `incomplete.diagnostics` is a bare linkId that
names an unanswered leaf present in the response — so a caller still learns exactly
which fields were empty, per occurrence, with duplicates where a leaf repeats.

**The `if issues:` guard holds.** A form where every leaf answers carries no `issues`
parameter — not a lone `informational` with nothing under it, which would have made
the biconditional read backwards.

**The property, attacked at the new shapes.** `_assert_zero_bits` over the patient
matrix × content shapes on both new questionnaire shapes: no falsification. Row count
equals record count in 16 shapes (0/1/3/7 records × {no `code` key, `code: {}`,
unlabelled SNOMED, labelled}) and in 4 more on the nested shape, where each row also
produces exactly two issues.

**The two attestations are never reported.** The intake form over 3 patients × 3
content shapes: `allergies.no-known-allergies` and `medications.no-current-medications`
appear in zero issues and carry zero answers. In the newly covered shapes the
exclusion still works by mechanism rather than by name — an NKA-shaped boolean nested
inside a supported list group, and one inside an unsupported-type repeating group, are
both silent without a `definition` and both reported *with* one. Reported is correct:
a definition under a repeating group is an attempted population, which is the rule as
written.

**The counting trap reproduces exactly as described.** With `_populate_list_children`
reverted to the direct-children loop, on `_nested_list_questionnaire()`:

```
attempted            : ['a.allergen', 'a.reactions.text']
expected (counted)   : {}
actual   (counted)   : {}
COUNTED BICONDITIONAL: PASS
STRUCTURAL MIRROR    : FAIL  missing=['a.reactions.text'] unexpected=['a.reactions']
```

The counted half passes on both sides because a dropped leaf leaves both counters
empty. One correction to the claim: the mirror is what catches it *inside
`_assert_zero_bits`*, but the file as a whole is not blind without it —
`test_a_nested_group_inside_a_repeating_list_group_keeps_its_leaves` and
`test_a_nested_list_leaf_that_resolves_nothing_names_itself` also go red under the
same revert. The mirror is what makes the *property* mean something, which is the
part that matters.

**Test honesty.** `git checkout origin/fix/populate-drop-subject-append -- r6/sdc/`
→ **17 failed, 28 passed** across the four populate test files. Restored, green.
And two drop-mutations, to check the file is not satisfiable by deletion: unanswered
leaves vanishing from ordinary groups reddens 15 tests, and from inside a list group
reddens 6.

---

### MEDIUM — the rule is stated in three clauses and implemented in one, inside a supported list group

`r6/sdc/populate.py:_populate_list_children` (the leaf branch, `elif
child.get("definition")`) against `r6/sdc/populate.py:12-20` (the module docstring)
and `tests/test_populate_issue_property.py:_attempted_link_ids`.

`_populate_list_children` resolves a row's leaves by `definition` and by nothing
else. A leaf carrying an `initialExpression` or an `item.code` inside a
MedicationRequest / AllergyIntolerance / Condition group is therefore never evaluated
**and never reported**. Live, one allergy supplied:

```
POST /r6/fhir/Questionnaire/$populate
  allergies.item (repeats, AllergyIntolerance)
    a.allergen  definition=...#AllergyIntolerance.code.text
    a.expr      initialExpression = %patient.gender
-> 200
   answered  : a.allergen
   unanswered: a.expr
   issues parameter: ABSENT
```

Same with `a.code` carrying an `item.code`. That is the state the ruling set out to
remove — "a caller reads silence and concludes the patient has no such value" — in the
third shape of the same defect.

**It is not a leak and there is no live impact.** The silence is a function of the
questionnaire's structure, identical on every patient and every record (checked over
the patient matrix × three content shapes), and the intake form has no such leaf. I am
filing it at the same grade the review of #576 filed findings 1 and 2, for the same
reason.

**Why it is a finding and not a note.** This PR's argument is that the divergence is
gone: "Attempted is now one rule about questionnaire structure and nothing else",
"adding a fourth resource type changes which leaves answer, never which leaves
report", and `LIST_POPULATED_TYPES` was deleted because "the engine no longer has a
type-dependent predicate to mirror". The type table is gone; the divergence moved.
`_attempted_link_ids` has three clauses, `_populate_list_children` implements the
third alone, no fixture crosses the difference, and the whole suite is green — which
is the mechanism by which the first two shapes shipped, restated.

**Reproduction** (both tests are on the branch now, red):

```
uv run python -m pytest tests/test_populate_issue_property.py -q -p no:cacheprovider
  FAILED test_an_expression_or_code_leaf_inside_a_supported_list_group_reports
  FAILED test_the_property_holds_with_a_mechanism_leaf_inside_a_list_group
  2 failed, 17 passed
```

**What I did about it.** I fixed the test only — I write no production code.
`9882fe5` adds the two shapes to the property file and leaves them **red on purpose**,
because a green suite with a known hole is the state this PR's own body argues
against. I checked the pin is satisfiable rather than merely red: under a modelled
engine change (a leaf inside a row that carries an expression or a code resolves
through `_resolve_answer`), the two new tests and every test in
`test_populate_issue_property` / `test_populate_lists` / `test_sdc_populate` pass —
**0 red of 33**.

**The call is yours, and either direction closes it** — but both places move together:

- widen the engine so expression and code leaves resolve inside a list row (the rule
  as written), or
- narrow the rule to "`definition` alone applies inside a list group" and rewrite the
  module docstring, `_attempted_link_ids` and these two tests, since the first two
  clauses would then be conditional on where the leaf sits.

### LOW / noted, no action asked

- `r6/sdc/populate.py:_list_group_resource_type` still reads the repeating group's
  **direct** children only. Move every supported `definition` one level deeper and the
  group is not recognised as a list group at all: with one stored allergy whose
  `code.text` is `Penicillin`, the response is one row, the leaf reports an issue, and
  the stored value is simply lost. The property holds (unanswered, and named), so this
  is a functional limit rather than a silence. Pre-existing, and the PR's "every depth"
  claim is about resolution inside a detected group, not about detection.
- `_assert_zero_bits`'s structural mirror is scoped to repeating groups
  (`_repeating_group_leaves`), so the vacuity hole it closes stays open for ordinary
  groups. Measured: a mutation dropping unanswered leaves outside repeating groups
  reddens 15 tests in the file — but none of them is the property. Worth one line in
  the mirror's docstring saying what it does not cover.
- `test_the_response_grows_with_the_leaf_count_not_with_a_paragraph`'s docstring says
  "a ~253-character constant"; `len(NOT_POPULATED)` is 285. The assertion reads the
  constant, so only the prose is stale.
- The summary issue's `diagnostics` is the sentence, so a caller splitting
  `diagnostics` on `": "` for a linkId — the old wire shape — now gets
  `not populated` from it. `code` distinguishes them and there is still exactly one
  producer and no consumer (`r6/sdc/routes.py:90`; `grep` across `r6/`, `careagents/`,
  `services/`, `static/`, `templates/` finds nothing reading it), so this is a note
  about the shape, not a break.

---

### What I ran

```
  before my commit: 3198 passed, 13 skipped, 1 xfailed        in 127.25s
  after  my commit: 2 failed, 3198 passed, 13 skipped, 1 xfailed  in 119.57s
  (the 2 are the reproduction above; nothing else moved)
uv run ruff check .
  All checks passed!
  50 passed                       — Grade A green, no pin moved
```

Live: `uv run flask --app main init-db`, then the app on a free port with a seeded
synthetic tenant; `$populate` driven six ways (intake form; unsupported-type repeating
group; nested group in a list group with 1 and 2 records; a form where every leaf
answers; the expression-in-list-group shape; a 10×10 amplification shape). No PHI-ish
marker appears in any issues OperationOutcome. Audit `detail` is unchanged
(`populate; issues=N`, a count).

### What I did NOT cover

- **Postgres.** SQLite only, as always locally — and this branch has no CI, so the
  Postgres lane has never seen it.
- **`$extract` round-trip** against a form populated by this code, and the
  duplicate-Patient write the ruling describes.
- **Any renderer**: the PDF, the intake UI, mobile or an in-app browser. Nothing reads
  the issues today, so there is nothing to render, but I did not look.
- The MCP `questionnaire_populate` caller against the new two-kind issue list — it
  returns the `Parameters` body verbatim and is `severity`-blind, so I expect nothing,
  but I did not exercise it.
- `r6/actions/review.py:250` still discards the issue list (`qr, _issues = ...`), so the
  action rail cannot surface any of this. Unchanged by this PR, still worth filing.
- Concurrency, and anything above the 10×10 shape over HTTP.

**Verdict: FAIL** on the third silent shape. Nothing I found is a leak, and the
projection, the redaction, the unconditional emission and the say-it-once design are
all correct and I want them. It flips to PASS when the rule and the engine agree —
either direction — and the two tests on the branch go green.

aks129:
Folded #594's two commits onto this branch (head d4ea18c) after rebasing onto main. Reason: the test this PR adds (`test(populate): the third silent shape, in the branch the fix did not reach`) pins the mechanism-leaf-inside-a-list-group property that only #594 fixes, so on its own this branch fails two tests and can never go green. Together: 48 tests in the four populate files pass, ruff clean. #594's branch now points at the same head and will close when this merges.

PR #588, opened 2026-09-04
candidate: run the test pipeline on stacked pull requests (#585)

aks129:
Closes #585.

## What

`.github/workflows/ci.yml` triggered on `pull_request: branches: [main, master]`.
That filter matches the pull request's **base** branch, so a change stacked on
another feature branch matched nothing and the entire workflow — every test lane
in it — was skipped. The filter is removed. The `push` trigger's branch list is
unchanged.

Also adds `test_ci_runs_on_every_pull_request_including_stacked_ones` to
`tests/test_ci_hardening.py`, next to the other static workflow guards, so the
filter cannot come back unnoticed.

## Why

Measured across the open pull requests before this change:

| Base | Checks |
|---|---|
| `main` (#580) | **19** |
| stacked on a feature branch (#584, #578, #571 — identical) | **6** |

The six were `claude-standards-review`, `auto-merge-when-satisfied`,
`scan / scan changed files` ×2, and Vercel ×2. **None of the six is a test.**

`auto-merge-when-satisfied` is one of them. On this repository a maintainer
approval arms a squash merge, and with no test job present there is no failing
required check to hold it — so approving a stacked change merged it having run
zero tests, over a pull request page that looked clean. Seven stacked changes
are open in that state right now (#566, #569, #571, #576, #578, #579, #584).

## Before and after

Per-job, from the actual check lists:

| Job | `main` PR, before | stacked PR, before | stacked PR, after |
|---|---|---|---|
| `lint` | yes | **no** | **yes** |
| `python-tests` | yes | **no** | **yes** |
| `postgres-tests` | yes | **no** | **yes** |
| `node-tests` | yes | **no** | **yes** |
| `playwright-tests` | yes | **no** | **yes** |
| `compose-smoke` | yes | **no** | **yes** |
| `secret-scan` | yes | **no** | **yes** |
| `dependency-audit` | yes | **no** | **yes** |
| `compliance-gates` | yes | **no** | **yes** |
| `claude-standards-review` | yes | yes | yes |
| `auto-merge-when-satisfied` | yes | yes | yes |
| `scan / scan changed files` ×2 | yes | yes | yes |
| Vercel, Vercel Preview Comments | yes | yes | yes |
| `CodeQL`, `Analyze` ×3 | yes | **no** | **still no** — see below |

**6 → 15.** A pull request into `main` gets 19. The remaining difference of 4 is
code scanning, and this pull request cannot close it: CodeQL here is GitHub
[...]

PR #594, opened 2026-09-04
fix(populate): resolve every mechanism inside a list row, not just definition

aks129:
Closes the MEDIUM on the QA review of #584. Based on `fix/576-report-every-leaf-once`; the two tests QA left red at `9882fe5` are green, and nothing else moved.

## The finding

`_populate_list_children` resolved a row's leaves by `definition` alone and carried its own copy of the reporting decision (`elif child.get("definition")`). A leaf carrying an `initialExpression` or an `item.code` inside a MedicationRequest / AllergyIntolerance / Condition group was therefore neither evaluated nor reported — an unanswered leaf, no issue, and over HTTP no `issues` parameter naming it. That is the silence this branch exists to remove, in the third shape of the same defect.

## Why widen rather than narrow

QA left the direction open. The two resolutions are not symmetric, and the suite's own fixtures show why.

Take one leaf — `{"linkId": "a.expr", initialExpression: "%patient.gender"}` — and put it inside a `repeats: true` group. Under `_unsupported_list_questionnaire()` (Immunization) it **is** evaluated and reported: `_list_group_resource_type` returns None, the group falls through to ordinary recursion, and the leaf reaches `_resolve_answer`. Under `_list_group_with(...)` (AllergyIntolerance) it was neither.

So "`definition` alone applies inside a list group" is not a rule about **where** the leaf sits. It is a rule conditional on `_LIST_RESOURCE_CONFIG` membership — the type table back in the predicate that `3be4c8a` deleted it from, and the direct contradiction of this PR's own claim that "adding a fourth resource type changes which leaves answer, never which leaves report." Narrowing would make adding a fourth type **silence leaves that report today**. Widening is the only direction under which the rule stays independent of the type table.

## The change

`r6/sdc/populate.py:_populate_list_children` — a row's leaf takes this row's record value when the definition names an element of it, and otherwise goes to `_resolve_answer` with `in_repeating_group=True`. The local reporting branch is deleted. The engine now has **one** implementation of "attempted" rather than two, which is the fix; the two tests going green is the symptom clearing.

`context` and `observations` thread down through `_populate_list_group`. No callers outside the module.

Precedence, stated rather than handled: a leaf carrying both a matching `definition` and an `initialExpression` resolves record-first and falls back to the expression. That is author-pathological, no fixture or intake leaf has one, and no code was written for it.

## The constraint the property test could not see

A mechanism-less boolean inside a list row now reaches `_resolve_answer`'s "nothing was attempted" branch by a route it never took before — and that branch is what keeps `allergies.no-known-allergies` out of the issue list. The real intake form puts the attestations *beside* the repeating group, never inside it, so no fixture exercised it and `_assert_zero_bits`'s `unattempted_named` assertion had nothing to catch.

`tests/test_populate_lists.py::test_a_mechanism_less_boolean_inside_a_list_row_is_neither_answered_nor_reported` puts one there. Mutation-checked: replacing the `_resolve_answer` handoff with an unconditional `_report_unpopulated` reddens that test **and nothing else** — `test_populate_issue_property.py` stays fully green under it, which is exactly the gap it fills.

## Live, over HTTP

`uv run flask --app main init-db`, app on **port 60662** (picked by binding `("127.0.0.1", 0)` and letting the kernel choose), synthetic tenant `populate-live-584`, seeded demo data plus two synthetic AllergyIntolerance rows (one `code.text`, one SNOMED-coded and therefore unlabelled after `apply_redaction`).

**The newly covered shape, before and after, same app and same request** — an AllergyIntolerance group holding a `definition` leaf, an allowlisted expression, a withheld expression, an `item.code`, and a bare boolean:

| | before (`9882fe5`) | after |
|---|---|---|
| answered | 0 | 2 (`a.expr` ×2) |
| unanswered | 10 | 8 |
| `incomplete` issues | 2 | 6 |
| unanswered and unreported | `a.expr`, `a.withheld`, `a.code`, `a.no-known-allergies` | `a.no-known-allergies` only |
| rows / records | 2 / 2 | 2 / 2 |
| `%patient projection` in the response | 1 | 1 |

The one leaf still silent after the change is the mechanism-less boolean, which is the design.

**The intake form**, 2 allergies + 1 medication + 1 condition on file:

```
answered  (11)  unanswered (7)
issues: 1 informational + 5 incomplete
unanswered leaves NOT named by an incomplete issue:
  ['allergies.no-known-allergies', 'medications.no-current-medications']
rows[allergies.item]=2   records=2   OK
rows[medications.item]=1 records=1   OK
rows[conditions.item]=1  records=1   OK
'%patient projection' occurrences in the whole response: 1
allergies.no-known-allergies:    in issues=False  answered=False
medications.no-current-medications: in issues=False  answered=False
```

Both allergy rows are unanswered because `apply_redaction` strips the upstream `code.text` and `r6/terminology.py` has no SNOMED entries — today's real behaviour on a MEDENT-shaped import, and the row-count invariant holding through it.

## The amplification this would have moved, closed rather than disclosed

Resolving expression leaves inside a list row means **evaluating** them once
per row — that is deliberate, and it is what makes the leaf answer at all. The
failure **log** was the problem: the line is identical for every row (a linkId
and an exception class name), so one malformed expression logged one warning
per record.

| records | beside the row | inside the row, before | inside the row, after |
|---|---|---|---|
| 1 | 1 | 1 | 1 |
| 10 | 1 | 10 | 1 |
| 100 | 1 | 100 | 1 |

Worth closing even though it is only log volume: the record count comes from
the caller's own inline `content` Bundle, so an authenticated caller can
multiply our log lines by supplying more records. Cost that scales with
attacker-controlled input is worth removing while it is one small change
rather than an incident.

**Deduped at the log site**, not by evaluating once per group. `evaluate` takes
a caller-owned set of linkIds already warned (`r6/sdc/expressions.py:154`), and
`populate_questionnaire` builds one per call, threaded exactly like `issues`
beside it. Evaluating once per group would have produced the same number and
was rejected: it would rebuild a second resolution path beside the one
`_populate_list_children` was just collapsed into, which is the divergence this
branch exists to remove.

Resolution is untouched, and
`tests/test_sdc_expressions.py::test_a_failing_expression_in_a_list_row_logs_once_not_once_per_record`
asserts that rather than assuming it: every row still evaluates, every row
still emits its own issue, and rows still equal records. Mutation-checked —
dropping the guard reddens it at 10 records. Both live drives were re-run over
HTTP after the change and came back identical to the run before it.

Not a leak either way: `r6/sdc/expressions.py` logs the linkId and the
exception *type name* only, never the expression or `str(exc)`.

## Gates

```
  3202 passed, 13 skipped, 1 xfailed in 118.20s
  (3198 + QA's 2, now green + 2 new pins)
uv run ruff check .
  All checks passed!
  50 passed          — Grade A, no pin moved
```

## Not touched

The three LOW items on the QA review are untouched and still stand: `_list_group_resource_type` reading direct children only, the structural mirror's scope being repeating groups, and the stale "~253-character" docstring. Also unaddressed and unchanged: `r6/actions/review.py:250` still discards the issue list, `$extract` round-trip, and the Postgres lane (this stack is SQLite locally, and this branch chain still has no test job in CI until #588).

Do not merge or approve — merge order is unchanged: #562 → #576 → #578 → #584 → this.

### DIFF
--- a/r6/sdc/expressions.py
+++ b/r6/sdc/expressions.py
@@ -151,7 +151,7 @@ def build_context(subject=None):
     return {'patient': projection, 'subject': projection}
 
 
-def evaluate(expression, resource, context=None, link_id=None):
+def evaluate(expression, resource, context=None, link_id=None, warned=None):
     """Evaluate a FHIRPath expression, returning a scalar, list, or None.
 
     Returns the single value when the result has one element, the list when
@@ -159,6 +159,16 @@ def evaluate(expression, resource, context=None, link_id=None):
 
     `link_id` names the questionnaire item, for the failure log only — see
     below for why it is what gets logged and the expression is not.
+
+    `warned` is an optional set of linkIds already logged, owned by the
+    caller and scoped to ONE REQUEST — populate_questionnaire builds it per
+    call. It exists because the same leaf is evaluated once per repeat
+    inside a list group, and a malformed expression would otherwise log a
+    line per row: the caller supplies the records through the inline
+    `content` Bundle, so log volume would scale with attacker-controlled
+    input. The evaluation itself is unchanged and still happens per row —
+    this is a logging concern only, and every line after the first says the
+    same thing about the same linkId.
     """
     if not expression:
         return None
@@ -175,6 +185,13 @@ def evaluate(expression, resource, context=None, link_id=None):
         # is caller-supplied too, and %r is what escapes control characters
         # in it. Pinned by tests/test_sdc_expressions.py::
         # test_a_failing_expression_is_not_echoed_into_the_log.
+        #
+        # ONCE PER REQUEST PER LINKID. `warned` is None for callers with no
+        # request scope (the direct unit tests), and those log every time.
+        if warned is not None:
+            if link_id in warned:
+                return None
+            warned.add(link_id)
         logger.warning('FHIRPath evaluation failed for item %r: %s',
                        link_id, type(exc).__name__)
         return None
--- a/r6/sdc/populate.py
+++ b/r6/sdc/populate.py
@@ -225,6 +225,13 @@ def populate_questionnaire(questionnaire, subject, content_resources):
         copied onto each entry.
     """
     issues = []
+    # Per-request, and threaded exactly like `issues` beside it: the set of
+    # linkIds whose expression has already logged a failure. A leaf inside a
+    # list group is evaluated once per row, so without this a malformed
+    # expression logs a line per record — and the records come from the
+    # caller's own inline `content` Bundle. See r6/sdc/expressions.py's
+    # `warned`. It is NOT returned: nothing outside this call reads it.
+    warned = set()
     # The context carries a BOUNDED projection of `subject`, never the stored
     # Patient (r6/sdc/expressions.py). `subject` itself stays whole here
     # because the list-group and reference paths below match on it; only the
@@ -238,7 +245,7 @@ def populate_questionnaire(questionnaire, subject, content_resources):
     for item in questionnaire.get("item", []):
         answer_items.extend(_populate_item(
             item, subject, context, observations, issues, content_resources,
-            in_repeating_group=False))
+            warned, in_repeating_group=False))
 
     qr = {
         "resourceType": "QuestionnaireResponse",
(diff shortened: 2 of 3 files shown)


aks129:
Carried by #584: its branch now includes these two commits on top of the rebased #584 head (d4ea18c), because #584's own test needs this fix to pass. This PR will show no diff once #584 merges and can be closed then.

PR #595, opened 2026-09-04
feat(audit): bring agent_runs under the audit rule, and strengthen the pin (B2)

aks129:
Ratchet slice B2 (council ruling D11, parallel lane). Touches nothing else in flight.

## The gap

`r6/agent_runs` was the last package in the codebase that mutated state and audited none of it. Fourteen routes, ten of them authenticated by a shared secret rather than step-up, every one writing to the store, none recording that it had. An agent could register a tool call, execute a side effect, and persist an assistant answer with nothing in the audit trail to say it happened.

## What is audited, and what is not

Eight routes audit; six are classified as timer chatter with a stated reason. The line: **a mutation a principal asked for is audited; a mutation the system makes to itself on a timer is not.** The run's own append-only `AgentRunEvent` log already holds the operational story — deadline sweeps, lease recovery, claim redelivery, heartbeats and the event log itself stay out, because putting them in the audit trail would bury the records an auditor needs under queue chatter, which is how an audit trail becomes unreadable.

| Route | Audited | Why |
|---|---|---|
| `POST /runs` | `create AgentRun` | a principal enqueues agent work |
| `POST /runs/<id>/cancel` | `update AgentRun` | a person stops an agent |
| `POST /runs/<id>/transition` | `update AgentRun` | once-per-attempt outcome — including entering the human gate |
| `POST /runs/<id>/resume` | `update AgentRun` | the other half of the human gate |
| `POST /runs/<id>/tool-calls` | `create AgentToolCall` | the moment an agent commits to a real-world side effect |
| `POST .../tool-calls/<id>/transition` | `update AgentToolCall` | the outcome of that side effect |
| `POST /runs/<id>/finalize` | `create ConversationMessage` + `update AgentRun` | the assistant's answer, mirroring the command centre's audit of the user's turn |
| `POST .../tool-calls/<id>/reconcile` | `update AgentToolCall` | a human asserting what happened in the world |
| `POST /runs/claim`, `POST /runs/<id>/heartbeat` | no | polled continuously by every live worker |
| `POST /runs/<id>/events` | no | the run's own log; auditing would copy one log into another per streamed chunk |
| the three GETs | no | reads; their only write is the shared deadline sweep, which no principal requested |

The finalize case is the one that matters most: `command_center` audits the user's turn, so auditing the run's assistant turn is what stops the trail being able to show a question and never its answer.

## Two judgment calls, stated

**Audit lives in the service layer, not the routes.** Each of these functions owns its own transaction. An audit written at the route would commit separately from the change it describes — fail-open on the data, fail-closed on the response, the exact shape `record_audit_event` has. Every call site sits after the last raise and before the one commit, so a refusal leaves nothing pending.

**It adopts the kernel's `audit()` rather than `add_audit_event` directly.** This package has no prior audit call sites, so it can land on the spec's one audit call directly rather than adopting a primitive slice 13 exists to retire. It also sets the `g` marker `install_read_audit_assertion` needs. Only `audit` is imported — the step-up gate here still calls the validator directly and is still counted by `_STEP_UP_CALLSITES`, because one guard per PR is the protocol.

## The pin needed strengthening, not just lowering

`_UNAUDITED_MUTATOR_PACKAGES` measured two things its name did not claim:

1. **Presence per package, not per mutation.** One `audit()` call anywhere in `r6/agent_runs` would have emptied the set while thirteen endpoints stayed silent.
2. **Its own subject was mostly out of scope.** It qualified a package as a mutator only if the package validated step-up. Ten of these fourteen routes gate on a shared secret, and the other four reach step-up through a single call site inside `_tenant_authorized`. Verified by mutation: with step-up removed from this package, the old predicate does not flag it at all — it caught this one by accident, and would have missed a package that used no step-up.

So the predicate now also counts a package that commits (if it commits, it mutates), with a floor so a predicate that matches nothing cannot pass. Measured before widening: the commit predicate adds `r6/fasten` to the mutator set and **nothing** to the silent set — inert today, a trap laid for tomorrow. The set goes to empty and becomes a tripwire (playbook A7's shape).

`tests/test_agent_run_writes_are_audited.py` is the per-mutation pin the ratchet structurally cannot be. It classifies all fourteen routes as a set equality, so route fifteen is red until somebody decides which half it belongs to; proves each audited route at the wire; proves the exempt ones emit nothing, so the classification is false somewhere and is therefore a decision; and proves a refused write, an idempotent replay, and a cancel of an already-finished run leave no evidence of a write that did not happen.

## PHI

`detail` carries ids, states, counts, a regex-constrained tool name and a 12-character digest. Never a message body, a tool argument, or a result. `outcome_ref` is a free-form 256-character worker string and is deliberately absent; `evidence_ref` and `checkpoint_id` are `_ID`-constrained at the route, so they are opaque handles rather than prose. Two tests drive a distinctive sentence through `finalize` and through tool arguments and results, then assert no token of it reaches the trail.

## Gates

- `uv run python -m pytest tests/ -q` — **3174 passed, 13 skipped, 1 xfailed**
- `uv run ruff check .` — **All checks passed**
- Ratchets pass; the mutator tripwire verified to fire (neutralise the audit calls → red)
- Conformance Grade A: `tests/test_guardrail_conformance.py` — 37 passed

Do not merge or approve — maintainer gate.

### DIFF
--- a/docs/2026-08-05-healthclaw-2.0-playbook.md
+++ b/docs/2026-08-05-healthclaw-2.0-playbook.md
@@ -71,7 +71,7 @@ names its ratchet and its test-first artifact.
 | # | Chunk | Ratchet | Test-first |
 |---|---|---|---|
 | B1 | Audit for `command_center` (3 gated writes, currently 0 events) | coverage gap −1 | audit-assertion (#338 after-flush) turned on for these routes |
-| B2 | Audit for `agent_runs` (2 gated writes) | coverage gap −2 | same |
+| B2 | Audit for `agent_runs` — **shipped 2026-09-04**. "2 gated writes" was an undercount: the package has 14 routes and 10 of them are gated on a shared secret rather than step-up, which is also why the ratchet's step-up-only predicate barely covered its own subject. 8 routes audit, 6 are classified as timer chatter with reasons | coverage gap −2; unaudited-mutator set → empty (tripwire) | `tests/test_agent_run_writes_are_audited.py` — per-route classification, wire proof, PHI-free detail |
 | B3–B8 | `record_audit_event` → `add_audit_event`, one blueprint per PR (routes, curatr, labs, caregaps, shc/quality, actions/review) | 88→0 stepwise | audit-atomicity matrix: for each write path, assert audit row in `session.new` before commit |
 | B9 | Delete `record_audit_event` + its ambient commit | primitive count 2→1 | the ratchet at 0 is the precondition |
 | B10 | Enable `HC_ASSERT_AUDIT_COMMITTED` in production once B9 lands | tripwire live | prod_watch check |
--- a/r6/agent_runs/service.py
+++ b/r6/agent_runs/service.py
@@ -9,6 +9,7 @@ from datetime import timedelta, timezone
 from sqlalchemy.exc import IntegrityError
 
 from models import db
+from r6.access import audit
 from r6.agent_runs.models import (
     AgentRun,
     AgentRunEvent,
@@ -23,6 +24,76 @@ from r6.agent_runs.state import (
     require_tool_transition,
 )
 from r6.command_center.models import Conversation, ConversationMessage
+
+
+# ---------------------------------------------------------------------------
+# Audit (playbook B2)
+# ---------------------------------------------------------------------------
+#
+# Which mutations record evidence is a decision, and it is made here rather
+# than at the routes: every one of these functions owns its own transaction,
+# so an audit written anywhere else would commit separately from the change it
+# describes — fail-open on the data, fail-closed on the response. The kernel's
+# audit() flushes inside THIS transaction and never commits, so the row and
+# the state change resolve together.
+#
+# The line: a mutation a principal asked for is audited; a mutation the system
+# makes to itself on a timer is not. Deadline sweeps, lease recovery, claim
+# redelivery, heartbeats and the run's own event log stay out — they are
+# already fully recorded in AgentRunEvent, and putting them here would bury
+# the records an auditor needs under queue chatter.
+# tests/test_agent_run_writes_are_audited.py classifies all fourteen routes
+# and proves each audited one at the wire.
+
+def _detail(**fields) -> str:
+    """A PHI-FREE `key=value` detail line, in the command centre's shape.
+
+    Only ids, states, counts and regex-constrained identifiers may be passed.
+    Never a message body, a tool argument, or a tool result: this package
+    handles the assistant's answer and the arguments a model chose, and both
+    are the most sensitive field in their request. Fields that are None are
+    dropped rather than rendered, so a detail says nothing it does not know.
+    """
+    return " ".join(f"{key}={value}" for key, value in fields.items()
+                    if value is not None)
+
+
+def _audit_run_change(run: AgentRun, previous_status: str, **extra) -> None:
+    """Record one principal-requested change to a run's state."""
+    audit(
+        tenant=run.tenant_id,
+        event_type="update",
+        resource_type="AgentRun",
+        resource_id=run.id,
+        agent_id=run.agent_id,
+        detail=_detail(**{"from": previous_status, "status": run.status,
+                          "attempt": run.attempt,
+                          "error_class": run.error_class, **extra}),
+    )
+
+
+def _audit_tool_change(
+    run: AgentRun,
+    call: AgentToolCall,
+    event_type: str,
+    **extra,
+) -> None:
+    """Record one change to a tool call — the agent's real-world side effect.
+
+    `tool_name` is `_ID`-constrained at the route and `input_hash` is a digest,
+    so both are safe to name. `outcome_ref` is a free-form 256-character worker
+    string and is deliberately absent.
+    """
+    audit(
+        tenant=run.tenant_id,
+        event_type=event_type,
+        resource_type="AgentToolCall",
+        resource_id=call.id,
+        agent_id=run.agent_id,
+        detail=_detail(**{"run": run.id, "tool": call.tool_name,
+                          "status": call.status, "attempt": call.attempt,
+                          "error_class": call.error_class, **extra}),
+    )
 
 
 def create_run(
@@ -57,6 +128,19 @@ def create_run(
     try:
         db.session.flush()
         append_event(run, "run.queued", {"status": "queued"})
+        # Inside the try: a concurrent creation rolls this transaction back,
+        # and the audit row goes with it. The replay above returns before
+        # here, so a run that already existed is never recorded as created.
+        audit(
+            tenant=run.tenant_id,
+            event_type="create",
+            resource_type="AgentRun",
+            resource_id=run.id,
+            agent_id=run.agent_id,
+            detail=_detail(conversation=run.conversation_id,
+                           message=run.message_id, surface=run.surface,
+                           deadline_s=deadline_seconds),
+        )
         db.session.commit()
     except IntegrityError:
         db.session.rollback()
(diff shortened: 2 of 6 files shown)


aks129:
## QA review: FAIL as received, PASS as the branch now stands

The production change is sound. I found no PHI leak, no guardrail bypass, and no
path that writes an audit for a write that did not happen. The defects are in
the guard, which is the part this PR asks to be judged on.

Three claims the branch makes in a docstring stay green when falsified. I fixed
all three and pushed them as `e7f21d6`; nothing under `r6/` was touched. Re-read
that commit before merging, because it is now part of what a maintainer approves.

### Gates

| Gate | Result |
|---|---|
| `uv run python -m pytest tests/ -q` (at `3e1318a`) | 3174 passed, 13 skipped, 1 xfailed in 114.03s |
| `uv run python -m pytest tests/ -q` (at `e7f21d6`) | 3175 passed, 13 skipped, 1 xfailed in 104.35s |
| `uv run ruff check .` | All checks passed |
| `tests/test_ratchets.py` | 13 passed |
| `tests/test_guardrail_conformance.py` | 37 passed (Grade A holds) |
| Live app, `PORT=5099`, real SQLite | 7 audit rows, 0 PHI tokens |

### 1. The widening was not pinned. HIGH, `tests/test_ratchets.py:271`

The whole argument for taking the set to empty is that the predicate filling it
got wider. That widening is reversible green.

```
sed -i "s/, 'commit')/)/" tests/test_ratchets.py
uv run python -m pytest tests/test_ratchets.py::test_no_new_package_mutates_without_auditing -q
# 1 passed
```

The `len(gates) > 5` floor added at line 278 is described as guarding "exactly
the edit that could break it silently." It does not. Dropping `'commit'` takes
the mutator set from 8 packages to 7, which still clears 5. A future PR can
revert the strengthening, keep the empty set, and pass every gate.

Fixed by pinning the property that motivated the widening: the commit half must
still be reaching a package the step-up half does not. If that ever goes empty
the widening has become redundant, and deleting it is then a deliberate edit at
the assertion rather than a silent one at the tuple.

### 2. The stated mutation is the wrong one. MEDIUM, `tests/test_ratchets.py:252`

"MUTATION: delete the audit import from a gated blueprint -> red" is false. The
scan walks the AST for call names.

```
sed -i 's/^from r6.access import audit$/# gone/' r6/agent_runs/service.py
uv run python -m pytest tests/test_ratchets.py::test_no_new_package_mutates_without_auditing -q
# 1 passed
```

Renaming the calls is what goes red, which is what the PR description says was
verified. The docstring predates this PR but sits on a function this PR
rewrites, and it tells the next reader to run a mutation that proves nothing.
Corrected to the mutation that holds.

### 3. The refusal test does not exercise its own mutation. MEDIUM, `tests/test_agent_run_writes_are_audited.py:533`

Both refusals in that test are raised before an audit could be reached. The 401
never enters the service, and the wrong-worker 409 raises inside
`lock_owned_run`, where there is no `run` to audit yet. So the stated mutation,
"audit before the fence check in transition_owned_run," cannot be caught here.
I confirmed it: moving `_audit_run_change` above the fence leaves this test
green and is caught by the row count in `test_a_worker_transition_is_audited`
instead.

Fixed by adding a right-worker `status: "completed"` request. It passes the
fence and is refused on the next line, so an audit placed there does run. That
case also exercises the fail-closed half this test claims to cover: the flushed
row behind a 409 trips `install_audit_assertions` at teardown. Verified red
under the mutation, green without it.

### A shape the tripwire still misses

The docstring is honest that this is a presence check per package. It does not
say the predicate has a second hole: a package can mutate without calling
`commit` at all, because a commit-performing helper already exists in another
package.

I built the smallest realistic instance. A new `r6/qa_probe/` gated on a shared
secret, in the same shape as ten of these fourteen routes, writing through
`persist_intake_document` (`r6/sdc/documents.py:44`, which does the add and the
commit), auditing nothing. All 13 ratchets pass. A new top-level `r6/foo.py`
that commits directly also passes, because every file directly under `r6/`
collapses into one bucket that `r6/routes.py` already audits for. At file
granularity, nine r6 modules commit or gate and never audit today, hidden by
that bucketing, one of which is `r6/audit.py` itself and so does not count.

Neither is a defect in this PR. Both bound the claim in the docstring that "a
new silent mutator goes red on arrival," which is true only for a new
subdirectory that also commits inside itself. Worth one sentence there. The scan
also skips everything outside `r6/`, so `careagents/`, `services/` and `api/`
are outside the tripwire entirely.

### Two calls for the CTO, not blockers

**The exemption is per code path, not per route.** The table classifies routes
by their happy path. The deadline sweep runs inside `_enforce_worker_fence`
(`r6/agent_runs/service.py:278,282`), which commits and then raises, so
`/transition`, `/heartbeat`, `/tool-calls`, `/finalize` and `POST /events` each
have a path that commits a state change and answers 409 with no audit. On a run
holding a running tool call that state change is entry into the human gate: the
run to `waiting_for_human`, its tool calls to `needs_reconciliation`. That is
the same state `POST /transition` reaches, where it is audited.

The cost is one asymmetry. A `reconcile` row, which the PR calls the most
evidence-worthy mutation in the package, can stand in the trail with no record
of how its call became ambiguous. I think the line still holds, since a timer is
not a principal and `AgentRunEvent` has the story. I added a test that asserts
the write really happens and that nothing is audited for it, so it is a recorded
decision. Ruling welcome.

**Three identifier fields are argued rather than constrained.** `evidence_ref`,
`tool_name` and `input_sha256[:12]` are in `detail`. The PR calls them opaque
handles because `_ID` matches them, but `_ID` constrains charset, not semantics:
`MRN:12345` matches. `outcome_ref` was excluded for being free-form, and
`reconcile` stores `evidence_ref` into that same column. The exposure is not new. All three already
sit in `AgentRunEvent` and on the tool-call row, and an operator typing an MRN
into an evidence field is a training problem before it is a code problem.
Flagging it so the ruling is on the record.

### Spec compliance

Council ruling D11 is not in the repo. `grep -rn "D11" docs/` returns nothing, so
I could not check the change against the ruling it cites. I checked it against
the task brief and the playbook row instead. That row said "2 gated writes,
coverage gap −2"; this PR audits eight routes and rewrites the row to match its
own scope. The brief sanctioned judging the classification and the predicate
change, so the extra six are argued scope rather than an unrequested extra. If
D11 constrains which routes, someone who can read it should confirm.

### What I exercised

- Read every one of the fourteen routes, the eight audited service functions,
  and the fence and sweep helpers they share.
- Traced the audit-before-commit ordering on every audited path, including both
  replay returns, both `IntegrityError` rollbacks, and every raise after the
  lock.
- Drove a real server on `:5099` against a real database: message, create,
  claim, tool register, two tool transitions, finalize, plus a 401 and a 409.
  Seven audit rows, all PHI-free, and the refusals added nothing. The user turn
  and the assistant turn both appear, which is the "question and its answer"
  claim holding at the wire and not only in a fixture.
- Four mutations on the guard and two on the service, each reported above with
  its command.

### What I did not cover

- **The Postgres lane.** No Docker and no local Postgres on this machine. Note
  that `tests/test_agent_run_writes_are_audited.py` is not in the
  `postgres-tests` list in `ci.yml`, though its sibling `tests/test_agent_runs.py`
  is. `with_for_update(skip_locked=True)` is a no-op on SQLite and this PR
  changes when locks are released by moving five call sites to `commit=False`.
  B1's equivalent file is not in that lane either, so this is consistent rather
  than a regression. Adding it is one line, and I did not add it because I
  cannot run the lane to prove it passes there.
- **Concurrency.** No two-writer test. The `commit=False` refactor widens the
  window in which `transition_owned_run` holds its row locks.
- **The `request_cancel` running-path audit.** The comment says a repeat cancel
  there audits again, correctly, because it writes again. No test covers it.
- **CareAgents.** Untouched by this PR and not exercised.

Not approving on GitHub, per instruction.

PR #601, opened 2026-09-04
test(connectors): the "2 of 4 proven live" claim, re-run by someone else (#530)

### LINKED ISSUE #530: "2 of 4 connectors proven live" rests on scripts nobody else can run
The topology's headline, PRD 02's measured state line, and hard-truths §1's table all assert flatly that **two of four connector kinds are proven live**. It is the single most load-bearing measured claim in the merged process documents.

The set-2 evidence pack's own register entry R8 says: *"There is no `hapi` or `generic` walkthrough in the repo. No issue yet."* The scripts that produced those two runs, and their raw output, live only in an uncommitted local scratch directory.

So the claim rests on a run **nobody but its author can re-execute**, sitting in the document whose entire thesis is that unreproducible assertions are the defect. The three summary documents repeat it without that caveat.

Raised by the review board on #519 as the claim across all four PRs most likely to be false — not because anything suggests it is, but because nothing can currently show it is not.

What was confirmed: both upstreams are real and current, fetched live from their `metadata` endpoints (HAPI FHIR 8.11.16-SNAPSHOT, Firely Server 6.9.1), and the pack's transcripts are internally consistent and self-implicating.

## To settle it — minutes, not a project

1. Commit `walkthrough-hapi.sh` (and the generic variant) alongside the existing `scripts/walkthrough.sh`
2. Commit the raw run transcripts
3. Have someone who is not the author re-run both against `hapi.fhir.org` and `server.fire.ly`, and diff

Until step 3, the three documents should say *proven once, by its author, from an uncommitted script* — or this closes.

aks129:
Closes #530.

## The finding

**The claim reproduces.** Two of four connector kinds are proven live, and as
of this PR someone other than the original author has shown it, from a script
anyone can run.

Three merged documents assert "2 of 4 proven live". The 2026-08-16 run behind
that claim came from a re-pointed copy of the Aidbox walkthrough living in an
uncommitted scratch directory — the evidence pack says so itself, in R8 — so
the most load-bearing measured claim in the process documents sat in the
document whose thesis is that unreproducible assertions are the defect.

The scratch directory is gone. Nothing was recovered from it. The script here
was written fresh from the pack's transcripts, which is itself the check on
whether those transcripts describe something reproducible. They do.

## What was run

`scripts/walkthrough-upstream.sh hapi` and `… generic`, against
`hapi.fhir.org/baseR4` and `server.fire.ly/R4`, on 2026-09-04, at `89b42fb`.
Both servers answered (0.98s and 7.46s on `/metadata`), so neither run has an
outage to discount. One run per server; no retries, no load.

| Step | `hapi` 08-16 | `hapi` today | `generic` 08-16 | `generic` today |
|---|---|---|---|---|
| 0 preflight — mode, kind, connected | PASS | PASS | PASS | PASS |
| 1a create reached upstream | PASS | PASS | PASS | PASS |
| 1b gate matrix 428/401/428/201 | PASS x4 | PASS x4 | PASS x4 | PASS x4 |
| 1b Observation landed upstream | PASS | PASS | PASS | PASS |
| 2 redaction + `_source: upstream` | PASS | PASS | PASS | PASS |
| 3 audit written, PHI-free | PASS | PASS | PASS | PASS |
| 4 `$conformance` | **FAIL F 1/7** | **PASS B 6/7** | PASS B 6/7 | PASS B 6/7 |
| 5 MCP `tools/list` | FAIL — 000 | FAIL — 000 | FAIL — 000 | FAIL — 000 |

**One line differs, and it is better.** `$conformance` against HAPI now grades
B 6/7 because #514 fixed the probe collision the pack diagnosed. The pack
argued the F came from the harness rather than the guardrails, with Firely's B
as the control; changing only the probe makes HAPI grade what Firely graded, so
both halves of that diagnosis are now confirmed directly.

Step 5 is red on both days for one reason: the MCP server runs in Docker, and
Docker was down on this machine both times. That is an absence, not a refusal.

## The script can fail

`docs/evidence/2026-09-04-set2-rerun/negative-control-local-mode.txt`: told
`hapi` with the app in local mode, it exits 2 at step 0 and reports no pass.
That is the trap `smoke_medplum.py` fell into — 7 of 8 green against a Medplum
that did not exist (R2) — set for this script deliberately.

## What is now committed

- `scripts/walkthrough-upstream.sh` — one parametrized script, not
  `walkthrough-hapi.sh` plus a near-copy. Two copies of one assertion list is
  how the Aidbox script's status codes drifted from its own README (#499), and
  a second copy would need its own pin to stop it drifting again. Flagging the
  deviation from #530's literal step 1 for the reviewer.
- `docs/evidence/2026-09-04-set2-rerun/*.txt` — raw transcripts, ANSI stripped,
  scanned for home-directory paths (the 2026-08-16 incident).
- `docs/evidence/2026-09-04-set2-connectors-rerun.md` — what was run, what
  matched, what did not, and what this still does not prove.
- `tests/test_public_walkthrough_tells_the_truth.py` — 14 tests. Pins the four
  status codes against this app, the three lines that stop the script passing
  vacuously, and that every document citing this evidence still names a path
  that exists.

## Documents corrected

Five claim sites, not the three #530 names — `docs/prd/README.md` and
`docs/2026-08-16-tester-program.md` repeat the number too. Each now carries the
re-run date, that a second person ran it, the script path and the transcript
path.

The 2026-08-16 pack is **annotated, not rewritten**. R8 is closed;
"Reproducing this" named a directory that no longer exists and now points at
the committed paths. Its sections 5, 6 and 7 rest on four other uncommitted
scripts (`registry-contract.sh`, `auth_probe.py`, `halfconfig.sh`,
`medplum-qa.sh`) that are also gone, and are marked as still not independently
re-run. hard-truths §4's "graded F" row stays as written — it was true that
day — with a note that #514 and #512 have since fixed two of its five rows.

## What this does NOT cover

- `aidbox` and `medplum` — not attempted. Docker still down, no Medplum
  credentials here. The table is still two rows of yes.
- Either kind **with a credential**. Both public sandboxes are anonymous, so
  the "proxy holds its own credential" property is not demonstrated by either
  run; the script reports that as a NOTE rather than a guaranteed red.
- The MCP step, the pinned 1.10.0 images, and the `qa/demo.spec.ts` recording.
  The 2026-08-16 pack stays EVIDENCE PARTIAL.
- Cleanup on the public servers: one Patient and one Observation each, plus
  whatever `$conformance` created, all synthetic, left in place — same
  footprint as the pack's day. Nonces `86425f1cde39` and `471213ce4181`.

## Gates

```
uv run ruff check .   -> All checks passed!
  -> 3171 passed, 13 skipped, 1 xfailed, 2 warnings in 100.97s
  -> 37 passed
```

No production code changed. Three mutations were run against the new tests to
confirm they fail without what they pin; the first version of the `_source`
guard survived one of them and was rewritten (noted in the test's docstring).

**Do not approve or enable auto-merge.** An approval here arms a merge that
fires on the next green check.

### DIFF
--- a/docs/2026-08-16-hard-truths.md
+++ b/docs/2026-08-16-hard-truths.md
@@ -17,7 +17,7 @@ found wanting" — never run.
 | Set | Measured state |
 |---|---|
 | Guardrail core | Grade A local; **proxy mode unmeasured** |
-| Connectors | **2 of 4** kinds proven live |
+| Connectors | **2 of 4** kinds proven live — reproduced by a second person 2026-09-04 |
 | Consumer journey | **unmeasured** |
 | Action rail | **unmeasured** |
 | Clinical rails | **unmeasured** |
@@ -65,6 +65,13 @@ This is the uncomfortable pattern, and it is consistent:
 
 The polish is the risk. A rough edge invites inspection; a confident sentence
 next to a passing test does not.
+
+Row 1 and row 3 have since been fixed (#514, #512). The 2026-09-04 re-run of
+the same walkthrough grades that same HAPI **B 6/7** — the pack's diagnosis
+that the F came from the probe and not from the guardrails, confirmed by
+changing only the probe. The rows stay as written because they are what was
+true on 2026-08-16, and the pattern they illustrate is not repealed by fixing
+two instances of it.
 
 ## 5. Guards are written narrower than the property they are named after
 
--- a/docs/2026-08-16-system-topology.md
+++ b/docs/2026-08-16-system-topology.md
@@ -44,14 +44,21 @@ problem it was written to solve.
 ## The four upstream connectors, and which are proven
 
 Measured 2026-08-16 by the set-2 evidence run
-(`docs/evidence/2026-08-16-set2-connectors.md`):
+(`docs/evidence/2026-08-16-set2-connectors.md`), and **re-run on 2026-09-04 by
+someone other than its author** (`docs/evidence/2026-09-04-set2-connectors-rerun.md`,
+#530). The two `yes` rows reproduced step for step.
 
 | Kind | Auth | Proven live against |
 |---|---|---|
-| `hapi` | Basic / anonymous | **HAPI FHIR 8.11.16** ✓ |
-| `generic` | Basic / anonymous | **Firely Server 6.9.1** ✓ |
-| `aidbox` | HTTP Basic | **not run** — local server down |
-| `medplum` | OAuth2 client-credentials | **not run** — local server down |
+| `hapi` | Basic / anonymous | **HAPI FHIR 8.11.16** ✓ twice |
+| `generic` | Basic / anonymous | **Firely Server 6.9.1** ✓ twice |
+| `aidbox` | HTTP Basic | **not run** — local server down, both days |
+| `medplum` | OAuth2 client-credentials | **not run** — local server down, both days |
+
+Anyone can now re-run the two: `scripts/walkthrough-upstream.sh hapi|generic`,
+transcripts in `docs/evidence/2026-09-04-set2-rerun/`. Until 2026-09-04 the
+claim rested on an uncommitted script and could not be checked by anyone but
+its author, which is what #530 was.
 
 ## Architecture ratchets
 
--- a/docs/2026-08-16-system-topology.md
+++ b/docs/2026-08-16-system-topology.md
@@ -50,8 +50,8 @@ someone other than its author** (`docs/evidence/2026-09-04-set2-connectors-rerun
 
 | Kind | Auth | Proven live against |
 |---|---|---|
-| `hapi` | Basic / anonymous | **HAPI FHIR 8.11.16** ✓ twice |
-| `generic` | Basic / anonymous | **Firely Server 6.9.1** ✓ twice |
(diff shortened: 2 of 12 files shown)


aks129:
Follow-up commit b8ed646, self-caught on review.

The first commit marked HAPI 8.11.16 and Firely 6.9.1 as proven "twice", but neither version was measured on 2026-09-04. The walkthrough prints `software` and `fhir_version` from the health payload — "HAPI FHIR Server" and "4.0.1" — and never reads `software.version`, so nothing in the transcripts records which build answered. Both numbers were the 08-16 pack's, carried forward on trust and given today's tick. hapi.fhir.org serves SNAPSHOT builds that move, so that was a stale number presented as re-verified — in the document being corrected for exactly that.

Measured, two curl calls:

```
HAPI FHIR Server 8.11.16-SNAPSHOT/7e7129efb5/2026-07-08
Firely Server 6.9.1+e4f2e63c5e20f326898f61339c3b3fcd48017201
```

Both match the pack, and HAPI is the same build rather than merely the same version string. The claim stands; the documents now say what was measured.

The script is deliberately unchanged — printing `software.version` would break the property that the committed script is byte-identical to the one that produced the committed transcripts. Logged in the rerun pack as a follow-up for the next run.

Gates after the correction: ruff clean; 23 passed on the two "tells the truth" files. The full suite ran green at 3171 before this docs-only delta.

Issue #608, opened 2026-09-04
the standards review has never run — the key is unset, the step skips, and the check reports success anyway


aks129:
Found while quieting the routine pull-request emails (#607). The message being removed was not only repetitive — **it asserted a review verdict that never happened.**

## What is true today

`ANTHROPIC_API_KEY` is unset, so in `.github/workflows/claude-pr-review.yml` the review step and the verdict gate both **skip**. The job `claude-standards-review` still reports **success**, because a skipped step is not a failed one.

So:

- Every open pull request shows `claude-standards-review: SUCCESS`.
- **None of them carries the `claude:approve` label** the verdict gate is supposed to apply. Checked across #566, #579, #595, #601 and #605 — all empty.
- The comment posted on unapproved pull requests said "Standards review passed (claude:approve)". It said that on pull requests where no review ran.
- Auto-merge arms on a maintainer approval with **no standards review having taken place**.

Evidence: run `33868870495` shows both `Claude standards review` and `Gate on verdict label` as `skipped`, with both jobs green.

## Why this is the worst instance of a pattern this project keeps finding

The catalogue calls it a control that reports on something other than what its name says. This is that, at the top of the stack: the check named for the review reports whether the *job* fell over, not whether a *review* occurred. Everyone reading the pull request page — including me, all night — has been reading it as the second.

The workflow header still lists the verdict as a merge condition, which describes an arrangement that is not in force.

## What #607 does and does not do

It stops the arming step from claiming a verdict it cannot see. That is the honest half.

It deliberately does **not** close the gate, because correcting the words while leaving the hole is the exact failure this project has written down: the prose gets fixed, everyone relaxes, and the mechanism stays open.

## The decision, which is the owner's

Should an unconfigured reviewer **block** merges, or should merges proceed without it?

1. **Block.** The gate fails closed when the key is missing. Safest, and it stops everything until the key is set.
2. **Proceed, honestly.** The check reports "not configured" rather than success, so nobody reads a green that means nothing. Cheapest, and it keeps the current behaviour while ending the false signal.
3. **Configure it.** Set the key and let the reviewer actually run, which is presumably what was intended when this was built.

Whichever is chosen, the workflow header needs to describe it, and a test should pin that the check cannot report success when no review ran.

## Related

#607, and the traps in `docs/agent-task-guide.md` §6 on approval and merging.

PR #612, opened 2026-09-04
candidate: demo_e2e gates 6, 8, 9 execute instead of silently skipping

aks129:
## What

Three of the ten pre-deploy gates in `scripts/demo_e2e.sh` were not running,
while the script reported them green.

`PATIENT_ID` was extracted from `SEED_RESP.created[]`, an array the seed
endpoint no longer returns (it returns `created_count`); the extraction
always produced an empty string. The check that was supposed to catch that —
`check "Seeded patient ID extracted" "." "${PATIENT_ID:-none}"` — matches the
literal fallback string `"none"` against the pattern `.` (any character), so
it always passed regardless of whether extraction worked. With `PATIENT_ID`
empty, **gate 6 (PHI redaction on read) and gate 8 (cross-tenant isolation)**
never ran their bodies — no assertion, no failure, just silently absent from
the output, between two gate headers that printed as normal.

Gate 9 (curatr) POSTed `$curatr-evaluate`, which is a GET-only operation
(405), and searched `Condition?patient=$PATIENT_ID` with a bare id — the
`patient` search param matches `subject.reference` and needs the full
`Patient/{id}` form, per the CapabilityStatement's own documented behavior.

Gate 11 asserted commit reaches `status=completed`. The human gate parks a
committed action at `awaiting_confirmation` by design
(`r6/actions/models.py` STATUS_TRANSITIONS) — the old assertion could only
pass if the out-of-band approval gate were bypassed.

## Why

Property protected: the two highest-value guards in this codebase — PHI
redaction and tenant isolation — actually execute before every deploy, not
just report green.

## Ran

Local stack (Flask :5001, MCP :3001), fresh seed each run:

- Before: `10 / 13` passed, with gates 6, 8 and part of 9 silently absent
  from the output (not failing — missing).
- After: `16 / 16`, every gate present and asserting.

## Mutation evidence

Reverted the patient-ID extraction and the check pattern to the pre-fix
form (`created[]` read + `"." ` pattern) while leaving the rest of the fix
in place, and re-ran against the same live stack. Result: gate 6 and gate 8
went back to silently absent from the output — same shape as the original
defect, confirming the check itself (not just the extraction) was the
decoration. Restored.

## What was NOT run

Against the deployed (Railway) stack — only the local one built this
session.

## Generalization check

No real names, tenant ids, or tokens in this PR body or the diff.

### DIFF
--- a/scripts/demo_e2e.sh
+++ b/scripts/demo_e2e.sh
@@ -104,7 +104,7 @@ SEED_RESP=$(curl -sf -X POST "$FHIR_BASE/internal/seed" \
   -H "X-Tenant-ID: $TENANT_ID" \
   -H "X-Step-Up-Token: $STEP_UP_TOKEN" \
   -d "{\"tenant_id\":\"$TENANT_ID\"}" 2>/dev/null || echo '{}')
-check "Seed created resources" '"created"' "$SEED_RESP"
+check "Seed created resources" '"created_count"' "$SEED_RESP"
 
 # Extract seeded token if provided (seed returns a fresh token)
 SEED_TOKEN=$(echo "$SEED_RESP" | python3 -c "import sys,json; d=json.load(sys.stdin); print(d.get('step_up_token', ''))" 2>/dev/null || echo "")
@@ -112,16 +112,15 @@ if [ -n "$SEED_TOKEN" ]; then
   STEP_UP_TOKEN="$SEED_TOKEN"
 fi
 
-PATIENT_ID=$(echo "$SEED_RESP" | python3 -c "
+PATIENT_ID=$(curl -sf "$FHIR_BASE/Patient?_count=1" \
+  -H "X-Tenant-ID: $TENANT_ID" 2>/dev/null | python3 -c "
 import sys, json
 d = json.load(sys.stdin)
-created = d.get('created', [])
-for r in created:
-  if isinstance(r, dict) and r.get('resourceType') == 'Patient':
-    print(r.get('id',''))
-    break
+entries = d.get('entry', [])
+print(entries[0].get('resource', {}).get('id', '') if entries else '')
 " 2>/dev/null || echo "")
-check "Seeded patient ID extracted" "." "${PATIENT_ID:-none}"
+# Sentinel starts with '!' so an empty id can never satisfy the pattern.
+check "Seeded patient ID extracted" "^[A-Za-z0-9][A-Za-z0-9._-]*$" "${PATIENT_ID:-!none}"
 
 # ─────────────────────────────────────────────────────
 # GATE 6: Read with PHI redaction
@@ -175,7 +174,9 @@ fi
 # ─────────────────────────────────────────────────────
 _blue "Gate 9: Curatr evaluation"
 
-CONDITION_RESP=$(curl -sf "$FHIR_BASE/Condition?patient=$PATIENT_ID" \
+# The `patient` search param matches subject.reference, so it needs the
+# full "Patient/{id}" form — a bare id matches nothing.
+CONDITION_RESP=$(curl -sf "$FHIR_BASE/Condition?patient=Patient/$PATIENT_ID" \
   -H "X-Tenant-ID: $TENANT_ID" 2>/dev/null || echo '{}')
 CONDITION_ID=$(echo "$CONDITION_RESP" | python3 -c "
 import sys, json
@@ -186,10 +187,8 @@ if entries:
 " 2>/dev/null || echo "")
 
 if [ -n "$CONDITION_ID" ]; then
-  CURATR_RESP=$(curl -sf -X POST "$FHIR_BASE/Condition/$CONDITION_ID/\$curatr-evaluate" \
-    -H "Content-Type: application/json" \
-    -H "X-Tenant-ID: $TENANT_ID" \
-    -d '{}' 2>/dev/null || echo '{}')
+  CURATR_RESP=$(curl -sf "$FHIR_BASE/Condition/$CONDITION_ID/\$curatr-evaluate" \
+    -H "X-Tenant-ID: $TENANT_ID" 2>/dev/null || echo '{}')
   check "Curatr evaluation returns result" '"issues"\|"quality_score"\|"resourceType"' "$CURATR_RESP"
 else
   _blue "  (skip — no Condition found in seeded data)"
@@ -241,7 +240,9 @@ ACTION_ID=$(echo "$PROPOSE_RESP" | python3 -c "import sys,json; print(json.load(
 check "Action proposed — id returned" "^[0-9a-f-]\{36\}$" "${ACTION_ID:-none}"
 
 if [ -n "$ACTION_ID" ]; then
-  # Commit the action (simulation mode — no provider keys → completes synchronously)
+  # Commit submits the action; the human gate holds it at awaiting_confirmation
+  # until an out-of-band approval claims it (proposed -> awaiting_confirmation
+  # -> executing). See STATUS_TRANSITIONS in r6/actions/models.py.
   COMMIT_RESP=$(curl -s -X POST "$APP_BASE/r6/actions/$ACTION_ID/commit" \
     -H "Content-Type: application/json" \
     -H "X-Tenant-Id: $TENANT_ID" \
@@ -249,7 +250,7 @@ if [ -n "$ACTION_ID" ]; then
     -H "X-Human-Confirmed: true" \
     2>/dev/null || echo '{}')
   COMMIT_STATUS=$(echo "$COMMIT_RESP" | python3 -c "import sys,json; print(json.load(sys.stdin).get('status',''))" 2>/dev/null || echo "")
-  check "Action commit returns status=completed" "completed" "$COMMIT_STATUS"
+  check "Action commit parks at the human gate" "awaiting_confirmation" "$COMMIT_STATUS"
 
   # Verify audit trail recorded ProposedAction events (propose + commit = ≥ 2)
   AUDIT_ACTIONS=$(curl -s "$FHIR_BASE/AuditEvent?_count=50" \
--- a/scripts/demo_e2e.sh
+++ b/scripts/demo_e2e.sh
@@ -18,6 +18,12 @@ TENANT_ID="${TENANT_ID:-demo-e2e-$(date +%s)}"
 STEP_UP_SECRET="${STEP_UP_SECRET:-dev-secret-change-in-production}"
 PASS=0
 FAIL=0
+# Every gate that runs, whether it passes or fails, marks itself ran. A gate
+# whose body is skipped (an empty variable, a short-circuited `if`) never
+# calls this and the final ran-count catches it — PASS+FAIL alone cannot,
+# because a silently-skipped gate contributes to neither.
+declare -a GATES_RAN=()
+gate_ran() { GATES_RAN+=("$1"); }
 
 _green() { printf '\033[0;32m✓ %s\033[0m\n' "$*"; }
 _red()   { printf '\033[0;31m✗ %s\033[0m\n' "$*"; }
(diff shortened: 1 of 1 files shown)


aks129:
Additional mutation evidence, on the *new* assertions rather than just the old absence (per a reviewer's challenge on this queue): mutated the system under test, not the test script.

**Gate 6.** Broke `r6/redaction.py`'s family-name truncation to a no-op (`name_entry['family'] = f` instead of `f[0] + '.'`), restarted Flask, re-ran the fixed script against live data. Gate 6 failed: `PHI redacted: family name is initial only (expected '^[A-Z]\.$' in: Rivera)`. Restored, diff clean.

**Gate 8.** Broke `r6/routes.py`'s single-resource read (`read_resource`) by dropping `tenant_id=tenant_id` from the `R6Resource.query.filter_by(...)` call. Gate 8 failed: `Cross-tenant read returns 404 (not 200) (expected '404' in: 200)` — i.e. it caught a real cross-tenant PHI leak. Restored, diff clean.

Both new assertions catch the exact defect they're named for, not just the shape of the old skip.

aks129:
## QA review — FAIL (changes requested)

Reviewed as an adversarial second pass, on a live local stack (Flask :5099 with a
fresh SQLite DB, the real MCP orchestrator on :3001), not against fakes.

**The diagnosis and the fix are correct.** I reproduced both baselines exactly,
same stack, same seed:

| script | result |
| --- | --- |
| `origin/main` | `Gates passed: 10 / 13`, exit 1 — gates 6 and 8 print a header and nothing else |
| this branch | `Gates passed: 16 / 16`, exit 0 — every gate emits an assertion |

Your mutation evidence stands too. I did not re-run it, because reverting the fix
is not the question. The question is whether the *new* assertions can pass while
the thing they name is broken. They can. Three findings below, then two smaller ones.

---

### HIGH-1 — gate 6 passes while PHI leaks

`scripts/demo_e2e.sh:143` asserts one field of one element: `name[0].family`
matches `^[A-Z]\.$`.

I mutated the line *immediately below* the one you mutated — the given-name
truncation in `r6/redaction.py:50-53` — to a no-op, restarted Flask, re-ran this
branch's script against live data.

`GET /Patient/demo-patient-rivera` now returns:

```json
"name": [{"family": "R.", "given": ["Maria", "Elena"], "use": "official"}]
```

Full first and middle name in the clear. The script printed
`✓ PHI redacted: family name is initial only`, `Gates passed: 16 / 16`, exit 0.

Also unasserted, on a record where redaction handles every one of them:
`birthDate` (year truncation), `identifier` (MRN last-4), `telecom`, `address`.
`name[].text` — a full-name string that `apply_redaction` pops at
`r6/redaction.py:55` — is asserted by nothing here. And the seeded patient carries
no `contact[]`, so `Patient.contact` redaction (`r6/redaction.py:101-124`,
emergency-contact name and phone) cannot be exercised by this script under any
input.

Scope, honestly: `tests/test_claimed_properties_are_pinned.py::test_search_redacts_phi`
*does* catch that mutation, so this is not an unguarded production leak. The
finding is that gate 6 is not the guard its name claims. That matters most for
the failures pytest cannot see — `demo_e2e.sh` is the only gate that runs against
a **deployed** stack (`RELEASING.md:18`), so a configuration-shaped redaction
failure (wrong env, upstream-proxy mode, a stale image) is exactly what it exists
to catch, and it would pass.

Repro:
```
# r6/redaction.py:50-53 -> drop the g[0]+'.' truncation, keep the list comprehension
export SQLALCHEMY_DATABASE_URI=sqlite:////abs/path/e2e.db STEP_UP_SECRET=dev-secret
uv run flask --app main init-db          # or the first request dies as AuditWriteError
PORT=5099 uv run python main.py &
FHIR_BASE=http://localhost:5099/r6/fhir MCP_BASE=http://localhost:3001 bash scripts/demo_e2e.sh   # 16/16, exit 0
curl -s "$FHIR_BASE/Patient/demo-patient-rivera" -H "X-Tenant-ID: <seeded tenant>" | python3 -m json.tool
```

### HIGH-2 — gate 8 passes while a cross-tenant read succeeds

`scripts/demo_e2e.sh:167-169` tests one URL shape: `GET /Patient/{id}`.

I dropped `tenant_id=tenant_id` from the **search** query at `r6/routes.py:1017-1019`,
leaving `read_resource` (`r6/routes.py:605-607`, the path you mutated) intact.
A tenant that has seeded nothing:

```
Probing as tenant 'empty-tenant-1788540412' (never seeded, owns no data)
--- GET /Patient?_count=5  (search path) ---
  total   = 5
  entries = 5
   -> Patient demo-patient-rivera [{"family": "R.", "given": ["M.", "E."], ...}]
   ... x5, all belonging to other tenants
--- GET /Patient/demo-patient-rivera  (single-read path, what gate 8 tests) ---
  HTTP 404
```

Five other tenants' records handed to a stranger, and the script printed
`✓ Cross-tenant read returns 404 (not 200)`, `Gates passed: 16 / 16`, exit 0.

Search is the path that builds an agent's context. Untested by gate 8: search,
`_history`/vread, update, delete, AuditEvent search, context envelopes.

Same honest scope as HIGH-1:
`tests/test_r6_dashboard.py::TestCrossTenantIsolationComprehensive::test_search_tenant_isolated`
catches this mutation. Gate 8 is narrower than the property it is named for; it
is not redundant coverage of tenant isolation, and should not be read as such in
a release checklist.

### HIGH-3 — "16 / 16" means sixteen were *counted*, not sixteen *ran* — and gate 9 can still vanish silently

`scripts/demo_e2e.sh:273`: `TOTAL=$((PASS+FAIL))`. The denominator is derived
from the checks that executed, so a skipped gate shrinks numerator and
denominator together and leaves the ratio looking perfect. That is the defect
this PR is about, and it is still in the file — in gate 9, one of the gates the
PR claims to fix. Lines 193-194 are still a bare `else` with a `_blue` message
and no failing sentinel.

I simulated the same class of drift that caused the original defect (the
Condition search stops matching), in a **copy** of the script in my own
scratchpad — the repo script was not touched:

```
→ Gate 9: Curatr evaluation
→   (skip — no Condition found in seeded data)
...
  Gates passed: 15 / 15
  All gates passed.
EXIT=0
```

Green, exit 0, gate 9 never ran. `docs/2026-08-02-retro.md`, third pattern, says
this in as many words: *"Nothing asserts that the e2e suite ran at all. That is
the actual gap."*

Gates 6 and 8 are now protected only indirectly — the new sentinel at line 123
turns the run red — but their `if [ -n "$PATIENT_ID" ]` guards (lines 130, 165)
are unchanged, so on that path the two gate bodies still disappear from the
output, now merely alongside one red line.

**Were the other gates ever silently skipping too?** Every conditional in the
script: line 111 (token swap, contains no check), 130 and 165 (gates 6 and 8, now
protected upstream by the sentinel at 123), 189-194 (gate 9, **unprotected**),
242 (gate 11, protected by the anchored check at line 240). So three gates could
vanish silently on main — 6, 8 and 9. This PR protects two of them; 11 was always
protected; 9 still is not.

Suggested (your call on shape):
- `EXPECTED_CHECKS=16`, asserted against `TOTAL` before the summary.
- Gate 9's `else` becomes `gate_fail`, not `_blue`.
- Drop the `if [ -n "$PATIENT_ID" ]` guards on 6 and 8 now that line 123 fails
  hard. A skip should not be a code path.

---

### MEDIUM — `RELEASING.md:18` and the script's summary count different things

The checklist says "all 11 gates" — gate *headers*. The summary counts *checks*:
13 on main, 16 here. Different units, and nothing reconciles them. The mismatch
predates this PR, but this PR is what moves the number a release captain reads,
and there is still no declared expected count anywhere, so "16 / 16" cannot be
told from "15 / 15" by anyone following the checklist.

### MEDIUM — the `.` idiom that caused this defect survives at line 95

`check "Step-up token issued" "." "$STEP_UP_TOKEN"` is the same pattern you
correctly killed at line 123. It is safe today only because it has no `:-`
fallback (`echo "" | grep -q "."` exits 1). Add a tidy-looking `${STEP_UP_TOKEN:-none}`
and the original bug is back, in the gate that issues the credential every later
write depends on. Anchor it the way you anchored 123.

### LOW — gate 9's assertion is satisfied by any FHIR resource, including an error

Line 192: `'"issues"\|"quality_score"\|"resourceType"'`. The third alternative
matches a bare `{"resourceType":"OperationOutcome",...}` (verified). Latent rather
than live — `$curatr-evaluate` returns 200 with `"issues"` today, and `curl -sf`
suppresses non-2xx bodies — but as written the check asserts "a FHIR response came
back", not "an evaluation ran". Dropping `"resourceType"` costs nothing.

### LOW — line 61-62's "informational, not a blocking gate" comment is false

`check` always returns 0 (the last statement in `gate_pass`/`gate_fail` is an
arithmetic assignment), so the trailing `|| true` is dead and the MCP health
check counts toward the total and can fail the run. Pre-existing, but this PR
changes what the total means: without the MCP server running you get 15/16 and
exit 1, which now reads as a gate failure.

### One correction for the PR body

Two counts in the body do not survive a re-run. "while the script reported them
green" — main exited 1 with `Gates failed: 3`. Three checks failed **visibly**
(`"created"` vs `created_count`, the 405, `completed` vs `awaiting_confirmation`);
two gates were **silently absent**. And "three of the ten pre-deploy gates" —
there are 11 gate headers, and 13 counted checks. In a review about counting
semantics these are worth fixing. Separately, the empty
`PATIENT_ID` also degraded gate 10 rather than skipping it: main posted
`"subject":{"reference":"Patient/"}` and still printed ✓ 428 — passing, but on a
malformed resource.

---

### Gates run

- `uv run python -m pytest tests/ -q` — `3157 passed, 13 skipped, 1 xfailed in 102.29s`, exit 0
- `uv run ruff check .` — `All checks passed!`
- `uv run python -m pytest tests/test_guardrail_conformance.py -q` — `37 passed`
  (run separately to confirm Grade A is executed, not among the 13 skipped)
- `scripts/demo_e2e.sh` on this branch against the live stack — `16 / 16`, exit 0

### What I did NOT cover

- The deployed Railway stack. Local only, same limitation you noted.
- MCP tool surface beyond `/health`; `mcp_call` is defined and never used.
- Whether the gate 1-5, 7, 10 and 11 assertions are narrower than their names. I
  attacked 6, 8 and 9 — the two guarantees named in the PR, plus the one whose
  skip path survives.
- Upstream-proxy mode. Both mutations were run in local SQLite mode only.

No production code changed on this branch; both mutations were reverted and
`git status` is clean. The demonstration scripts live in my own worktree's
scratchpad, not in the repo. Not approving — an approval arms the merge, and
`#608` means the standards-review check reports success without running.

aks129:
All three findings fixed in dcd33dd — thank you, this was the right kind of attack and it found real gaps in what "the gate passed" actually meant.

**Gate 6** now asserts: a positive check that none of the five raw seeded PII values (family, given×2, birthDate, phone) appear anywhere in the response body, plus the specific redacted form of family, given, birthDate and telecom individually. Reran your exact mutation (given-name truncation removed) against a genuinely-restarted stack — confirmed live via direct curl before trusting the gate — and it now fails on 2 assertions instead of passing on 16/16.

**Gate 8** now also searches under the other tenant and asserts the seeded patient's id is absent from the result set, alongside the existing targeted-read check. Reran your exact mutation (tenant filter dropped from the search query, read path left intact) — the read assertion correctly stayed green (different code path, genuinely unaffected) while the new search assertion catches the leak.

**The "16 of 16 means 16 ran" gap** — fixed with a fixed 21-item `EXPECTED_CHECKS` list compared against what `check()` actually recorded running; anything missing is printed by name and forces `exit 1` regardless of the pass/fail ratio of what did run. Simulated your gate-9 scenario (forced `CONDITION_ID=""` to skip its check entirely): before this fix it would have read "20 of 20, all passed", exit 0. Now: "20 of 20" for what ran, plus "Gates that never ran: 1 — Curatr evaluation returns result", exit 1.

One thing I'll flag rather than bury: my first mutation test attempt gave a false negative — a still-running Flask process from a failed restart (port already in use, silent) served the old unmutated code while I thought I was testing the mutant. Caught it by checking the direct curl response matched what the mutation should produce before trusting the gate's verdict, which is exactly the discipline you're describing. Left the near-miss in the commit message rather than quietly redoing it clean.

Restored all three mutations, reran clean: 21/21, exit 0.

Issue #112, opened 2026-07-13
Production rigor for de-identification (Expert Determination) and validation (StructureDefinition + terminology)


aks129:
## Honest limits → production rigor

Today's guardrail contract demonstrates **redact + audit + step-up + human-confirm + tenant isolation** (and, with #108, error fidelity). But two of those pieces are demonstrated at *reference* rigor, not *production* rigor, and we should be explicit about it and track the gap to close.

> **Honest limits (see the README [Known Limitations](../blob/main/README.md#known-limitations)):**
> Redaction is HIPAA **Safe-Harbor-style field redaction** (demographics), **not Expert Determination**; validation is **structural**, not full StructureDefinition / terminology conformance. The guardrail *contract* (redact + audit + step-up + human-confirm + tenant isolation) is what's demonstrated here — production de-id / validation rigor is on the roadmap.

### Scope — two workstreams

**1. De-identification rigor (redaction)**
- [ ] Document the current posture precisely: Safe-Harbor-*style* field redaction of the 18 identifier classes on demographics, applied recursively (nested `contact`, extensions, attachments) — a compensating control, **not** a legal Safe Harbor or Expert Determination determination.
- [ ] Profile-specific recursive allowlists (deny-by-default) rather than field-name denylists, so a new PHI-bearing field can't leak by omission.
- [ ] An **Expert Determination** path option (statistical de-id / k-anonymity-style checks with documented re-identification risk), gated behind an explicit profile — never the default.
- [ ] Stop labeling any output "HIPAA Safe Harbor" in code/docs until a profile is independently validated; use "Safe-Harbor-style" until then.
- [ ] A PHI-canary test suite that asserts *nested* displays, extensions, attachments, URLs, dates, and identifiers are all stripped for each profile.

**2. Validation rigor**
- [ ] Current `fhir_validate` is **structural** (shape + a few required-field warnings). Document that clearly at the tool boundary.
- [ ] Full **StructureDefinition conformance** (profile-aware: US Core / the project IG) with cardinality, invariants, and slicing.
- [ ] **Terminology binding** validation (ValueSet/CodeSystem membership) against a terminology service, with a documented offline fallback.
- [ ] Grade validation depth in `$conformance` so a deployment's validation rigor is visible, not assumed.

### Why now
[...]

Issue #585, opened 2026-09-04
CI never runs on stacked pull requests, but the auto-merge job does


aks129:
Found while checking why a dependency check was missing on some pull requests. The missing check was a symptom.

## What happens

`.github/workflows/ci.yml` triggers on:

```yaml
on:
  pull_request:
    branches: [main, master]
```

That filter is on the **base** branch. A pull request whose base is another feature branch — a stacked change — matches nothing, so **the entire CI workflow never runs on it**.

Measured across the open pull requests today:

| | Checks run |
|---|---|
| Targeting `main` | 19 |
| Stacked on another branch | 6 |

The six that do run are the standards review, the secret scanner, the deployment preview, and the auto-merge job. **None of the six is a test.** No lint, no Python tests, no Postgres lane, no browser tests, no compose smoke test, no compliance gates, no code scanning, no dependency audit.

## Why this is urgent rather than tidy-up

`auto-merge-when-satisfied` **does** run on stacked pull requests, and on this repository a maintainer approval arms an automatic squash merge. With no test job present, there is no failing required check to hold it. So approving a stacked change merges it having run **zero tests**, and the pull request page looks clean while doing it.

Seven open changes are in that state right now, including the fix for a data leak on the intake path and the patient-facing corrections on the approval path. Each was tested locally by its author and its reviewer, so this is not a claim that they are broken — it is that the gate everyone believes is holding is not present.

## Fix

Drop the base-branch filter so the workflow runs on every pull request:

```yaml
on:
  pull_request:
```

The `push` trigger can keep its branch list. If some jobs are genuinely too expensive for every stack, gate those individually rather than skipping the whole workflow, and say which and why in a comment.

## Tests

- A pull request targeting a non-`main` base runs the test jobs. Verify on a real stacked pull request rather than by reading the file.
- The count of jobs that run is the same for a stacked change as for one targeting `main`, or the difference is deliberate and commented.

## Related

The 2026-08-16 incident recorded in the project notes is the same family: a gate everyone believed was holding, which was not.

aks129:
Re-measured today against live check rollups, because this is now load-bearing: eleven open changes across four chains are stacked, and the merge order people are being handed depends on knowing what their green means.

**Sixteen checks on a change based on `main`. Six on a stacked one.**

`#629` (base `main`):
```
Analyze (actions)          Analyze (javascript-typescript)   Analyze (python)
CodeQL                     Vercel                            Vercel Preview Comments
[...]

PR #615, opened 2026-09-04
candidate: remove identifier values instead of last-4 truncation, and close the patient-controlled recursion leak (#112, #617)

aks129:
**This is no longer a routine Safe Harbor cleanup — it closes a live PHI leak (#617) on `main`.** Patient-controlled de-identification (feeds `$share-bundle` and `$deidentify` — the paths whose entire purpose is producing something safe to hand to someone else) stopped at the top level. A contained resource's name, a `subject.display`, or a `generalPractitioner[].display` survived untouched inside output stamped `ANONYED`. #617 stays exposed on `main` until this merges — the fix existing in an open, unmerged PR does not close it. Recommend this jumps the queue ahead of PRs that are merely correct.

---

candidate: remove identifier values from redaction instead of keeping the last four characters (#112)

## What

`r6.redaction.apply_redaction` (the Safe Harbor / standard profile) kept the
last four characters of every `identifier.value` (`'***' + val[-4:]`). It now
removes `value` entirely and keeps `system` and `type`, so a reader can still
see that a patient had an SSN or an MRN of some kind, but not any part of the
value.

`apply_patient_controlled_redaction` had the same shape one layer down: its
docstring said the injected `https://healthclaw.io/patient-id` identifier is
the *sole* identifier on the output, but the code filtered by a keyword
denylist (`'mrn', 'facility', 'org/', 'example.org'`, plus two hardcoded SSN
systems) and let anything that didn't match through with its value intact —
a MRN under `https://fhir.example-health.test/ids` survived untouched. That
function is now built, not filtered: the output identifier list is always
exactly the one injected entry, matching what the docstring always claimed.

Touched:
- `r6/redaction.py` — both functions, plus the module docstring and the
  `apply_patient_controlled_redaction` docstring.
- `SECURITY.md`, `README.md` (two lines), `skills/phi-redaction/SKILL.md`,
  `skills/fhir-r6-guardrails/SKILL.md` — the "masked to last 4 characters"
  claim, corrected everywhere it appeared on a public surface.
- `tests/test_redaction_identifiers.py` (new) — the direct guard.
- `tests/test_terminology_labels.py` — added an ordering pin for the
  strip-then-label sequence (see Architecture review).
- Seven existing tests that pinned `'***'` fixed to assert the value is gone:
  `test_r6_routes.py`, `test_fhir_proxy.py` (x2), `test_medplum_in_front.py`,
  `test_context_builder.py`, `test_r6_dashboard.py` (x2),
  `test_public_fhir_servers.py`.
- `tests/test_deidentification_language.py` — the guard that forbids an
  unhedged "Safe Harbor" claim on a public surface. Its own explanatory text
  named the last-four truncation as the reason the claim was wrong; updated
  to say identifier values are now removed, and that the profile remains a
  compensating control (birth year, state, and country still pass through)
  so the disclaimer requirement itself does not change.

Not touched: telecom (`[Redacted]`), address line/city/postalCode (removed,
state/country kept), birthDate (year only), names (initials) — none of those
were the finding, and the ruling text asked specifically about identifiers.

## Why (patient problem + property protected)

The patient problem: a redacted record that still carries `***-4567` is not
de-identified for someone who already has a partial SSN, an insurance card,
or a related record — the classic mosaic/linkage attack Safe Harbor's
"remove, don't truncate" rule exists to close. `docs/2026-08-16-hard-truths.md`
§4 named this exact gap as the worst-scored artifact in the codebase: the
most polished document (SECURITY.md) made the least true claim.

The property protected: **an identifier value that could re-identify a
patient does not leave the redaction boundary, in either redaction profile.**
Before this change that property was false for both `apply_redaction` and
`apply_patient_controlled_redaction`.

## Decision: code was wrong, not the policy document

The brief asked me to decide explicitly which side was correct — SECURITY.md
said "HIPAA Safe Harbor" in spirit and the code truncated; those are opposite
claims and only one can be right.

**The policy document is right. The code was wrong, and I fixed the code.**

HIPAA Safe Harbor §164.514(b)(2)(i) lists, among the 18 identifiers to be
removed: (G) Social Security numbers, (H) medical record numbers, (I) health
plan beneficiary numbers, (J) account numbers. The rule's verb is "removed,"
full stop — there is no "except the last four characters" carve-out anywhere
in the enumerated list, and a truncated SSN or MRN is a standard
re-identification vector precisely because so few individuals share a given
last-four suffix within a small population (the same reasoning the rule
applies to dates and ZIP codes elsewhere in the same section). Keeping
`system`/`type` is not itself an identifier value under (G)–(J): it says
"this person had an SSN on file," not what the SSN was, and it is not on the
enumerated list.

I did not soften SECURITY.md's disclaimer language to match a truncating
implementation, because there was no reading of §164.514(b)(2)(i) under which
truncation satisfies it. The disclaimer itself (Safe-Harbor-*style*, not a
legal determination) stays, because the profile still isn't Expert
Determination and still keeps birth year / state / country — that hedge was
never about the identifier truncation specifically, and removing the
identifier value doesn't retire it.

## Architecture review

**Serves the vision or adjacent?** Directly serves it. The constitution says
documentation is product surface and a doc that contradicts the system is a
defect with the same severity as the code being wrong — this was exactly
that defect, in the one document (SECURITY.md) a partner's compliance
reviewer reads first.

**Honest failure mode and who notices?** The remaining failure mode is
unchanged and disclosed: this is field-level redaction, not Expert
Determination or a certified statistical/expert method, and free text
outside the fields this profile touches (a narrative, an extension, a
clinician's note field this profile doesn't know about) can still carry PHI
if it isn't one of the fields `_redact_recursive` walks. Nobody but a future
code reader notices that gap today; it's the same limit `#112` already
tracks and this PR does not change its scope. What *did* change: identifiers
specifically are now removed rather than partially retained, closing the
specific mosaic-attack vector the hard-truths audit named.

**What does it make harder later?** Nothing structural. If a future feature
needs to correlate the SAME patient's redacted records across two reads
(e.g. to show "you have 3 records with an MRN on file" without re-identifying
which), `system`/`type` survive and support that; if it needs a *stable but
non-reversible* identifier token, that's a new, deliberate feature (a hash or
a per-tenant pseudonym), not a reason to keep truncated values around as a
found-money substitute. `apply_patient_controlled_redaction`'s canonical
`healthclaw.io/patient-id` identifier already is that stable token for the
patient's own copy of their data.

**How is it proven, with what data, run by whom?** By this session, against
synthetic fixtures only (`test_redaction_identifiers.py`, and the existing
suite's synthetic SSNs/MRNs like `000-00-9999`, `MRN-99`). Mutation-checked
directly (see below) rather than only asserted. Not proven against a live
upstream feed's actual identifier shapes — that's the same limit every other
redaction test in this suite already carries.

## Ran

```
$ uv run python -m pytest tests/test_redaction_identifiers.py -q
5 passed in 0.09s

$ uv run python -m pytest tests/test_ratchets.py tests/test_guardrail_conformance.py tests/test_redaction*.py tests/test_terminology_labels.py -q
83 passed in 2.14s

$ uv run python -m pytest tests/ -q
3163 passed, 13 skipped, 1 xfailed, 2 warnings in 82.65s (0:01:22)

$ uv run ruff check .
All checks passed!
```

Grade A confirmed: `tests/test_guardrail_conformance.py::test_our_deployment_records_full_conformance`
asserts `report.score == (7, 7)` and `report.grade == "A"`, and it passed.
The probe that checks identifiers is `probe_phi_redaction` in
`r6/conformance/probes.py:600`, specifically the checks named
`"SSN-class identifier masked"` (line 624, direct read) and
`"SSN-class identifier masked (search)"` (line 649, search path) — both
assert the literal synthetic SSN (`_SSN = "000-00-9999"`) is absent from the
response body. Removing the value outright still satisfies an absence check;
nothing about this change weakens what the probe verifies, and the change
makes the check strictly harder to pass by accident (a stub with an empty
body still passes it, same as before — see the probe's own two-sided-check
caveat in its comments, unrelated to this PR).

## Mutation evidence

Three guards, each broken in place, confirmed red, then restored and
confirmed green. Observed output only, all in this worktree:

**1. Standard-profile identifier removal** — restored
`ident['value'] = '***' + val[-4:] if len(val) > 4 else '***'` in
`_redact_fields`:
```
FAILED tests/test_redaction_identifiers.py::test_standard_redaction_removes_every_identifier_value
FAILED tests/test_redaction_identifiers.py::test_standard_redaction_reaches_dict_shaped_and_nested_identifiers
2 failed, 3 passed in 0.09s
```
Restored the fix, re-ran: `5 passed in 0.07s`.

**2. Patient-controlled sole-identifier guarantee** — restored the old
keyword-denylist block in `apply_patient_controlled_redaction`:
```
FAILED tests/test_redaction_identifiers.py::test_patient_controlled_leaves_only_the_healthclaw_identifier
AssertionError: ... Left contains one more item: {'system':
'https://fhir.example-health.test/ids', ..., 'value': 'MRN-7749-XYZ'}
1 failed, 4 passed in 0.08s
```
Restored the fix, re-ran: `5 passed in 0.06s`.

**3. Strip-before-label ordering** (the non-negotiable flagged for this PR:
`label_codings` must run on the already-redacted tree, not the raw upstream
one) — swapped the two calls in `apply_redaction` so `label_codings` ran
first:
```
FAILED tests/test_terminology_labels.py::test_label_codings_runs_after_the_redaction_strip
AssertionError: label_codings must run after the strip removed the upstream
display; it instead saw ['Glucose for Jane Secret']
1 failed in 0.07s
```
Restored the original order (`_redact_recursive(redacted)` then
`label_codings(redacted)`), re-ran: `5 passed in 0.06s`. I did not move this
ordering — it was already correct — and added the test because none of the
existing tests pinned the *order* directly (they pinned the outcome, which a
swap could still pass by coincidence on a fixture with only one coding).
Full suite re-run after all three restores: `3163 passed, 13 skipped,
1 xfailed` — identical counts to the pre-mutation run, confirming the
restores were exact.

## What was NOT run

- `npx tsc --noEmit` / `npm test` in `services/agent-orchestrator` — no
  TypeScript was touched.
- Playwright e2e — out of scope per `docs/agent-task-guide.md` §6 (currently
  red on `main` for environment reasons and gives no signal either way).
- No live upstream FHIR server (Aidbox/HAPI/Medplum) run — all evidence is
  against the local SQLite store and synthetic fixtures, consistent with
  every other redaction test in this suite.
- Did not run the Postgres CI lane locally (no column width changed, so nothing
  in this diff is column-length-sensitive, but I have not independently
  confirmed the lane green).

## Generalization check

The fix is `ident.pop('value', None)` in one loop (standard profile) and a
constructed one-entry list (patient-controlled profile) — not a fix that
special-cases the SSN system or any particular shape. It applies uniformly to
every identifier on every resource type that flows through either function,
including the dict-shaped `Reference.identifier` and nested
`Observation.identifier` cases the new test exercises specifically because
the old code's `isinstance(identifiers, list)` guard made it easy to assume
identifiers only ever appear as `resource.identifier` list entries. It does
not depend on English-language system-name keywords (the exact class of bug
being removed from the patient-controlled path) and needs no future
maintenance as new identifier systems appear in upstream feeds.

## Was truncated identifier output already committed to the repo?

Checked, not assumed. Grepped every tracked file (not just code) for the old
`***XXXX` shape. Four hits, all synthetic:

- `docs/evidence/2026-08-16-set1-guardrail-core.md` and
  `docs/evidence/2026-08-16-set2-connectors.md` — dated run-log transcripts,
  each explicitly labeled synthetic in its own text (`urn:set2:evidence` as
  the identifier system on one; "returned record (whole body, synthetic)" on
  the other). Left untouched as history and annotated in place (this PR) with
  a dated note that redaction behaviour changed 2026-09-04 — a reader six
  months out should not mistake `***6789` for current behaviour.
- `examples/aidbox-healthclaw-guardrails/README.md` — a living usage example
  on synthetic patient `pt-demo`, not a dated artifact. Fixed in this PR (it
  would otherwise have silently contradicted the code the moment this merged).
- `skills/phi-redaction/SKILL.md` — prose description, already corrected
  above.

**No real credential or patient data was ever committed to this repository.**
That is a checked claim (the grep above), not an inference from the absence
of alarm.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/PR_BODY-safe-harbor.md
+++ b/PR_BODY-safe-harbor.md
@@ -1,0 +1,220 @@
+candidate: remove identifier values from redaction instead of keeping the last four characters (#112)
+
+## What
+
+`r6.redaction.apply_redaction` (the Safe Harbor / standard profile) kept the
+last four characters of every `identifier.value` (`'***' + val[-4:]`). It now
+removes `value` entirely and keeps `system` and `type`, so a reader can still
+see that a patient had an SSN or an MRN of some kind, but not any part of the
+value.
+
+`apply_patient_controlled_redaction` had the same shape one layer down: its
+docstring said the injected `https://healthclaw.io/patient-id` identifier is
+the *sole* identifier on the output, but the code filtered by a keyword
+denylist (`'mrn', 'facility', 'org/', 'example.org'`, plus two hardcoded SSN
+systems) and let anything that didn't match through with its value intact —
+a MRN under `https://fhir.example-health.test/ids` survived untouched. That
+function is now built, not filtered: the output identifier list is always
+exactly the one injected entry, matching what the docstring always claimed.
+
+Touched:
+- `r6/redaction.py` — both functions, plus the module docstring and the
+  `apply_patient_controlled_redaction` docstring.
+- `SECURITY.md`, `README.md` (two lines), `skills/phi-redaction/SKILL.md`,
+  `skills/fhir-r6-guardrails/SKILL.md` — the "masked to last 4 characters"
+  claim, corrected everywhere it appeared on a public surface.
+- `tests/test_redaction_identifiers.py` (new) — the direct guard.
+- `tests/test_terminology_labels.py` — added an ordering pin for the
+  strip-then-label sequence (see Architecture review).
+- Seven existing tests that pinned `'***'` fixed to assert the value is gone:
+  `test_r6_routes.py`, `test_fhir_proxy.py` (x2), `test_medplum_in_front.py`,
+  `test_context_builder.py`, `test_r6_dashboard.py` (x2),
+  `test_public_fhir_servers.py`.
+- `tests/test_deidentification_language.py` — the guard that forbids an
+  unhedged "Safe Harbor" claim on a public surface. Its own explanatory text
+  named the last-four truncation as the reason the claim was wrong; updated
+  to say identifier values are now removed, and that the profile remains a
+  compensating control (birth year, state, and country still pass through)
+  so the disclaimer requirement itself does not change.
+
+Not touched: telecom (`[Redacted]`), address line/city/postalCode (removed,
+state/country kept), birthDate (year only), names (initials) — none of those
+were the finding, and the ruling text asked specifically about identifiers.
+
+## Why (patient problem + property protected)
+
+The patient problem: a redacted record that still carries `***-4567` is not
+de-identified for someone who already has a partial SSN, an insurance card,
+or a related record — the classic mosaic/linkage attack Safe Harbor's
+"remove, don't truncate" rule exists to close. `docs/2026-08-16-hard-truths.md`
+§4 named this exact gap as the worst-scored artifact in the codebase: the
+most polished document (SECURITY.md) made the least true claim.
+
+The property protected: **an identifier value that could re-identify a
+patient does not leave the redaction boundary, in either redaction profile.**
+Before this change that property was false for both `apply_redaction` and
+`apply_patient_controlled_redaction`.
+
+## Decision: code was wrong, not the policy document
+
+The brief asked me to decide explicitly which side was correct — SECURITY.md
+said "HIPAA Safe Harbor" in spirit and the code truncated; those are opposite
+claims and only one can be right.
+
+**The policy document is right. The code was wrong, and I fixed the code.**
+
+HIPAA Safe Harbor §164.514(b)(2)(i) lists, among the 18 identifiers to be
+removed: (G) Social Security numbers, (H) medical record numbers, (I) health
+plan beneficiary numbers, (J) account numbers. The rule's verb is "removed,"
+full stop — there is no "except the last four characters" carve-out anywhere
+in the enumerated list, and a truncated SSN or MRN is a standard
+re-identification vector precisely because so few individuals share a given
+last-four suffix within a small population (the same reasoning the rule
+applies to dates and ZIP codes elsewhere in the same section). Keeping
+`system`/`type` is not itself an identifier value under (G)–(J): it says
+"this person had an SSN on file," not what the SSN was, and it is not on the
+enumerated list.
+
+I did not soften SECURITY.md's disclaimer language to match a truncating
+implementation, because there was no reading of §164.514(b)(2)(i) under which
+truncation satisfies it. The disclaimer itself (Safe-Harbor-*style*, not a
+legal determination) stays, because the profile still isn't Expert
+Determination and still keeps birth year / state / country — that hedge was
+never about the identifier truncation specifically, and removing the
+identifier value doesn't retire it.
+
+## Architecture review
+
+**Serves the vision or adjacent?** Directly serves it. The constitution says
+documentation is product surface and a doc that contradicts the system is a
+defect with the same severity as the code being wrong — this was exactly
+that defect, in the one document (SECURITY.md) a partner's compliance
+reviewer reads first.
+
+**Honest failure mode and who notices?** The remaining failure mode is
+unchanged and disclosed: this is field-level redaction, not Expert
+Determination or a certified statistical/expert method, and free text
+outside the fields this profile touches (a narrative, an extension, a
+clinician's note field this profile doesn't know about) can still carry PHI
+if it isn't one of the fields `_redact_recursive` walks. Nobody but a future
+code reader notices that gap today; it's the same limit `#112` already
+tracks and this PR does not change its scope. What *did* change: identifiers
+specifically are now removed rather than partially retained, closing the
+specific mosaic-attack vector the hard-truths audit named.
+
+**What does it make harder later?** Nothing structural. If a future feature
+needs to correlate the SAME patient's redacted records across two reads
+(e.g. to show "you have 3 records with an MRN on file" without re-identifying
+which), `system`/`type` survive and support that; if it needs a *stable but
+non-reversible* identifier token, that's a new, deliberate feature (a hash or
+a per-tenant pseudonym), not a reason to keep truncated values around as a
+found-money substitute. `apply_patient_controlled_redaction`'s canonical
+`healthclaw.io/patient-id` identifier already is that stable token for the
+patient's own copy of their data.
+
+**How is it proven, with what data, run by whom?** By this session, against
+synthetic fixtures only (`test_redaction_identifiers.py`, and the existing
+suite's synthetic SSNs/MRNs like `000-00-9999`, `MRN-99`). Mutation-checked
+directly (see below) rather than only asserted. Not proven against a live
+upstream feed's actual identifier shapes — that's the same limit every other
+redaction test in this suite already carries.
+
+## Ran
+
+```
+$ uv run python -m pytest tests/test_redaction_identifiers.py -q
+5 passed in 0.09s
+
+$ uv run python -m pytest tests/test_ratchets.py tests/test_guardrail_conformance.py tests/test_redaction*.py tests/test_terminology_labels.py -q
+83 passed in 2.14s
+
+$ uv run python -m pytest tests/ -q
+3163 passed, 13 skipped, 1 xfailed, 2 warnings in 82.65s (0:01:22)
+
+$ uv run ruff check .
+All checks passed!
+```
+
+Grade A confirmed: `tests/test_guardrail_conformance.py::test_our_deployment_records_full_conformance`
+asserts `report.score == (7, 7)` and `report.grade == "A"`, and it passed.
+The probe that checks identifiers is `probe_phi_redaction` in
+`r6/conformance/probes.py:600`, specifically the checks named
+`"SSN-class identifier masked"` (line 624, direct read) and
+`"SSN-class identifier masked (search)"` (line 649, search path) — both
+assert the literal synthetic SSN (`_SSN = "000-00-9999"`) is absent from the
+response body. Removing the value outright still satisfies an absence check;
+nothing about this change weakens what the probe verifies, and the change
+makes the check strictly harder to pass by accident (a stub with an empty
+body still passes it, same as before — see the probe's own two-sided-check
+caveat in its comments, unrelated to this PR).
+
+## Mutation evidence
+
+Three guards, each broken in place, confirmed red, then restored and
+confirmed green. Observed output only, all in this worktree:
+
+**1. Standard-profile identifier removal** — restored
+`ident['value'] = '***' + val[-4:] if len(val) > 4 else '***'` in
+`_redact_fields`:
+```
+FAILED tests/test_redaction_identifiers.py::test_standard_redaction_removes_every_identifier_value
+FAILED tests/test_redaction_identifiers.py::test_standard_redaction_reaches_dict_shaped_and_nested_identifiers
+2 failed, 3 passed in 0.09s
+```
+Restored the fix, re-ran: `5 passed in 0.07s`.
+
+**2. Patient-controlled sole-identifier guarantee** — restored the old
+keyword-denylist block in `apply_patient_controlled_redaction`:
+```
+FAILED tests/test_redaction_identifiers.py::test_patient_controlled_leaves_only_the_healthclaw_identifier
+AssertionError: ... Left contains one more item: {'system':
+'https://fhir.example-health.test/ids', ..., 'value': 'MRN-7749-XYZ'}
+1 failed, 4 passed in 0.08s
+```
+Restored the fix, re-ran: `5 passed in 0.06s`.
+
+**3. Strip-before-label ordering** (the non-negotiable flagged for this PR:
+`label_codings` must run on the already-redacted tree, not the raw upstream
+one) — swapped the two calls in `apply_redaction` so `label_codings` ran
+first:
+```
+FAILED tests/test_terminology_labels.py::test_label_codings_runs_after_the_redaction_strip
+AssertionError: label_codings must run after the strip removed the upstream
+display; it instead saw ['Glucose for Jane Secret']
+1 failed in 0.07s
+```
+Restored the original order (`_redact_recursive(redacted)` then
+`label_codings(redacted)`), re-ran: `5 passed in 0.06s`. I did not move this
+ordering — it was already correct — and added the test because none of the
+existing tests pinned the *order* directly (they pinned the outcome, which a
+swap could still pass by coincidence on a fixture with only one coding).
+Full suite re-run after all three restores: `3163 passed, 13 skipped,
+1 xfailed` — identical counts to the pre-mutation run, confirming the
+restores were exact.
+
+## What was NOT run
+
+- `npx tsc --noEmit` / `npm test` in `services/agent-orchestrator` — no
+  TypeScript was touched.
+- Playwright e2e — out of scope per `docs/agent-task-guide.md` §6 (currently
+  red on `main` for environment reasons and gives no signal either way).
+- No live upstream FHIR server (Aidbox/HAPI/Medplum) run — all evidence is
+  against the local SQLite store and synthetic fixtures, consistent with
+  every other redaction test in this suite.
+- Did not run the Postgres CI lane locally (no column width changed, so nothing
+  in this diff is column-length-sensitive, but I have not independently
+  confirmed the lane green).
+
+## Generalization check
+
+The fix is `ident.pop('value', None)` in one loop (standard profile) and a
+constructed one-entry list (patient-controlled profile) — not a fix that
+special-cases the SSN system or any particular shape. It applies uniformly to
+every identifier on every resource type that flows through either function,
+including the dict-shaped `Reference.identifier` and nested
+`Observation.identifier` cases the new test exercises specifically because
+the old code's `isinstance(identifiers, list)` guard made it easy to assume
+identifiers only ever appear as `resource.identifier` list entries. It does
+not depend on English-language system-name keywords (the exact class of bug
+being removed from the patient-controlled path) and needs no future
+maintenance as new identifier systems appear in upstream feeds.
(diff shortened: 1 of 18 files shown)


aks129:
## QA review — **FAIL** (on the claims, not the code delta)

The code change is real and is an improvement over `main` in every case I tried. What it **says** it now guarantees is not true, and the two claims are load-bearing: they sit in the module docstring of the highest-stakes function in the system, and in a PR body that will be read as the record of what was closed. This PR fixes a doc/code contradiction in `SECURITY.md` while writing a new one into `r6/redaction.py` — by its own standard ("a doc that contradicts the system is a defect with the same severity as the code being wrong"), that is the same defect class.

Nothing below is a regression. Every leak is byte-identical on `origin/main`. The finding is the distance between what this PR changed and what it claims.

### Gates (run here, verbatim)

```
3163 passed, 13 skipped, 1 xfailed, 2 warnings in 109.71s

uv run ruff check .
All checks passed!

37 passed in 1.07s
```

Grade A holds. Counts match the PR body.

---

## HIGH — the patient-controlled "sole identifier" guarantee is false, on the external sharing path

The docstring says the healthclaw id is "the SOLE canonical identifier"; the PR body says "the output identifier list is always exactly the one injected entry." True of `result['identifier']`. False of the resource. `apply_patient_controlled_redaction` never walks the tree, so this is only a top-level rewrite.

Reproduced against a running HealthClaw (`uv run flask --app main init-db`, then `PORT=5099 STEP_UP_SECRET=dev-secret python main.py`), on this branch:

```
POST /r6/fhir/$share-bundle  {"profile":"deidentified", ...}
```

returned, in a Bundle stamped `ANONYED` **and** `patient-controlled`:

```json
"subject": {
  "display": "Marisol Rivera",
  "identifier": {"system": "urn:mrn", "value": "MRN-7749-XYZ"},
  "reference": "Patient/qa-p1"
},
"subscriberId": "HP-000-00-9999",
"identifier": [{"system": "https://healthclaw.io/patient-id", "value": ""}]
```

Three things in that output:

1. **`subject.display` carries the patient's full name.** This is the field `CLAUDE.md`'s non-negotiable names specifically — "Real feeds put patient names there" — and it is the single most common shape in a real Observation feed.
2. **`subject.identifier.value` carries an MRN.** So does `contained[].identifier`, and `extension[].valueIdentifier`. Function-level probe:

```python
from r6.redaction import apply_patient_controlled_redaction
r = {"resourceType": "Observation", "status": "final",
     "subject": {"reference": "Patient/p1", "display": "Marisol Rivera",
                 "identifier": {"system": "urn:mrn", "value": "MRN-7749-XYZ"}},
     "contained": [{"resourceType": "RelatedPerson",
                    "identifier": [{"system": "urn:ssn", "value": "000-00-9999"}],
                    "name": [{"family": "Rivera", "given": ["Marisol"]}],
                    "telecom": [{"system": "phone", "value": "+1-555-0100"}]}]}
print(apply_patient_controlled_redaction(r, "hc-123"))
# every one of those values is in the output, next to the ANONYED tag
```

3. **The injected "sole canonical identifier" is the empty string.** `$share-bundle` passes `patient_id or ''` for non-Patient rows, so when the request body omits `patient_id` — which is exactly the whole-tenant export — every Observation and Coverage in the bundle gets `value: ""`. The canonical id identifies nothing on the rows carrying the clinical content. (My request omitted `patient_id`; with one supplied, this row does not apply.)

The new test cannot see any of this: `test_patient_controlled_leaves_only_the_healthclaw_identifier` asserts on a flat fixture, and its `blob` check only holds because that fixture has nothing nested.

**This is the failure mode the PR itself says it is removing** — the old keyword denylist let anything it did not name through with its value intact. Replacing the denylist closed that at the top level and left the same shape one layer down.

**Two ways out, and the choice is the CTO's, not mine:**

- **Narrow the claims.** Docstring and PR body say "the top-level `identifier` list", and the recursive gap is filed against #112 — whose checklist already names it: *"Profile-specific recursive allowlists (deny-by-default) rather than field-name denylists, so a new PHI-bearing field can't leak by omission"* and *"a PHI-canary test suite that asserts **nested** displays, extensions, attachments, URLs, dates, and identifiers are all stripped for each profile."* Cheapest, and honest.
- **Make it recursive.** Larger than it looks: `_redact_recursive` would also truncate the `birthDate` this profile deliberately preserves and strip the `display` values in the `meta.tag` it adds, so ordering and exclusions need designing.

Either is fine. Shipping the current claims is not.

## MEDIUM — "Identifiers: Remove the value" overstates the standard profile

`_redact_fields` keys off `resource['identifier']`. FHIR puts the Identifier datatype under other names, and those keep their values. All of these resource types are in `R6Resource.SUPPORTED_TYPES`, so all are storable and readable through the redacting path:

| element | after `apply_redaction` |
| --- | --- |
| `ServiceRequest.requisition` | value intact |
| `ServiceRequest.groupIdentifier` | value intact |
| `Specimen.accessionIdentifier` | value intact |
| `DocumentReference.masterIdentifier` | value intact |
| `extension[].valueIdentifier` | value intact |
| `Coverage.subscriberId` | value intact |

`Coverage.subscriberId` is the sharpest one: a health plan beneficiary number, Safe Harbor §164.514(b)(2)(i)**(I)** — the paragraph immediately after the two this PR's own reasoning cites. On the wire:

```
GET /r6/fhir/Coverage        (standard profile, apply_redaction)
  "identifier": [{"system": "urn:cov"}],        <- value removed, as intended
  "subscriberId": "HP-000-00-9999"              <- returned in full
```

Also: `_values()` in the new test file is documented as *"Every identifier value anywhere in a resource"* but walks only `identifier` keys. That docstring makes the test read as a universal sweep. Please narrow it — a helper that claims more than it checks is how the next reader concludes this area is covered.

A second reachable route for the HIGH above, so it is not rediscovered as new: `Bundle` is in `R6Resource.SUPPORTED_TYPES`, so `GET /r6/fhir/Bundle/<id>/$deidentify?mode=patient-controlled` passes a stored Bundle whole into `apply_patient_controlled_redaction` — nothing inside any entry is touched, and the Bundle is stamped `ANONYED`.

`r6/health_compliance.py::deidentify_resource` (the `mode=deidentified-preview` default) has the same class of hole — it pops `identifier` recursively but leaks `extension[].valueIdentifier`. Out of scope for this PR; noting it so it isn't discovered as new.

## LOW — `PR_BODY-safe-harbor.md` is committed at the repo root

244 lines, tracked, not on `main`, duplicating the PR description in a public repo where it will silently drift. Please delete it from the branch.

---

## On the mutation evidence

I re-ran your mutation #3 rather than taking it: swapping `_redact_recursive` / `label_codings` in `apply_redaction` does turn `test_label_codings_runs_after_the_redaction_strip` red. Confirmed. I could not find any reordering that keeps it green.

One caveat on the framing, because you called it "what makes it executable instead of a comment nobody re-checks." It pins the **call order**, not the property. I made the strip stash each upstream `display` and restored it after labelling — ordering untouched, so the non-negotiable is violated in exactly the "make records readable" direction `CLAUDE.md` warns about:

```
apply_redaction({"resourceType":"Observation","code":{"coding":[
  {"system":"http://loinc.org","code":"9999-XX",
   "display":"Glucose for Jane Secret"}]}})
-> display "Glucose for Jane Secret" survives into the output
```

Under that mutation: your ordering test **passed**, both new identifier tests **passed**, and `test_guardrail_conformance.py` **passed 37/37 at Grade A**. Five other tests elsewhere in the suite caught it, so the suite is not blind — but the "Grade A confirmed" line in the PR body is weaker evidence than it reads. Grade A does not discriminate this.

Credit where it is due, from the same exercise: dropping `'display'` from `_FREE_TEXT_KEYS` is caught **only** by your new ordering test. Conformance stayed green at 37/37 and the identifier tests all passed. That test earns its place.

## What is right in this PR

- Identifier values are removed, not truncated. Verified across list-shaped, dict-shaped, short (`len <= 4`), and nested-under-`identifier` cases.
- **Non-string identifier values are now removed too.** `main`'s `isinstance(..., str)` guard let `{"value": 4471}` through intact; `pop` does not. That is a real fix the PR body does not claim.
- The seven `'***'` pins were flipped correctly, not deleted.
- `tests/test_deidentification_language.py` was **not** weakened — the assertion's force is unchanged and the rationale was updated to a reason that is still true.
- Doc edits are accurate on every surface I checked, and annotating the two dated evidence packs rather than rewriting them is the right call.
- Audit `detail` is PHI-free on every path I grepped; redaction is read-time, so stored JSON and ingest dedup are untouched.

## What I did not cover

- No live upstream FHIR server (Aidbox / HAPI / Medplum) — SQLite store only.
- No TypeScript or Postgres CI lane.
- `scripts/healthclaw_redact.py` is a hashing profile, not truncation. Noted, not audited.
- I did not test whether any partner integration reads `identifier[0]['value']` and now gets a `KeyError`. The suite says no in-repo caller does.
- Redacted **responses** are not persisted by `ContextItem` (refs and sha256 only), but `AgentToolCall.result_json` stores whole tool results. If a tool ever returned a redacted resource, historical prod rows may still hold `***XXXX` values. I cannot inspect prod from here — whether that needs a backfill or purge is a CTO call.

I have **not** approved and have **not** merged. Note that `claude-standards-review` reported SUCCESS 5 seconds after starting (16:49:04 → 16:49:09) — consistent with #608, so please do not read that green as a review.

I pushed nothing to this branch. The four characterization rows behind the findings above are on a local review branch; say the word and I will open them as their own PR against `main`, or hand them over for you to include.

aks129:
Both findings fixed in a162745. Thank you — the stash-restore attack found a real hole in test discipline, not just this test.

**Ordering test.** Replaced `test_label_codings_runs_after_the_redaction_strip` (which spied on what `label_codings` was *called with*) with `test_no_upstream_display_survives_into_the_redacted_output`, which asserts on `apply_redaction`'s actual return value. Verified both directions independently before committing: applied the exact stash-restore mutation you're describing (strip cleanly, let a call-order spy fire clean, then restore the raw display afterward) to a temporary copy of `apply_redaction` — the old spy assertion (`calls == [None]`) stayed `True` while `"Jane Secret"` was confirmed present in the real return value. The new test catches both that attack and the original call-order swap; a positive assertion (a recognised code still gets labelled) confirms the fix isn't "assert nothing survives."

**Docstring.** Narrowed the module docstring's Safe Harbor claim: it now says explicitly that (G)/(H)/(I)/(J) coverage applies to the `identifier` array shape, names `Coverage.subscriberId` as an example of a field carrying the same category outside that shape which this function doesn't inspect, and points at #112 rather than implying total coverage.

**#617 — not touched here, on purpose.** Confirmed by reading the code directly (not just trusting the finding): `apply_patient_controlled_redaction` pops fields only at the top level of its single dict, `subject`/`contained`/`extension` are never descended into. That's a correctly-scoped separate issue — a redesign question ("why two implementations, should the patient-controlled variant be a policy on the recursive walker") deserves its own review, not a rushed patch bolted onto this PR's remit.

3163 passed, 13 skipped, 1 xfailed. ruff clean.

aks129:
Folded the #617 recursion fix into this PR (6fe876e), per the coordination above — this branch was already rewriting `apply_patient_controlled_redaction`'s body, so a fix built against `main` would collide, and per #585 a fix stacked on this PR's unmerged branch would lose CI (python-tests, secret-scan) while auto-merge still runs, which is not how a PHI-leak fix should land.

**Reproduced the exact scenario independently before writing anything**, then again after the fix, not just trusting the report:

Before — `Marisol`, `Esteban`, `617-555-0199`, `Alice Nguyen`, `Rivera` all present in the output of a Patient with a contained RelatedPerson + subject.display + generalPractitioner[].display, stamped `ANONYED`.

After — all five confirmed absent.

**Fix shape**: run the existing `_redact_recursive` walker (the same one `apply_redaction` already uses) over the whole tree first, closing every nested `display`/name/telecom/address/text at any depth generically, then layer `apply_patient_controlled_redaction`'s own stronger top-level-only policy on top (full field removal rather than redaction, the verbatim original `birthDate` restored, the single canonical identifier). A policy on the one walker, per the fix shape suggested — not a second walker.

**Verified nothing pinned on the old top-level-only behavior broke**: all 152 existing tests across every file touching this function pass unchanged. Conformance harness: 37 passed, Grade A held. Full suite: 3164 passed.

**Mutation-checked**: removed the `_redact_recursive(result)` call — the new test's first assertion goes red (`Marisol` survives). Restored, confirmed clean.

aks129:
Independent verification of 6fe876e, from the session that filed the #617 reproduction. I did not re-run the author's tests; I ran my own probe against the branch in a detached worktree, on the principle that a PHI fix should be confirmed by someone who did not write it.

**The reported case, re-checked:**

```
top-level name removed : True
meta tags              : ['ANONYED', 'patient-controlled']
  LEAKED 'Marisol': False      'Esteban': False      '617-555-0199': False
  LEAKED 'Alice Nguyen': False 'Rivera': False
```

All five values that leaked on `main` an hour earlier are gone, and the anonymized stamp and top-level strip are unchanged.

**Then four cases nobody handed the author,** because fixing the reported scenario and fixing the class are different things:

```
[clean] contained inside contained            two levels deep
[clean] CodeableConcept.text                  "Diabetes - patient Marisol Rivera"
[clean] performer/requester display + note.text carrying a phone number
[clean] Bundle of entries                     the case the original report flagged
[clean] birthDate preserved + nested name gone
```

The last row is the one that matters most, and the one a recursion fix most easily breaks: `1962-03-04` survives while a nested RelatedPerson name is removed. The docstring promises birthDate is preserved because the patient wants their own DOB, and the verbatim-restore ordering keeps that true.

**On the approach.** Running the existing `_redact_recursive` walker over the tree and layering this function's stronger top-level policy on top is the right shape. A second walker would have been two policies to keep in sync, and the next `display`-bearing field added upstream would have landed in only one of them — which is how the original defect got here.

**One thing this does not change.** #617 remains live on `main` until this merges. A fix in an open PR closes nothing, and nothing has merged in three days. That makes this the strongest merge case in the queue: most open changes are correctness or honesty work, and this one is an active PHI exposure on the sharing path.

Worth stating at the top of the body that this now closes #617, so it is not read as a Safe Harbor cleanup and queued behind fifty others.

Not approving and not arming auto-merge — that gate is the owner's.

Issue #626, opened 2026-09-04
four hostnames we publish do not answer, two of them from running code


aks129:
Found by the deployed-surface sweep that generalises #624
(`docs/evidence/2026-09-04-deployed-surface-inventory.md`).

Four hostnames this repository publishes do not answer. Two are documentation,
which costs a reader their time. Two are served by running code, where the
reader is a machine.

Each was probed once, read-only, on 2026-09-04.

## Served by running code

**`schemas.agentskills.io` — NXDOMAIN.** `app.py:339` sets
`_SKILLS_DISCOVERY_SCHEMA` to
`https://schemas.agentskills.io/discovery/0.2.0/schema.json`, and that value
goes into the `$schema` field of the live discovery document at
`https://healthclaw.io/.well-known/agent-skills/index.json` (200, confirmed).
The parent zone `agentskills.io` resolves; the `schemas` label does not.

`tests/test_wellknown_skills.py:21` asserts the URL's prefix, so the test passes
on a URL no client can fetch.

**`sharponmcp.com` — NXDOMAIN at the registry.**
`services/agent-orchestrator/src/index.ts:259,704` returns
`spec: "https://sharponmcp.com"` in the MCP `initialize` response. Also linked
from `templates/wiki.html:562` and `templates/faq.html:328`.

Neither is a security problem, and neither breaks a client that does not
dereference the URL. Both are claims our servers make that no longer resolve.

## Documentation

**`shl.healthclaw.io` — 404 from Vercel.**
`skills/share-health-qr/SKILL.md:214,215,226,230` shows viewer and manage links
on that host as the skill's output.

**`healthclaw.up.railway.app` — 404, `Application not found` (railway-hikari).**
`skills/personal-health-records/SKILL.md:108,132` tells a reader to configure
`https://healthclaw.up.railway.app/mcp`; `:158` gives a `curl -X POST` against
`/r6/fhir/internal/seed` on it; `:171` sets it as `BASE`.
`scripts/build_quickstart_pdf.py:485` prints the same URL into the generated
quickstart PDF, so it is reproduced every time that runs.

Both live MCP endpoints are in `server.json`. A reader following either skill
today configures a name that hosts nothing.

## Already tracked

`mcp.healthclaw.io` also 404s from Vercel and is already written up in
`docs/specs/2026-08-16-mcp-authorization.md` §9.6 as a dangling record. Listed
here only so this issue covers the whole set the sweep found.

## Reproduction

```bash
[...]

Issue #625, opened 2026-09-04
the production monitor covers four of the six live hosts we own, and two MCP builds are behind main


aks129:
Found by the deployed-surface sweep that generalises #624
(`docs/evidence/2026-09-04-deployed-surface-inventory.md`).

`scripts/prod_watch.py` requests 11 URLs across **4 hosts**. Counting distinct
live hosts we own, this repository names **6**, plus the retired box in #624
that no name resolves to. So the monitor covers 4 of 6, and none of the 7th.
Three of the gaps matter.

## 1. `careagents.cloud` is the host users type, and no check requests it

Every check named `careagents:` requests `careagents-production.up.railway.app`.
`CARE_ORIGIN` and `CARE_RP_ID` are `careagents.cloud`, so that is the host a
person's browser shows and the origin WebAuthn binds to.

Today both return the same `build` and `built_at`, so the domain is a custom
name in front of the same Railway service rather than a second instance. The
gap is narrow, not empty. This repo already records what happens when a
domain-level failure is invisible to a service-level check: on 2026-08-06 a
Vercel domain list omitted the hostname the project served, the serving project
was deleted, and production went down
(`docs/2026-08-06-two-generators-three-laws.md:41`). Under that shape of
failure all four `careagents:` checks pass, the workflow stays green, and
nobody can sign in.

`scripts/prod_watch.py:60` names the constant `CAREAGENTS` and sets it to the
platform hostname. Reading the constant name would not reveal this; the sweep
found it by recording what the module actually requests.

## 2. `healthclaw.io` is a first-class host in the code and unwatched

`deployment.py:1-20` describes two hosts that "are not interchangeable" and
names `healthclaw.io` as the second. It answers 200. It serves the public site,
the skills catalogue, and the `r6-dashboard` the README sends visitors to. No
check requests it.

Its Vercel purpose is an open owner decision
(`docs/2026-08-16-system-topology.md`, "purpose unsettled"). Unwatched is a
defensible answer for a host nobody has decided to keep. It is not a defensible
answer by default, which is what it is today.

## 3. Both MCP deployments run an older build than `main`, and no check asks

```
$ curl -s https://mcp-demo-production-ee2c.up.railway.app/health
{"status":"healthy","version":"1.9.0", ...}
$ grep '"version"' services/agent-orchestrator/package.json
  "version": "1.10.0",
```

`/health` reports `SERVER_VERSION`, read from that package.json
[...]

Issue #624, opened 2026-09-04
an abandoned CareAgents instance is still serving on the old VPS, with an accounts store and no monitoring


aks129:
Found while settling whether the VPS deploy path is still live. It is not — but **the VPS is**.

## What is running

```
curl --resolve careagents.cloud:443:187.77.4.50 https://careagents.cloud/healthz
{"accounts":true,"provider":"openai","status":"ok"}
```

A CareAgents instance, answering, with an accounts store. The payload has no `build`, no `built_at` and no `run_workers`, which means it predates the build-stamping and worker changes — so it is **months behind `main`**.

Public DNS for `careagents.cloud` points at the platform, not here, so no user is routed to it. It is reachable by address with the right hostname presented.

## Why it matters

- It holds **accounts** — email addresses and passkey credentials, whatever existed when it was last deployed.
- It presumably still holds whatever credential lets CareAgents talk to the engine. If that secret is shared with production, an old box nobody watches holds a working key.
- It has **none** of the guardrail work done since: not the redaction fixes, not the payload seal on approved actions, not the honesty corrections, not the beta record switch. Whatever its behaviour was months ago is its behaviour now.
- **Nothing watches it.** The production monitor checks the platform host only. A compromise, a failure, or a data change there produces no signal anywhere.

None of this is urgent in the sense of being under attack. It is quiet rather than safe, and quiet is how the previous incident here happened — two deployments serving code months apart while the monitor reported everything green, because nothing it checked could tell them apart.

## What only the owner can do

1. **Decide whether it is shut down or kept.** If kept, it needs to be watched and updated like anything else serving account data.
2. **Rotate anything it shares with production**, particularly the engine credential. A key on an unwatched host is a key you cannot reason about.
3. **Decide what happens to the account data on it** — whether it is exported, deleted, or reconciled against the live store.

Nothing here can be done from the repository. PR #623 retires the deploy path so nothing new is shipped there, which stops the situation getting worse but does not address the box.

## Related

#264 called for this cutover and for retiring the path. The cutover happened; the retirement reached neither the code nor the documents until now, and the machine was never dealt with.

PR #627, opened 2026-09-04
chore(evidence): inventory every deployed surface, and what watches it (#624)

aks129:
Generalises #624. That issue was an abandoned CareAgents instance still serving
on an old VPS: months behind `main`, holding an accounts store, watched by
nothing. It surfaced because somebody happened to ask whether a deploy script
was still used. Nothing in this repository could answer "what else is like
that?", so this adds the thing that can, and runs it once.

## What is here

- `scripts/surface_inventory.py` — enumerates every host the repository names
  as a deployed surface, probes each one read-only, and sorts them into **live
  and watched**, **live and unwatched**, and **referenced but dead**.
- `docs/evidence/2026-09-04-deployed-surface-inventory.md` — the run, dated and
  pinned to a commit, with the exclusion list and the coverage gaps stated.
- `docs/evidence/2026-09-04-surface-inventory-check-count.py` — the
  reproduction for one finding about the monitor.
- `tests/test_surface_inventory.py` — guards on the tool's own claims.

## Two rules it follows rather than assumes

**The watched set is measured.** It imports `scripts/prod_watch.py`, replaces
`get`/`post` with recorders, runs it offline, and takes the URLs it actually
requested. Reading the constants would have reproduced the error it went
looking for: the checks named `careagents:` measure a platform hostname, never
the domain users type.

**DNS is resolved twice.** Over DoH and through the system resolver, with any
disagreement reported. This network answers port 53 from a stale cache, and for
`careagents.cloud` a wrong answer would have flipped the #624 finding in either
direction.

## What the run found

Measured 2026-09-04 at `89b42fb`. 25 endpoint rows: 4 watched, 16 live and
unwatched (4 ours, 12 third-party), 5 dead. Counted as distinct live hosts we
own, that is 6, of which the monitor requests 4, plus the retired #624 box that
no name resolves to.

1. **The #624 box still answers**, with an accounts store and no build marker.
   Confirmed from the other direction, by enumeration rather than by accident.
   Owner decision, unchanged: #624.
2. **Both MCP servers run `1.9.0`; their `package.json` on `main` says
   `1.10.0`.** Four checks cover those two hosts and all four pass, because
   none reads the version field the response already carries. The #258 shape,
   still open on a different pair of deployments. → #625
3. **`careagents.cloud` is unwatched.** It is `CARE_ORIGIN` and `CARE_RP_ID`,
   the host a browser shows and WebAuthn binds to. Today it fronts the same
   Railway service as the watched hostname, so the gap is narrow. The 08-06
   incident is the shape that opens it: a domain list omitted the hostname the
   project served and production went down while the service was fine. Under
   that, all four `careagents:` checks pass while nobody can sign in. → #625
4. **`healthclaw.io` is unwatched**, while `deployment.py` treats it as one of
   two hosts that "are not interchangeable". → #625
5. **Four published hostnames do not answer**, two of them served by running
   code: the `$schema` in our live agent-skills discovery document, and the
   `spec` URL in the MCP `initialize` response. → #626
6. **Two honesty defects in `prod_watch`.** `all {len(results)} checks passing`
   counts what ran, so the same stubbed deployment reports 11 or 12 depending
   on the flag. And check names carry a product while targets carry a hostname.
   → #625

## Scope

No production code is changed and no finding is patched. Each is a decision
with a cost: watching `healthclaw.io` means deciding to keep it, and asserting
an MCP version means deciding the accepted set for a deployment that is manual
by design. Evidence and reproductions are in the document and the issues.

## Probe discipline

One GET per endpoint, `allow_redirects=False`, no retries, no credential sent,
no method other than GET. 14 of the 25 rows are other people's servers.

## Redaction

The retired VPS is addressed by IP in `deploy/careagents/deploy.sh`. The script
reads that address at run time and scrubs it from every line it prints, so the
report can say whether a name resolves to it without restating it. The diff
carries no address, credential, or local path; `tests/test_surface_inventory.py`
pins the scrub.

## Gates

```
uv run ruff check .          All checks passed!
                            3165 passed, 13 skipped, 1 xfailed
```

Both new tests fail without their fixes (verified by reverting each): the
reference scan silently scanned nothing when the checkout sat under a skipped
directory name, and deriving the watched set printed the monitor's own eleven
stubbed failures at the top of the report.

A second commit on this branch corrects five claims that did not survive
review: a check miscounted three as four, two `prod_watch` line numbers that
were estimates rather than `grep -n` output, three citations to a gitignored
file, an overstated reproduction, and a test that asserted a defect as
specification. Its message says what each was.

## Known gaps, stated rather than papered over

Only port 443, only names this repository writes down, nothing behind
authentication. A service on another port or on a host nobody wrote down is
invisible to this method. That is the residual #624 risk and this PR does not
close it. The Mac mini iMessage relay holds a mint secret and has no public
name here; it was not probed.

### DIFF
--- a/docs/evidence/2026-09-04-deployed-surface-inventory.md
+++ b/docs/evidence/2026-09-04-deployed-surface-inventory.md
@@ -8,6 +8,10 @@ repository names as deployed, read-only. `scripts/prod_watch.py` requests 4 of
 them. 16 answer and nothing checks them (4 ours, 12 third-party). 5 are named
 here and do not answer.
 
+Those are endpoint rows, and one host can carry two of them. Counted as
+distinct live hosts we own: 6, of which the monitor requests 4. The retired box
+in #624 is a 7th that no name resolves to.
+
 #624 (an abandoned CareAgents instance on an old VPS) was found by accident,
 when somebody asked whether a deploy script was still used. This exercise asks
 the question that accident raised: what else is out there? The answer includes
@@ -29,7 +33,8 @@ output directly.
 ### On the DNS answers
 
 Every name resolved twice, over DoH and through the system resolver, because
-this network answers port 53 from a stale cache (`CLAUDE.md`). The two agreed
+this network answers port 53 from a stale cache (operator notes, not published
+in this repository). The two agreed
 for every host except `healthclaw.io`, `mcp.healthclaw.io` and
 `shl.healthclaw.io`, where the address sets differ. All three sit behind
 Vercel, and both answers reached Vercel, so the likeliest cause is anycast or
@@ -159,10 +164,11 @@ decide what happens to the account data.
 `1.10.0`. Both deployments are therefore behind the source, by at least the
 change that raised the version.
 
-Three checks in `prod_watch` speak for these two hosts. All three pass. Two ask
-whether `/health` returns 200 and one asks whether the demo server answers an
-unauthenticated handshake. None reads the version field the response already
-carries, so all three are satisfied by any build at all. This is the #258 shape
+Four checks in `prod_watch` speak for these two hosts, and all four pass. Two
+ask whether `/health` returns 200. One asks whether the token-locked server
+refuses an unauthenticated caller with 401. One asks whether the demo server
+answers an unauthenticated handshake. None reads the version field the response
+already carries, so all four are satisfied by any build at all. This is the #258 shape
 exactly: the CareAgents build check exists because every other check was
 equally satisfied by a months-old build, and the same hole is still open on the
 MCP servers.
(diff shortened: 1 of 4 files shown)


aks129:
One row in "referenced but dead" is a different kind of thing, and the class
name is what will make the next sweep re-file #626.

`schemas.agentskills.io` is not an address the discovery document points at. It
is the version identifier the format requires. From the Cloudflare Agent Skills
Discovery RFC v0.2.0, section Versioning
([README.md:175](https://github.com/cloudflare/agent-skills-discovery-rfc/blob/main/README.md)):

> The `$schema` URI is an **opaque identifier**. Clients MUST match it against
> known schema URIs to determine how to process the index. The URI does not
> need to be resolvable, though it MAY point to a JSON Schema document that
> describes the index format.

The RFC names the exact string we serve, and NXDOMAIN is the state it expects.
A client that receives a resolvable substitute stops recognising the index
(line 177); one that receives no `$schema` at all reads it as v0.1.0.

So the probe result is right and the classification is wrong. Sorting it with
`shl.healthclaw.io` and `healthclaw.up.railway.app` reads as four defects where
there are three. The suggestion is a class for names a document cites without
dereferencing, whether that is an entry in the exclusion list or a third state
next to live and dead. Either keeps the report's claim true, which is the
property the rest of this PR is careful about.

#628 handles the `app.py` half of #626 on that basis: the value stays, the
citation goes next to it, and the prefix assertion in
`tests/test_wellknown_skills.py` becomes an exact pin so a well-meaning repair
to a host that answers turns a test red. It cannot touch
`scripts/surface_inventory.py`, which only exists on this branch.

Issue #635, opened 2026-09-05
The default test suite calls the public internet at import time, and its skip guard is a point-in-time check


aks129:
Surfaced as an incidental during the #634 sweep: `tests/test_public_fhir_servers.py::TestFlaskGuardrailsWithHAPI::test_observation_search_via_route` failed once on a **verified-clean tree** and passed on immediate re-run. Chased it down; the mechanism is structural, not a one-off.

`tests/test_public_fhir_servers.py:38-45`:

```python
def _server_reachable(url: str) -> bool:
    resp = httpx.get(f'{url}/metadata', params={'_summary': 'true'}, timeout=10)
    return resp.status_code == 200

# Check reachability once at module load
_hapi_available  = _server_reachable(HAPI_FHIR_R4)      # https://hapi.fhir.org/baseR4
_smart_available = _server_reachable(SMART_HEALTH_IT)   # https://r4.smarthealthit.org

skip_hapi = pytest.mark.skipif(not _hapi_available, reason='HAPI FHIR R4 server unreachable')
```

Three separate problems, and only the first is obvious.

**1. Every `pytest` invocation in this repo makes two live HTTP calls, whether or not these tests run.** The probes are module-level, so they fire at collection. Measured today:

```
pytest tests/test_public_fhir_servers.py --collect-only     31 tests collected in 2.47s
pytest tests/test_public_fhir_servers.py -k no_such_test    31 deselected in 2.08s
```

The second is the one that matters: **every test in the file was deselected and the network calls still happened.** With both hosts unreachable that is a 20-second floor (two 10s timeouts) added to every run, including a developer running one unrelated test, and including CI on a runner with no outbound access.

**2. The skip guard is named for a property it does not have.** `skip_hapi` reads as "this test runs only when the server is available." What it actually encodes is "the server answered once, before the run started." A server that is up at collection and flaky sixty seconds later produces a **failure**, not a skip. That is the observed flake, and it is the exact defect shape #630 and #634 are about — a control reporting on something other than what its name says. Worth noting it slipped past two censuses because it lives in a file neither swept.

**3. `hapi.fhir.org` is a shared public sandbox that anyone can write to.** Reachable is not the same as correct. Assertions against a mutable resource under nobody's control will eventually fail for reasons that have nothing to do with this repository, and the failure will land on whoever pushed last.

[...]

Issue #634, opened 2026-09-05
Security-file guard census: 8 more demonstrated gaps, incl. an SSRF enforcement point that deletes with the suite green


aks129:
Second sweep for the #630 defect shape — a guard that pins a *phrase* rather than a *property* — across the eight security-shaped test files #630 listed as never examined.

**Result: 8 demonstrated gaps across 5 files; 3 files sound.** Every finding was produced by mutating production code, running the named guard, and running the **full** suite. Nine full-suite runs, eight byte-identical to baseline (`3157 passed, 13 skipped, 1 xfailed`).

Harness closed both verified traps: `PYTHONDONTWRITEBYTECODE=1` with a `__pycache__` purge before every run, and every mutation applied through a helper that **asserts the replacement count** before writing. No running server was mutated.

---

## F3 — the SSRF guard has one enforcement point, and deleting it is invisible. Highest blast radius.

`tests/test_ssrf_upstream.py` has nine tests. All nine import `validate_upstream_url` and `_is_blocked_ip` and call them directly. **Not one uses a test client or touches a route.** So the file proves the predicate is correct and nothing proves it is ever *called*.

I verified this one myself rather than relay it. Mutation at `r6/fhir_proxy.py:988`, inside `get_proxy_for_request`:

```python
if False and server_url and not valid:     # was: if server_url and not valid:
```

```
tests/test_ssrf_upstream.py + test_sharp_on_mcp.py + test_fhir_proxy.py   81 passed
full suite                                                                3157 passed
```

Behavioural probe, both directions, `X-FHIR-Server-URL: http://169.254.169.254/latest/meta-data/` with `X-FHIR-Access-Token`:

```
MUTATED : FHIRUpstreamProxy   upstream_url = http://169.254.169.254/latest/meta-data
CLEAN   : None
```

A proxy is constructed against the cloud-metadata endpoint, over plain HTTP, with the caller's SMART token attached to its client. The mutation also defeats the https requirement, so the token would be on the wire in clear. The probe stops at construction; no request was fired at `169.254.169.254`.

**Ranking note, surfaced rather than resolved:** the sweep ranked by the stated non-negotiables as instructed, which puts F1 first. By blast radius F3 leads — it is upstream credential exposure. That ordering call is the CTO's.

## F1 — `/Bundle/$ingest-context` emits no AuditEvent that anything checks. Both paths.

Non-negotiable: *"Every FHIR resource access emits an AuditEvent."*

[...]

Issue #633, opened 2026-09-05
Make the MUTATION docstrings executable, because care alone has now failed three times


aks129:
This repository writes `MUTATION: <edit> -> red` lines in test docstrings as a discipline. It is a good discipline. It is also, right now, a set of unverified claims, and #630 measured the cost: **four of eight guards carried a `MUTATION:` line naming an edit that did not turn them red.**

## Why this needs a mechanism rather than more care

Three independent failures, in increasing order of how much they should worry us:

1. **The original decay.** Four `MUTATION:` lines were written, believed, and never re-run. Nobody was careless; the lines simply have no way to fail.
2. **A census about unverified claims made one.** #630 asserted that mutating with pytest instead of a running server "sidesteps the stale-process trap entirely". It does not — CPython invalidates bytecode on `(mtime, size)`, so a same-byte-length edit inside one clock second is silently ignored. That was my sentence.
3. **The strongest signal.** The PR repairing unexecuted `MUTATION:` lines shipped, in its first commit, two `MUTATION:` claims its author had not executed. They caught it themselves and the second commit runs both — but the defect reproduced *inside its own fix*, by someone actively thinking about it.

When a discipline fails for the person whose entire task is that discipline, the discipline is not the fix.

## What exists already

The work on #632 built a mutation harness in the scratchpad that does the right things: `PYTHONDONTWRITEBYTECODE=1`, a `__pycache__` purge either side, an **asserted replacement count** so an edit that silently does not apply is caught, a re-read from disk proving the edit landed, byte-restore rather than `git checkout --`, and an asserted-clean `git status` afterwards.

Every one of those guards against a specific false verdict we have actually produced. It was thrown away with the worktree, because nobody asked for it to ship. Rebuilding it per-agent is how the traps keep getting rediscovered.

## The shape of the ask

Deliberately not specifying the design, because the interesting question is one for whoever picks it up: **a prose `MUTATION:` line cannot be executed, so what replaces it?** A structured marker naming the file, the exact replacement, and the test expected to redden is the obvious answer, and the obvious answer has a real cost — it is more ceremony on every guard, and ceremony is what decays.

Worth weighing:

- A machine-readable form checked by a periodic job, rather than in the main suite. Running every mutation on every push would be slow and would make CI fail for reasons unrelated to the change under test.
- Whether it covers all such lines or only guards protecting the stated non-negotiables. Full coverage is the honest goal; partial coverage that actually runs beats full coverage that does not.
[...]

PR #636, opened 2026-09-05
test(guards): pin that the SSRF check is reached, and that ingest is audited (#634)

aks129:
Two guards from the #634 census. Both close a gap where **the control works today and nothing would notice if it stopped.** No production code is changed.

## F3 / F5 — the SSRF guard's enforcement point

`tests/test_ssrf_upstream.py` had nine tests. All nine call `validate_upstream_url` directly. None of them checks that anything ever *invokes* it, and for a guard that is the load-bearing half: a correct rule nobody applies refuses nothing. `get_proxy_for_request` has exactly one enforcement point.

Disabling it leaves the whole suite green, and a request naming the cloud-metadata endpoint then builds a live proxy at it, over plain HTTP, with the caller's SMART token on its client:

```
MUTATED : FHIRUpstreamProxy   upstream_url = http://169.254.169.254/latest/meta-data
CLEAN   : None
```

**On the shape, since there was a real choice here.** An AST assertion that the call site exists would be cheaper. I did not write one: a test that greps for the call pins a *spelling*, and pinning a spelling instead of a property is the exact defect this file was found to have. Fixing it with more of it would be absurd. The new class drives the resolver and pins the outcome.

## F1 — `/Bundle/$ingest-context` audit trail

`tests/test_ingest_auth.py` pinned the three credential states and never looked at the audit trail. Both `record_audit_event` calls on that route could be deleted with the suite green, against the standing rule that every FHIR resource access emits an AuditEvent.

The refused case is the one that matters most: a credential-probing loop against a private tenant left **no trace at all**.

## Non-vacuity, deliberately

Both audit tests count rows before and after rather than asserting a row exists, so they cannot pass on rows another test left behind. The SSRF class carries an explicit non-vacuity control for the same reason — without it, the four refusal assertions would pass if proxying were simply broken.

That is not a hypothetical. It is how #634 F2 stayed green: three leak assertions passed *vacuously* because nothing was captured at all, and only its sibling's positive assertion caught the real leak.

## Four mutations, each executed

Each reddens only its own half. The two pairs pin two separate properties and neither stands in for the other, so both are recorded in the docstrings rather than the one I happened to run first.

| Mutation | Result |
|---|---|
| `if server_url and not valid:` → `if False and …` | 4 × `builds_no_proxy` fail |
| `return valid` → `return bool(_raw)` in `is_sharp_context_active` | 4 × `does_not_activate_sharp_context` fail |
| delete deny-path `record_audit_event` (`r6/routes.py:1251`) | `test_a_refused_ingest_is_audited` fails |
| delete success-path `record_audit_event` (`:1269`) | `test_an_accepted_ingest_is_audited` fails |

Clean runs: 18 passed (ssrf), 3 passed (ingest).

Every mutation was applied with `PYTHONDONTWRITEBYTECODE=1`, a `__pycache__` purge either side, an **asserted replacement count** before writing, and a revert verified with `git diff --quiet`. That harness is not ceremony — see #633, where three separate failures came from exactly the traps it closes.

## Tests

```
uv run ruff check .                            All checks passed!
PYTHONDONTWRITEBYTECODE=1 pytest tests/ -q     3168 passed, 13 skipped, 1 xfailed in 274.74s
```

3157 baseline + 9 + 2 = 3168.

An earlier run of this branch showed `1 failed` — `TestSMARTHealthITProxy::test_patient_search`, a live-network test unrelated to this change and now recorded on #635, which it corroborated on a different host than the one that prompted that issue. It passes in the run above.

## Not included

**F8 is deferred, not forgotten.** `tests/test_metadata_security.py` is claimed by #553, and a second change in a contended file is not worth the conflict. F2, F4 and F6 from the census remain open.

Not approved, not merged, auto-merge not armed.

### DIFF
--- a/tests/test_ingest_auth.py
+++ b/tests/test_ingest_auth.py
@@ -1,5 +1,23 @@
-"""Bundle ingestion is a clinical write and requires write authorization."""
+"""Bundle ingestion is a clinical write: it requires write authorization,
+and it leaves an audit trail.
 
+The authorization half was already pinned by the three credential states
+below. The audit half was not pinned by anything: both `record_audit_event`
+calls on this route could be deleted with the full suite green (#634 F1),
+against the standing rule that every FHIR resource access emits an
+AuditEvent. A rejected write is the case that matters most — a credential
+probe against a private tenant left no trace at all.
+
+The two tests at the bottom count rows before and after rather than
+asserting a row exists, so they cannot pass on audit rows some other test
+left behind.
+
+MUTATION: delete the `record_audit_event` call on the deny path
+(r6/routes.py:1251) -> the deny test fails; delete the one on the success
+path (:1269) -> the success test fails. Both executed 2026-09-05.
+"""
+
+from r6.models import AuditEventRecord
 from r6.stepup import generate_step_up_token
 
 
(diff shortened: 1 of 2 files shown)


aks129:
Added **F4** to this branch (`3f49c1c`) rather than opening a second PR — it is the same file, and a separate PR touching `tests/test_ssrf_upstream.py` would only conflict with this one.

**The gap.** All nine original tests pass a literal IP or a malformed string, so `validate_upstream_url`'s resolve-then-check branch was never executed by anything. Its docstring claims *"Hostnames are resolved and EVERY resolved IP is checked"*, and **EVERY** is the word doing the work.

That needed two tests, because "the branch runs at all" and "the branch checks every address" are separate claims:

- `https://localhost/fhir` — a hostname rather than a literal, so it takes the path the nine tests skip.
- A substituted resolver answering with a public address **first** and an internal one second. That is the realistic shape — split-horizon DNS, or a domain the caller controls — and checking only the first answer accepts the host.
- A non-vacuity control: an all-public host must still be *allowed*, or the two refusals prove nothing.

**Mutation executed:** replacing the `for info in infos:` loop with `return True` reddens both detectors and leaves the control green. The control staying green is correct, not a miss — it asserts `True`, which the mutation always returns.

**Neither test touches the network.** `localhost` resolves from the hosts file, and the multi-address case replaces `getaddrinfo` outright. Deliberate: #635 is open about the default suite reaching public FHIR servers, and adding a network-dependent test to close an SSRF gap would have been its own small joke.

```
PYTHONDONTWRITEBYTECODE=1 pytest tests/ -q    3171 passed, 13 skipped, 1 xfailed
uv run ruff check .                           All checks passed!
```

3168 + 3.

This PR now closes **F1, F3, F4 and F5** of #634. F2 and F6 are in progress elsewhere; F8 stays deferred behind #553.

PR #555, opened 2026-09-03
candidate: /curatr says what it checked, not what it did not (#458)

aks129:
## What

`/curatr` now states the scope of what it graded, on **all three** surfaces that produced a verdict.

- **`r6/curatr.py:864`** — the engine's clean-record summary:

  > No coding or structural issues found in this one *{type}* record. This checked one resource, not your record; duplicates and other record types were not examined.

  `{type}` is the resource type actually evaluated, not a constant. Fixed in the **engine only**, so the `$curatr-evaluate` facade route inherits it and `r6/routes.py` is untouched (its 3927-line ratchet pin stands; the file is 3924).

- **`services/agent-orchestrator/src/tools.ts:1849`** — the MCP rollup carried its own second copy. `_mcp_summary.note` is now:

  > No coding or structural issues found in this one *{resourceType}* record. This checked one resource, not the patient's record; duplicates and other record types were not examined.

  and the issues-found branch names the scope too (`Found N issue(s) in this one {resourceType} record. …`). `next_steps` no longer says *"No action needed — data quality looks good."* — that is a claim about the record, which one clean resource is no evidence for. It now reads:

  > - Tell the patient this checked one resource, not their whole record
  > - Do not describe the record as free of duplicates — that was not checked
  > - Call curatr.evaluate on other resources to cover more of the record

  **This is the surface the clinician demo actually uses** (Claude → demo MCP server), so it is the copy a physician sees.

- **`openclaw/bot.py:634`** — the headline carries the scope *beside* the score on one line:

  > `*Curatr Evaluation* — Checked 1 of 12 Observations (most recent) — quality: good (score: 1.0)`

  N comes from the search Bundle's own `total`. With no total: `Checked 1 of ? Observations (most recent, count unknown)` — N is never invented. The bare `quality: good (score: 1)` line, readable as a verdict on the record, is gone.

- Tests: `tests/test_curatr_says_what_it_checked.py` (new, 5) and `services/agent-orchestrator/src/tools.test.ts` (+2).

**`grep -rn "No data quality issues" .` now returns nothing** — there was no third copy, and both known ones are fixed.

## Why

Closes the honesty half of #458 (council ruling **D14**).

The physician advisor rehearsing the launch demo got "quality: good (score: 1), no data quality issues found" on a tenant holding about a dozen duplicate Type 2 diabetes Conditions. Both halves were true of what ran and false of what they claimed.

[...]

PR #641, opened 2026-09-05
chore(site): the tests-passing figure was 2812; the suite passes 3168 today

aks129:
## What

One number in `templates/index.html`: the "Tests passing" figure goes from 2812 to 3168.

## Why now

The stale-number guard in `tests/test_site_version_sync.py` requires the figure to sit between the count of `def test_` functions under `tests/` and twice that count. On `main` today that floor is 2554, so 2812 passes.

Merging the open queue in the planned order adds test functions. On an integration branch of every open PR (2026-09-05), the floor passes 2812 at the merge of `fix/458-curatr-says-what-it-checked` (#555) and ends at 2877. From that point every later PR fails this guard on "update branch", for a reason that has nothing to do with the PR.

3168 is what the full suite passed on `main` at 89b42fb this week (`3168 passed, 13 skipped, 1 xfailed`). It is inside the band on `main` now and after the whole queue lands.

## Verification

`uv run pytest tests/test_site_version_sync.py` on this branch: 6 passed.

## Merge order

Right after #544 (the lockfile fix), before anything that adds tests.

### DIFF
--- a/templates/index.html
+++ b/templates/index.html
@@ -95,7 +95,7 @@
       <div class="spec__label">FHIR resource types</div>
     </div>
     <div class="spec">
-      <span class="spec__num" data-target="2812">0</span>
+      <span class="spec__num" data-target="3168">0</span>
       <div class="spec__label">Tests passing</div>
     </div>
     <div class="spec">


aks129:
Correction to the commit message: the floor passes 2812 at the forty-first merge in the planned order (#555, `fix/458-curatr-says-what-it-checked`), not "the fourteenth-or-so". The PR body has it right. Measured by counting `def test_` under `tests/` at each merge commit of the integration branch.

PR #657, opened 2026-09-06
fix(ingest-context): validate a caller-supplied resource id the way the Fasten ingester does (#548)

### LINKED ISSUE #548: $ingest-context stores a blank resource id as "" — bypasses the id validation the Fasten ingester has (#286 on a second path)
Found while pinning #286 (PR #546).

`r6/context_builder.py:85` reads the id as `resource.get('id', str(uuid.uuid4()))`. A resource carrying `"id": ""` therefore keeps `""` as its id and is stored under `(tenant, type, "")`; a second blank-id resource of the same type then upserts over the first. The Fasten path (`r6/fasten/ingester.py:401-408`) refuses the same input as `invalid_id` after #546, so the two ingest paths disagree.

Same shape as #286 and #509-2: a write that succeeds and quietly does the wrong thing.

Proposed: one guard in the context builder mirroring the ingester's (blank / non-string → refused with an OperationOutcome issue naming the entry index; absent → UUID; int → str), and a pin. Playbook C2 (`to_canonical()`) is the long-term single owner of ids; this is the interim.

Refs #286, #509. Council ruling D8 (docs/2026-09-02-council-ruling.md).

aks129:
Closes #548.

## What was happening

`r6/context_builder.py` read the id as `resource.get('id', uuid)`, so a resource carrying `"id": ""` was stored under `(tenant, type, "")`, and the next blank-id resource of the same type upserted over it: a write that succeeded and quietly did the wrong thing, the #286 shape on a second path. The Fasten path refuses the same input as `invalid_id` after #546, so the two ingest paths disagreed.

## What changes

The context builder applies the ingester's rule with the ingester's pattern: a present id must be a string or an int matching the FHIR id shape. Blank, bool, float, list or object refuses the **whole bundle** (the builder was already atomic for its other validation errors) with an OperationOutcome naming the entry index and never echoing the value. An int is stored as its string. An absent id still gets a UUID. The interim until the playbook's `to_canonical()` owns ids.

## Evidence

- **Pins first** (`tests/test_ingest_context_refuses_a_blank_id.py`, red on main: 9 of 11): seven bad shapes refuse with `entry 0` named and nothing stored; a blank id in entry 1 refuses the bundle and the good entry 0 is not stored either; int id stored as its string; absent id gets a UUID; two blank-id posts no longer collapse into one row.
- **Mutation**, executed 2026-09-06: put the old read back, the nine go red again.
- Context builder, ingest resilience, routes, write-guard matrix, healthex import and conformance suites green (390 passed). Full suite on the branch: 3672 passed, 13 skipped, 1 xfailed.

### DIFF
--- a/r6/context_builder.py
+++ b/r6/context_builder.py
@@ -63,11 +63,11 @@ class ContextBuilder:
         # medications in it (#377). Imported here rather than at module level
         # because `r6.fasten` pulls in its blueprint, which imports this
         # module back.
-        from r6.fasten.ingester import safe_skipped_type
+        from r6.fasten.ingester import _RESOURCE_ID_PATTERN, safe_skipped_type
         skipped_types: collections.Counter = collections.Counter()
 
         try:
-            for entry in entries:
+            for index, entry in enumerate(entries):
                 resource = entry.get('resource')
                 if not resource:
                     continue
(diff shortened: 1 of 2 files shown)


aks129:
## Reviewed, one correction to make: the pattern is the column width, not the FHIR id shape

Probed the branch rather than reading it. The two things I most expected to be wrong are right: `resource_type` reaches the refusal message only after `is_supported_type`, so the message cannot echo caller text, and the whole-bundle rollback does undo an earlier good entry.

Path-shaped and non-ASCII ids are refused, which is the part that matters most here:

```
ID 'a/b'      -> 400   ID '../etc' -> 400   ID 'Ünïcode' -> 400
```

**The mismatch.** The commit message and the module docstring both say a present id must match the FHIR id shape. `_RESOURCE_ID_PATTERN` is `^[A-Za-z0-9\-\.]{1,255}$`, which is the width of the `id` column, not the FHIR bound of 64:

```
ID 'a' * 64  -> 201 stored
ID 'a' * 65  -> 201 stored
```

The pinned parametrisation uses `'a' * 256` and passes, so it reads as a test of the FHIR bound while actually testing the column bound. Everything from 65 to 255 characters is accepted here and would be refused by a real FHIR server, which matters on a project whose ingested resources are meant to round-trip to one.

Two ways to settle it, and either is fine as long as the words match the code:

1. Tighten the shared pattern to 64 and keep the wording. Then both ingest paths refuse what a server downstream would refuse, and the pinned case at 65 says so.
2. Keep 255, and say "the id column's shape" in both places instead of "the FHIR id shape".

Nothing else came out of the probe. Ids of `.`, `-`, `0`, `null` and a negative integer are all accepted, which is consistent with the stated rule under either reading.

PR #661, opened 2026-09-06
fix(actions): the payload seal closed as a class: a 409 handler for PayloadSealed and an AST ratchet on bulk writers (#620)

### LINKED ISSUE #620: actions: the payload-immutability guard is an ORM validator, bypassable by any future bulk update; the seal error is unhandled

aks129:
Closes #620.

## The two gaps

1. **The seal is an ORM validator.** A bulk `Query.update()` or a Core `update(ProposedAction)` compiles straight to SQL and never fires `@validates('payload_json')`. No live call path carries the column that way today (three bulk writers exist; two update literal dicts without it, `transition_action` refuses it at runtime), and nothing went red if one did.
2. **`PayloadSealed` had no registered handler.** #566 caught it at the one review route that had produced a 500; any other raise site (a future writer, another route, a background job) still answered an unhandled 500.

## What changes

- **AST ratchet** (`tests/actions/test_payload_seal_class.py`): every bulk writer on `ProposedAction` is scanned (`.update(...)` on a `ProposedAction.query` chain, or Core `.values(...)`). A literal mapping may not name `payload_json`; a non-literal mapping is allowed only at a site listed as runtime-guarded, today `transition_action` alone, whose own tests pin the refusal. A second test proves the listed site still exists. Green on arrival, as a ratchet is; alive (three writers found).
- **App-wide handler** (`r6/actions/errors.py`, registered in the app factory beside the kernel's): `PayloadSealed` answers 409 with a fixed, allowlist-constructed message. The exception text names the action id and is logged, never rendered.

## Evidence

- **Mutations**, executed 2026-09-06: name `payload_json` in a literal bulk update in the confirm route, red naming `r6/actions/routes.py:confirm_action (line 620)`; delete the handler registration, the 409 test goes red (the exception escapes).
- `tests/actions`, actions routes, ratchets and conformance green (238 passed). Full suite on the branch: 3664 passed, 13 skipped, 1 xfailed.

Pairs with #658 (the digest is the evidence half; this is the "no silent writer" half). Independent of it: neither touches the other's lines.

### DIFF
--- a/main.py
+++ b/main.py
@@ -291,6 +291,9 @@ def _register_request_hooks(flask_app: Flask) -> None:
     from r6.access import install_audit_assertions, register_error_handlers
 
     register_error_handlers(flask_app)
+    # The payload seal (#528) is a class of refusal, not one route's (#620).
+    from r6.actions.errors import register_error_handlers as register_seal_handler
+    register_seal_handler(flask_app)
     install_audit_assertions(flask_app)
 
     @flask_app.context_processor
--- a/r6/actions/errors.py
+++ b/r6/actions/errors.py
@@ -15,3 +15,37 @@ ALL = (
     PAYLOAD_INVALID, PROVIDER_ERROR, EXTRACTION_AMBIGUOUS,
     EMERGENCY_INDICATED, STALE_SOURCE_DATA,
 )
+
+
+# ---------------------------------------------------------------------------
+# The payload seal, closed as a class (#620)
+# ---------------------------------------------------------------------------
+
+#: Fixed text, never built from the exception: PayloadSealed's message names
+#: the action id, and the reflection rule keeps caller-adjacent text off the
+#: wire (docs/agent-task-guide.md section 2).
+SEALED_MESSAGE = ('This action has already been approved. Its payload is '
+                  'sealed and cannot change; propose a new action instead.')
+
+
+def register_error_handlers(app):
+    """PayloadSealed answers 409 wherever it is raised.
+
+    #566 caught it at the one review route that had produced a 500; any
+    other raise site (a future writer, another route, a background job)
+    still answered an unhandled 500. One handler closes the class.
+    """
+    import logging
+
+    from flask import jsonify
+
+    from r6.actions.models import PayloadSealed
+
+    logger = logging.getLogger(__name__)
+
+    def _render_payload_sealed(exc):
+        logger.info('payload seal refused a write: %s', exc)
+        return jsonify({'error': SEALED_MESSAGE}), 409
+
+    app.register_error_handler(PayloadSealed, _render_payload_sealed)
+
--- a/tests/actions/test_payload_seal_class.py
+++ b/tests/actions/test_payload_seal_class.py
@@ -1,0 +1,126 @@
+"""The payload seal is closed as a class, not per call site (#620).
+
+Two gaps the adversarial review of #550 left open after the #528 repro
+itself was closed:
+
+1. The seal is an ORM validator. A bulk `Query.update()` or a Core
+   `update(ProposedAction)` compiles straight to SQL and never fires it.
+   No live call path carries payload_json that way today, and nothing went
+   red if one did. The AST ratchet below scans every bulk writer on
+   ProposedAction: a literal mapping may not name payload_json, and a
+   non-literal mapping is allowed only at a site that refuses payload_json
+   at runtime and is listed here for it (transition_action, #528).
+2. `PayloadSealed` had no registered handler, so any raise site other than
+   the one review route that catches it answered an unhandled 500. It is
+   now an app-wide 409 with a fixed, allowlist-constructed message that
+   never reflects the exception text.
+
+MUTATIONS: add `'payload_json': ...` to the literal mapping of any bulk
+update in r6/actions/routes.py -> the ratchet goes red naming file, line
+and function. Delete the register_error_handler line in
+r6/actions/errors.py -> the handler test goes red (the exception escapes).
+"""
+
+import ast
+import pathlib
+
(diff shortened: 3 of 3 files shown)


aks129:
## One local variable walks past the ratchet

The 409 handler half is right, and I would take this. The AST half claims more than it checks. Measured on the branch, not read off the diff.

The scan collects a call only when the receiver expression textually contains the name `ProposedAction`:

```python
if 'ProposedAction' not in _names(f.value):
    continue
```

So the write it is looking for is invisible the moment the query is bound to a name first. Same SQL, same bypassed validator:

```python
# red, named exactly as intended
ProposedAction.query.filter_by(id=action_id).update(
    {'payload_json': payload_json}, synchronize_session=False)
#   -> r6/actions/state.py:_probe_direct (line 64): bulk update names payload_json

# green, three tests pass
query = ProposedAction.query.filter_by(id=action_id)
query.update({'payload_json': payload_json}, synchronize_session=False)
```

Both were run against your branch in a scratch worktree, appended to `r6/actions/state.py` and reverted.

**Why keying on the receiver is the hard way.** It asks the scan to know which expressions are queries over this model, which means resolving aliases, and the next form after `query = …` is a helper that returns one.

**The column name is the thing you actually care about, and it is distinctive.** A bulk mapping that names `payload_json` is the defect regardless of what it is called on: no query in this codebase writes a column of that name on any other model. Keying the scan on the mapping key rather than on the receiver removes the alias question, and the runtime-guarded allowlist keeps working as it does now, since the non-literal case is still decided per site.

Two limits worth writing into the docstring rather than fixing, so the next reader knows the shape of what is covered: the scan walks `r6/` only, and raw SQL through `text()` is outside what any AST scan of call shapes can see. Neither has a live call path today.

Filed as #680 so it does not ride on this pull request.

Issue #572, opened 2026-09-03
$extract forks the patient record and writes an allergen name into Patient.code.text


aks129:
Found by the CTO review of PR #562, driven against a running app rather than read from the source. Pre-existing, and it blocks the real-patient path.

## What happens

A committed intake form does not update the patient. It creates a **new** Patient with a fresh uuid, and it writes the answer to a non-demographic question into an element Patient does not have.

```
POST /r6/fhir/QuestionnaireResponse/$extract   (step-up token, not dryRun) -> 200

stored row 52674a15-…:
{"resourceType":"Patient",
 "name":[{"family":"Synthetic"}],
 "code":{"text":"Allergy on file - name not available"}}
```

`Patient.code` is not an element of Patient. The string in it is the answer to `allergies.item.allergen`. With a labelled allergen the same path writes a real allergen name there instead.

## Cause

`_extract_by_definition` walks **every** item carrying a `definition` into a single target resource type. `allergies.item.allergen` is declared as `AllergyIntolerance#AllergyIntolerance.code.text`, but the extraction lands it on the Patient being built. Nothing reconciles the result with the existing Patient, so each committed form forks a new one.

This is documented as a version-one limit at `r6/sdc/intake.py:46-56`. Documenting it does not stop it reaching a record.

## Why it matters now

- **Record forking.** A tenant accumulates a Patient per submitted form. The demo tenant already shows what duplicate patients do to every downstream check: care gaps cannot resolve a subject, and the whole record becomes ambiguous.
- **Fabricated clinical text.** An allergen name written into a non-existent Patient element is data no clinician can find and no validator will catch, sitting in a record that looks fine.
- It rules out an otherwise reasonable fix for the blank-allergen problem, since any marker string placed in that answer would be committed into the record verbatim.

## What has to be true before a real patient uses this

1. Extraction targets the resource each item's `definition` names, rather than folding everything into one type.
2. A committed form updates the existing Patient rather than creating another one, or says plainly that it cannot and refuses.
3. A test that submits the intake form twice and asserts the tenant still holds exactly one Patient.

## Related

PR #562, the intake questionnaire at `r6/sdc/intake.py`, the duplicate-patient shape already seen in the demo tenant, and #570.

aks129:
## Part 2 brief (CTO design pass, 2026-09-06), after #664 lands

[...]

PR #664, opened 2026-09-06
fix(sdc): $extract never writes another resource type's answer into the target (#572, part 1)

aks129:
Part 1 of #572 (the fabricated element). Part 2 (the forked Patient) stays open on the issue; see below.

## What was happening

`_extract_by_definition` walks every item carrying a `definition` into the one target resource type the root `definitionExtract` names. The intake form's allergen item is defined as `AllergyIntolerance#AllergyIntolerance.code.text`, so a committed form wrote its answer into the Patient being built as `Patient.code.text`: an element Patient does not have, holding an allergen name (or the blank-allergen placeholder) that no clinician can find and no validator catches. Measured against a running app by the review that filed #572.

## What changes

A definition names its own resource type, twice: in the StructureDefinition URL before `#` and as the first segment of the element path after it. Both must be the type this questionnaire extracts where the URL says what type it is (the HL7 base namespace, or a bare type name); an answer declared for a type this engine is not building is **dropped rather than misfiled**, and the drop is logged naming the item and the types, never the answer. Demographics extract exactly as before.

The adversarial QA pass on this PR found the first version read only the path: `http://hl7.org/fhir/StructureDefinition/AllergyIntolerance#Patient.name.given` walked through. Fixed and pinned. **Stated limit:** a profile URL (`us-core-patient`, `http://example.org/SD`) does not carry its base type in its name and cannot be checked without resolving the profile, so there the element path's type governs, as it always did; the existing guardrail test that sends `http://example.org/SD#Foobar.name.family` to `$validate` is what draws that line.

## Evidence

- **Pins first** (`tests/test_sdc_extract.py`, red on main): an intake-shaped questionnaire with a Patient target and an allergen item; the extracted Patient carries no `code`, the answer appears nowhere in the bundle, the family name still extracts, and the warning names `allergies.item.allergen`, `AllergyIntolerance` and `Patient` and not the answer.
- **Mutation**, executed 2026-09-06: drop the type comparison, red (the Patient carries `code`).
- SDC extract, guardrails, roundtrip, routes, populate, intake and conformance suites green (86 passed). Full suite on the branch: 3666 passed, 13 skipped, 1 xfailed (with the both-halves rule; 3663 before the QA fix).

## Not in this PR (part 2 of #572)

Each committed form still creates a **new** Patient rather than updating the one the response's `subject` names. Reconciling is a design question (what to merge, keyed on what; a PUT with the form's demographics alone would replace identifiers the form never asked about), and belongs with the engine growing the other target types, so the allergen has somewhere to go instead of being dropped.

### DIFF
--- a/r6/sdc/extract.py
+++ b/r6/sdc/extract.py
@@ -92,15 +92,26 @@ def _extract_by_definition(questionnaire, answers, subject_ref):
         # element Patient does not have and no validator catches. Until the
         # engine builds those types too, the answer is dropped and the drop
         # is said out loud, naming the item and the types, never the answer.
-        defined_type = path.split(".", 1)[0]
-        if defined_type != target_type:
+        # A definition names its resource type twice: in the StructureDefinition
+        # URL before "#" and as the first segment of the element path after
+        # it. Both must be the type this questionnaire extracts. The QA pass
+        # on #664 showed why one is not enough: `AllergyIntolerance#Patient.
+        # name.given` passed a path-only check and landed the allergen answer
+        # in Patient.name.given, and a Questionnaire is a stored resource
+        # that can arrive through the ordinary ingest paths, not only the
+        # intake form this engine was written for.
+        defined_url_type = _definition_url_type(definition)
+        defined_path_type = path.split(".", 1)[0]
+        if defined_path_type != target_type or (
+                defined_url_type and defined_url_type != target_type):
             # %r for the linkId, as the handler in expressions.py does: it is
             # caller-supplied questionnaire structure, and %r is what escapes
             # a newline that would forge a log line.
             logger.warning(
                 "extract: item %r not extracted: its definition targets %s "
-                "and this questionnaire extracts %s (#572)",
-                item.get("linkId"), defined_type, target_type)
+                "(url) / %s (path) and this questionnaire extracts %s (#572)",
+                item.get("linkId"), defined_url_type or "none",
+                defined_path_type, target_type)
             continue
         _value_key, value = _answer_value(item_answers[0])
         if value is None:
@@ -110,6 +121,27 @@ def _extract_by_definition(questionnaire, answers, subject_ref):
     if not populated:
         return []
     return [_post_entry(resource)]
+
+
+_HL7_BASE_SD = "http://hl7.org/fhir/StructureDefinition/"
+
+
+def _definition_url_type(definition):
+    """The base resource type the StructureDefinition URL before "#" names,
+    when that can be read off the URL: the last segment under the HL7 base
+    namespace (http://hl7.org/fhir/StructureDefinition/Patient -> Patient),
+    or a bare type name ("Patient"). Any other URL is a profile whose base
+    type is not in its name (http://example.org/SD, us-core-patient), and
+    cannot be checked without resolving the profile: "" means unknown, and
+    the element path's own type is what the check has."""
+    url = definition.split("#", 1)[0].strip()
+    if not url:
+        return ""
+    if url.startswith(_HL7_BASE_SD):
+        return url[len(_HL7_BASE_SD):].strip("/")
+    if "/" not in url and ":" not in url:
+        return url
+    return ""
 
 
 def _set_path(resource, dotted_path, value):
--- a/r6/sdc/extract.py
+++ b/r6/sdc/extract.py
@@ -8,6 +8,10 @@ Pure function. Supports two SDC extraction mechanisms:
 Out of scope (v1): template-based and StructureMap-based extraction.
 """
 
+import logging
+
+logger = logging.getLogger(__name__)
+
 OBSERVATION_EXTRACT_URL = (
     "http://hl7.org/fhir/uv/sdc/StructureDefinition/"
     "sdc-questionnaire-observationExtract"
@@ -79,6 +83,21 @@ def _extract_by_definition(questionnaire, answers, subject_ref):
         path = definition.split("#", 1)[1]  # e.g. Patient.name.family
         item_answers = answers.get(item.get("linkId"), [])
         if not item_answers:
+            continue
+        # #572: a definition names its own resource type. This engine builds
+        # ONE target type, so an answer declared for another type (the intake
+        # form's allergen, AllergyIntolerance#AllergyIntolerance.code.text)
+        # must not be written into the target as if it were an element of
+        # it: that produced Patient.code.text holding an allergen name, an
+        # element Patient does not have and no validator catches. Until the
+        # engine builds those types too, the answer is dropped and the drop
+        # is said out loud, naming the item and the types, never the answer.
+        defined_type = path.split(".", 1)[0]
+        if defined_type != target_type:
+            logger.warning(
+                "extract: item %s not extracted: its definition targets %s "
+                "and this questionnaire extracts %s (#572)",
+                item.get("linkId"), defined_type, target_type)
             continue
         _value_key, value = _answer_value(item_answers[0])
         if value is None:
--- a/r6/sdc/extract.py
+++ b/r6/sdc/extract.py
@@ -94,8 +94,11 @@ def _extract_by_definition(questionnaire, answers, subject_ref):
         # is said out loud, naming the item and the types, never the answer.
         defined_type = path.split(".", 1)[0]
         if defined_type != target_type:
+            # %r for the linkId, as the handler in expressions.py does: it is
+            # caller-supplied questionnaire structure, and %r is what escapes
+            # a newline that would forge a log line.
             logger.warning(
-                "extract: item %s not extracted: its definition targets %s "
+                "extract: item %r not extracted: its definition targets %s "
                 "and this questionnaire extracts %s (#572)",
                 item.get("linkId"), defined_type, target_type)
             continue
--- a/tests/test_sdc_extract.py
+++ b/tests/test_sdc_extract.py
@@ -69,3 +69,74 @@ def test_extract_empty_when_no_directives():
 
     bundle = extract_resources(qr, q)
     assert bundle["entry"] == []
+
+
+# ---------------------------------------------------------------------------
+# #572, part 1: an answer whose definition names ANOTHER resource type is never
+# written into the extraction target.
+# ---------------------------------------------------------------------------
+
+def _intake_like_questionnaire():
+    return {
+        "resourceType": "Questionnaire",
+        "extension": [{"url": DEF_EXTRACT_URL, "valueCode": "Patient"}],
+        "item": [
+            {"linkId": "family", "type": "string",
+             "definition": "http://hl7.org/fhir/StructureDefinition/Patient#Patient.name.family"},
+            {"linkId": "allergies", "type": "group", "item": [
+                {"linkId": "allergies.item.allergen", "type": "string",
+                 "definition": "http://hl7.org/fhir/StructureDefinition/"
+                               "AllergyIntolerance#AllergyIntolerance.code.text"},
+            ]},
+        ],
+    }
+
+
+def _intake_like_response():
+    return {
+        "resourceType": "QuestionnaireResponse", "status": "completed",
+        "subject": {"reference": "Patient/p-572"},
+        "item": [
+            {"linkId": "family", "answer": [{"valueString": "Synthetic"}]},
+            {"linkId": "allergies", "item": [
+                {"linkId": "allergies.item.allergen",
+                 "answer": [{"valueString": "peanut-572"}]},
+            ]},
+        ],
+    }
+
+
+def test_an_answer_for_another_resource_type_never_lands_on_the_target(caplog):
+    """The engine walked every `definition` into the one target type, so the
+    allergen answer (AllergyIntolerance#AllergyIntolerance.code.text) was
+    written as Patient.code.text: an element Patient does not have, holding
+    clinical text no clinician can find and no validator catches (#572).
+
+    MUTATION: r6/sdc/extract.py, drop the resource-type comparison in
+    _extract_by_definition -> red (the Patient carries `code`).
+    """
+    import json
+    import logging
+    with caplog.at_level(logging.WARNING, logger="r6.sdc.extract"):
+        bundle = extract_resources(_intake_like_response(), _intake_like_questionnaire())
+    patients = [e["resource"] for e in bundle["entry"]
+                if e["resource"]["resourceType"] == "Patient"]
+    assert len(patients) == 1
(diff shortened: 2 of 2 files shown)


aks129:
Adversarial QA pass on the type-mismatch drop, run against real extraction calls (not read from source). One real bypass found; two other adversarial angles hold.

## Real bypass: the check reads only the path side of `definition`, never the URL side

A `definition` string is `<StructureDefinition-url>#<element-path>`. The fix's comparison is `defined_type = path.split(".", 1)[0]` — the leading segment of the *path*, taken from whatever follows `#`. It never looks at the StructureDefinition reference before `#`.

Constructed a definition where the two disagree — `AllergyIntolerance#Patient.name.given` (URL names AllergyIntolerance; the path claims `Patient.name.given`) — alongside a normal `Patient#Patient.name.family` item, extracting to a `Patient` target:

```
{'resourceType': 'Patient', 'name': [{'family': 'Synthetic', 'given': ['peanut-disagreement']}]}
```

The allergen answer lands in `Patient.name.given`. The check trusts that the path's own claimed type matches the URL it's attached to; nothing enforces that they agree, so a `definition` that lies about its own prefix walks straight past the comparison this PR adds — the same class of defect #572 named, just moved one string-segment over.

**This isn't purely theoretical**: `Questionnaire` is in `R6Resource.SUPPORTED_TYPES` (`r6/models.py`), reachable via the standard ingest paths, and `review.py`'s `_resolve_questionnaire` will load a tenant-stored `Questionnaire` by id for extraction. A malformed or crafted upstream Questionnaire — not just the canonical intake form this PR's own test exercises — could carry a mismatched `definition` and reach this code path.

**Fix shape**, not prescribing: derive `defined_type` from the URL segment before `#` instead of (or in addition to — cross-checking both and refusing on disagreement is probably the stronger version) the path segment after it. If the intent is "the path's claim is authoritative," that should probably be stated explicitly, since it's a much weaker property than "the definition names its own resource type" as the PR's docstring currently frames it.

## Two things that hold

- **Multiple mismatches in one extraction**: two mismatched items followed by a legitimate one — both mismatches drop independently (no short-circuit), the legitimate item still extracts, and the log names each item separately.
- **A definition with no `.` after `#`** (a bare type reference like `Patient#Patient`): does not crash; extracts an empty-ish resource cleanly.

## Test file (not pushed)

`tests/test_qa_adversarial_extract_and_review_content.py` — the bypass test plus the two clean ones, against a worktree with this PR's head merged onto current `main`, no conflicts. Full content in a follow-up comment.

No push, no arm, per the standing rule on this round.

aks129:
Full test file (covers both #664 and #665, same worktree):

```python
"""Adversarial QA on #664 ($extract type-mismatch drop, commit 53f5790) and
#665 (review content list holds no Patient). Standalone probe, not part of
either PR. Findings reported as PR comments, nothing pushed.
"""

import json
import logging

from r6.sdc.extract import extract_resources

DEF_EXTRACT_URL = (
    "http://hl7.org/fhir/uv/sdc/StructureDefinition/"
    "sdc-questionnaire-definitionExtract"
)


# ---------------------------------------------------------------------------
# #664: the fix compares `path.split("#", 1)[1].split(".", 1)[0]` (the
# element-path's own leading segment) against the root's target type. It
# never looks at the StructureDefinition reference BEFORE the "#". If those
# two disagree -- the URL says one type, the path claims another -- the
# check only ever sees the path's claim.
# ---------------------------------------------------------------------------

def test_a_definition_whose_url_and_path_prefix_disagree():
    """A definition string is `<StructureDefinition-url>#<element-path>`.
    Convention has the path's leading segment match the resource type in
    the URL, but the fix's comparison reads ONLY the path side
    (`defined_type = path.split(".", 1)[0]`) and never checks it against
    the URL. Constructing a definition where they disagree -- the URL
    names AllergyIntolerance, the path CLAIMS Patient -- to see whether
    the allergen answer lands in Patient.name.given anyway.
    """
    q = {
        "resourceType": "Questionnaire",
        "extension": [{"url": DEF_EXTRACT_URL, "valueCode": "Patient"}],
        "item": [
            {"linkId": "family", "type": "string",
             "definition": "http://hl7.org/fhir/StructureDefinition/"
                           "Patient#Patient.name.family"},
            {"linkId": "allergen", "type": "string",
             # URL says AllergyIntolerance. Path claims Patient.name.given.
             "definition": "http://hl7.org/fhir/StructureDefinition/"
                           "AllergyIntolerance#Patient.name.given"},
        ],
    }
    qr = {"resourceType": "QuestionnaireResponse", "status": "completed",
          "item": [
              {"linkId": "family", "answer": [{"valueString": "Synthetic"}]},
              {"linkId": "allergen",
               "answer": [{"valueString": "peanut-disagreement"}]},
          ]}

    bundle = extract_resources(qr, q)
    patient = [e["resource"] for e in bundle["entry"]
              if e["resource"]["resourceType"] == "Patient"][0]

    print('RESULT:', json.dumps(patient, indent=2))
    landed = "peanut-disagreement" in json.dumps(patient)
    assert not landed, (
        "the URL-vs-path-prefix disagreement bypassed the type check: "
        f"the allergen answer landed in Patient anyway: {patient}")


# ---------------------------------------------------------------------------
# #664: multiple mismatched items in the same extraction -- does the drop
# continue past the first one, or does it stop processing (silently
# dropping a LEGITIMATE later item too)?
# ---------------------------------------------------------------------------

def test_multiple_mismatches_each_drop_independently_later_legit_items_still_extract():
    q = {
        "resourceType": "Questionnaire",
        "extension": [{"url": DEF_EXTRACT_URL, "valueCode": "Patient"}],
        "item": [
            {"linkId": "allergen1", "type": "string",
             "definition": "http://hl7.org/fhir/StructureDefinition/"
                           "AllergyIntolerance#AllergyIntolerance.code.text"},
            {"linkId": "condition1", "type": "string",
             "definition": "http://hl7.org/fhir/StructureDefinition/"
                           "Condition#Condition.code.text"},
            # A legit Patient field AFTER two mismatches.
            {"linkId": "family", "type": "string",
             "definition": "http://hl7.org/fhir/StructureDefinition/"
                           "Patient#Patient.name.family"},
        ],
    }
    qr = {"resourceType": "QuestionnaireResponse", "status": "completed",
          "item": [
              {"linkId": "allergen1", "answer": [{"valueString": "peanut"}]},
              {"linkId": "condition1", "answer": [{"valueString": "asthma"}]},
              {"linkId": "family", "answer": [{"valueString": "Lovelace"}]},
          ]}
    bundle = extract_resources(qr, q)
    patient = [e["resource"] for e in bundle["entry"]
              if e["resource"]["resourceType"] == "Patient"][0]
    assert patient["name"][0]["family"] == "Lovelace"
    blob = json.dumps(patient)
    assert "peanut" not in blob and "asthma" not in blob


# ---------------------------------------------------------------------------
# #664: a definition with no "." after the "#" (defined_type derivation
# should not crash on a bare type with no element path).
# ---------------------------------------------------------------------------

def test_definition_with_no_dot_after_hash_does_not_crash():
    q = {
        "resourceType": "Questionnaire",
        "extension": [{"url": DEF_EXTRACT_URL, "valueCode": "Patient"}],
        "item": [{"linkId": "weird", "type": "string",
                  "definition": "http://hl7.org/fhir/StructureDefinition/"
                                "Patient#Patient"}],
    }
    qr = {"resourceType": "QuestionnaireResponse", "status": "completed",
          "item": [{"linkId": "weird", "answer": [{"valueString": "x"}]}]}
    # Must not raise.
    bundle = extract_resources(qr, q)
    print('NO-DOT RESULT:', bundle)


# ---------------------------------------------------------------------------
# #665: confirm the content list never carries the Patient across the
# three subject-resolution states (resolved / unresolved / ambiguous), not
# just the one state the PR's own pin exercises.
# ---------------------------------------------------------------------------

def test_content_list_holds_no_patient_across_resolution_states(app):
    """End to end through _gather_content, not a static read of
    _CONTENT_TYPES: seed a real Patient row plus an AllergyIntolerance
    that references it, and confirm the returned content list carries
    the allergy but never the Patient resource itself -- across a
    resolved patient, an unresolved one, and the unanchored-urn state."""
    from models import db
    from r6.actions.review import _gather_content, CONTENT_REASON_UNANCHORED
    from r6.models import R6Resource

    tenant = 'qa-665-tenant'
    with app.app_context():
        db.session.add(R6Resource(
            resource_type='Patient', resource_id='p-665',
            resource_json=json.dumps(
                {'resourceType': 'Patient', 'id': 'p-665',
                 'name': [{'family': 'Rivera'}]}),
            tenant_id=tenant))
        db.session.add(R6Resource(
            resource_type='AllergyIntolerance', resource_id='a-665',
            resource_json=json.dumps(
                {'resourceType': 'AllergyIntolerance', 'id': 'a-665',
                 'patient': {'reference': 'Patient/p-665'},
                 'code': {'text': 'Penicillin'}}),
            tenant_id=tenant))
        # Unanchored: an AllergyIntolerance whose urn resolves to nobody.
        db.session.add(R6Resource(
            resource_type='AllergyIntolerance', resource_id='a-orphan',
            resource_json=json.dumps(
                {'resourceType': 'AllergyIntolerance', 'id': 'a-orphan',
                 'patient': {'reference':
                            'urn:uuid:not-a-stored-patient-id'},
                 'code': {'text': 'Sulfa'}}),
            tenant_id=tenant))
        db.session.commit()

        # State 1: resolved patient.
        patient = {'resourceType': 'Patient', 'id': 'p-665'}
        result = _gather_content(tenant, patient)
        types_seen = {r['resourceType'] for r in result.resources}
        assert 'Patient' not in types_seen
        assert 'AllergyIntolerance' in types_seen
        # The unanchored row correctly does NOT appear in the content list
        # itself (it cannot be safely attributed to this patient, so it
        # must not auto-populate their form) -- only the unanchored FLAG
        # is expected to surface, via `reason`.
        assert not any(r.get('id') == 'a-orphan' for r in result.resources)
        assert result.reason == CONTENT_REASON_UNANCHORED

        # State 2: unresolved (no patient at all).
        result_unresolved = _gather_content(tenant, None)
        assert result_unresolved.resources == []
        assert 'Patient' not in [r.get('resourceType')
                                for r in result_unresolved.resources]

    print('content list never carries Patient in either state; '
         'the orphaned/unanchored row is correctly still present')
```

aks129:
Confirmed and fixed at f29fc53. The check read only the element path's first segment; `http://hl7.org/fhir/StructureDefinition/AllergyIntolerance#Patient.name.given` now refuses, because where the URL says what type it is (the HL7 base namespace, or a bare type name) it must match the target too. The disagreement is dropped and logged naming both halves; the answer appears nowhere in the bundle (pinned; mutation: drop the URL half, red).

One limit, stated in the code and pinned rather than hidden: a **profile** URL (`us-core-patient`, `http://example.org/SD`) does not carry its base type in its name and cannot be checked without resolving the profile, so there the element path's type governs, as it always did. The existing guardrail test (`http://example.org/SD#Foobar.name.family` must still reach `$validate` and answer 422) is what drew that line. A crafted profile URL with a matching path (`http://example.org/x#Patient.name.given` on an allergen item) is therefore still the form author's call, which is the design of definition-based extraction: the form decides where an answer lands. Closing that fully means the engine building the other target types so an allergen answer has somewhere to go, which is part 2 of #572.

Your other two angles (independent drops without short-circuiting, and a definition with no element path) are pinned too.

aks129:
Re-verified against f29fc53 — all three named variants behave as described, plus two extras.

| Variant | Definition | Result |
|---|---|---|
| 1. HL7 base URL vs. disagreeing path | `http://hl7.org/fhir/StructureDefinition/AllergyIntolerance#Patient.name.given` | **Refused** — the exact round-1 bypass, now caught |
| 2. Profile URL, matching path | `http://example.org/fhir/StructureDefinition/my-patient-profile#Patient.name.given` | **Extracts** — by design, as documented |
| 3. Bare type name vs. disagreeing path | `AllergyIntolerance#Patient.name.given` | **Refused** |

Two extras, to check the edges of the documented limitation:
- **Profile URL where the path *also* disagrees** (`http://example.org/.../my-allergy-profile#AllergyIntolerance.code.text` extracting into a Patient target) — still **refused**. The path-only floor catches it even when the URL side is unreadable, confirming variant 2's leniency is specific to "URL unreadable + path agrees," not "URL unreadable" alone.
- **Bare type name that agrees** (`Patient#Patient.name.given`) — still **extracts**, confirming the new check isn't overzealous on the legitimate case.

`tests/test_sdc_extract.py` — 8 passed, no regressions.

Agreed this closes the reproduction I filed. The remaining profile-URL gap is a fair, explicitly-documented limitation rather than an oversight — a profile URL genuinely can't be resolved to a base type without fetching the StructureDefinition, and the path governing there is a sensible floor given that. Nothing further from me on #664.

No push, no arm.

aks129:
Re-verification against f29fc53 received: all three named variants behave as stated, and the two extras (a profile URL whose path also disagrees still refuses on the path floor; a bare type name that agrees still extracts) confirm the leniency is specific to an unreadable URL with an agreeing path. The profile-URL gap stays a stated limitation until part 2 of #572. Nothing further needed on this PR before arming.

aks129:
## The check bounds the type that is named, not the element that is written

The URL-and-path fix in patch 4 is right, and the profile case is handled honestly. What is left is that both halves of the definition are written by whoever authors the questionnaire, so declaring them consistently is free. Measured on the branch:

```
PATH Patient.code.text            -> {"resourceType": "Patient", "code": {"text": "peanut-probe"}}
PATH Patient.notAnElement.at.all  -> {"resourceType": "Patient", "notAnElement": {"at": {"all": "peanut-probe"}}}
PATH Patient.deceasedBoolean      -> {"resourceType": "Patient", "deceasedBoolean": "peanut-probe"}
```

The first line is the resource this pull request's own docstring describes as the harm: an allergen in `Patient.code.text`, an element Patient does not have, holding clinical text no clinician can find and no validator catches. Reaching it now needs only `Patient#Patient.code.text` instead of `AllergyIntolerance#AllergyIntolerance.code.text`.

The third line is a second shape: a string written where FHIR requires a boolean, which is the same class as #666.

**What I would change.** The type comparison stays; add the element. The path's head after the type has to be an element the target type actually has, and an unknown one is dropped through the log line this pull request already writes. That is a per-type set of element names, which is a small table for the three or four types this engine builds, and it turns "declares consistently" from a bypass into a refusal.

If that is too much for a part-1 change, the alternative is to say in the docstring that the check constrains the declared type only, so the next reader does not take the stronger guarantee from it. What is not safe is the wording as it stands, because it reads as though an answer can no longer land on an element the target does not have.

Filed as #681. Probe is in a scratch worktree, not proposed for merge.

Issue #390, opened 2026-08-05
The intake form claims 'No allergies found in your records' — an absence claim, aimed at the control the demo showcases


aks129:
Found by Product's honesty sweep of the demo path. This one is pointed directly at the beat the webinar exists to show.

## What happens

If the tenant has no `Patient` row, or references its patient by `urn:uuid:`, `_gather_content` returns empty and the intake form renders:

> No allergies found in your records

immediately above the no-known-allergies attestation checkbox.

## Why this is worse than an ordinary absence claim

"No known allergies" is a **constitutional never-infer** in this codebase — it may come only from an explicit human attestation, and the checkbox exists precisely to obtain one. The gate itself is sound. But the line above it tells the patient the record already agrees, so the attestation it collects is an agreement with a sentence a lookup emptied.

A patient who skims, sees "no allergies found", and ticks the box has been walked into attesting to something nobody checked. The control is intact and its context is false, which is the harder failure to spot in a review and the one this week's pattern keeps producing.

It is also **beat 7 of the demo** — the differentiator, the part that shows a guardrail refusing to submit without human confirmation.

## Two triggers, both ordinary

- No `Patient` row in the tenant. CareAgents mints a tenant per connection; a source that sends clinical resources before demographics leaves this state for real.
- The patient referenced as `urn:uuid:` — standard inside a transaction Bundle, which is how bundles arrive.

## The fix

`_gather_content` must distinguish "looked, found none" from "could not resolve the patient". Only the first may render an absence line; the second should say the record could not be read and must not present the attestation as a confirmation of anything.

Same three-state shape as #381's care-gaps fix and #379's medication wording — this is the third instance, which is an argument for solving it once at the source rather than a fourth time at a call site.

## Also worth knowing

Product's biggest stated risk for Aug 18 is that **this beat has never been run** on the path a phone actually uses. The forms rail is proven on a synthetic tenant from an MCP desktop client with a hand-minted token; the CareAgents relay is covered only by a fake whose review page is a one-line string — the documented trap where a test proves the call is made, not accepted.

Related: #381, #379, #376.

PR #665, opened 2026-09-06
fix(review): the content list handed to the populate engine no longer holds the Patient, and the expression handler is reached (#581)

### LINKED ISSUE #581: the action rail holds an unredacted patient beside a redaction boundary, and one test does not cover the branch it names
Two small things found while closing the CTO's required items on PR #562. Neither is urgent; both are the same shapes that produced a real leak on the neighbouring path, so they are worth closing while the reasoning is fresh.

## 1. The action rail appends the patient the same way `$populate` used to

`r6/actions/review.py:208` does `content = [patient]` on the form-fill path.

On the `$populate` path that same append was the second door into an identifier oracle, and PR #578 deleted it after proving nothing read it. The same four-read proof appears to apply here: the content list is consumed by filters that a Patient does not survive. But this is a different route with a different caller, so it needs its own check rather than an assumption.

If nothing reads it, delete it. Dead weight that holds a whole unredacted patient beside a redaction boundary is the kind of thing that becomes a leak the next time someone adds a context variable.

If something does read it, that is more interesting, and the answer is a projection or a redacted copy rather than the whole resource.

## 2. A test reads as covering an error branch it never reaches

`test_evaluate_returns_none_on_bad_expression` passes `"this is not fhirpath ((("`, which the expression library returns an empty result for rather than raising. So the test exercises the empty-result path, not the exception handler it appears to cover.

It is not wrong, it just does not test what its name says. One test in PR #578 is currently the only thing that reaches the handler.

Either rename it to say what it does, or find an input that actually raises and keep both. The second is better, because the handler swallows exceptions so that one bad expression cannot abort a whole form, and nothing else pins that behaviour.

## Related

PR #562, PR #576, PR #578, and the QA review that found the original oracle.

aks129:
Closes #581.

## Item 1: an unredacted Patient beside a redaction boundary

`_gather_content` in `r6/actions/review.py` seeded its list with the Patient. `populate_questionnaire` takes the subject as its own argument, and its docstring already says the list must not hold it (the #562 review); the copy was dead weight sitting beside a redaction boundary, the same door `$populate` closed in #578 after proving nothing read it. Checked here with its own proof rather than the assumption the issue asked us not to make: the list reaches the engine and nowhere else (the page reads only `content.resolved` and `content.reason`), and the engine matches the list groups (medications, allergies, conditions) against it, never the subject. The list now starts empty; the clinical rows reach the engine unchanged.

**Pin** (red on main): a resolved patient with one allergy yields a content list of exactly `['AllergyIntolerance']`, and on the page's own path the engine receives the subject separately and a list with no Patient. **Mutation**, executed 2026-09-06: seed the list with the patient again, red.

## Item 2: a test that did not reach the branch it named

`test_evaluate_returns_none_on_bad_expression` passed `"this is not fhirpath ((("`, which the library answers `[]` for rather than raising, so it exercised the empty-result path. It is renamed to say so. A new test reaches the handler with an unknown function (`foo()`, which the library raises `Not implemented: foo` on), and pins that `evaluate` answers `None` and logs the failure naming the item and never the expression (the #581 reflection point, already documented beside the handler).

Actions, SDC expressions/populate/roundtrip, intake, careagents and conformance suites green (487 passed). Full suite on the branch: 3663 passed, 13 skipped, 1 xfailed, 2 warnings.

### DIFF
--- a/r6/actions/review.py
+++ b/r6/actions/review.py
@@ -205,7 +205,12 @@ def _gather_content(tenant_id, patient, subject_ref=None):
 
     patient_id = patient['id']
     ref = 'Patient/%s' % patient_id
-    content = [patient]
+    # The Patient is NOT in this list (#581). populate_questionnaire takes
+    # the subject as its own argument; a copy in here was dead weight held
+    # beside a redaction boundary, the same door $populate closed in #578
+    # after proving nothing read it. Nothing reads it here either: the
+    # list reaches the engine and nowhere else.
+    content = []
     unanchored = False
     for resource_type, subject_field in _CONTENT_TYPES:
         for row in R6Resource.query.filter_by(
--- a/tests/actions/test_review_flow.py
+++ b/tests/actions/test_review_flow.py
@@ -668,3 +668,56 @@ def test_review_refuses_a_read_scoped_token(client, app, tenant_headers,
             tenant_headers['X-Tenant-Id'], scope='read'),
     }, action_id)
     assert resp.status_code == 401, resp.get_data(as_text=True)
+
+
+# ---------------------------------------------------------------------------
+# #581 item 1: the content list handed to the populate engine never holds the
+# Patient. The Patient arrives as `subject`; a copy inside the list was dead
+# weight sitting beside a redaction boundary, the same door $populate closed
+# in #578 after proving nothing read it.
+# ---------------------------------------------------------------------------
+
+def test_the_content_list_never_holds_the_patient(app, tenant_id, monkeypatch):
+    """MUTATION: r6/actions/review.py, seed the list with the patient again
+    (`content = [patient]`) -> red."""
+    from r6.actions import review
+    from r6.models import R6Resource
+    from models import db
+
+    with app.app_context():
+        db.session.add(R6Resource(
+            resource_type='Patient', resource_id='p-581', tenant_id=tenant_id,
+            resource_json='{"resourceType":"Patient","id":"p-581","name":[{"family":"Quux581"}]}'))
+        db.session.add(R6Resource(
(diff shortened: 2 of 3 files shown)


aks129:
Adversarial QA pass, run end to end through `_gather_content` (seeded real rows, not a static read of `_CONTENT_TYPES`). No findings — this one holds.

## Confirmed clean

Seeded a real `Patient` row, an `AllergyIntolerance` correctly referencing it, and a second `AllergyIntolerance` with an unanchored urn (referencing nobody stored). Called `_gather_content` across the resolved-patient state and the unresolved (`None`) state:

- **Patient never appears in `resources`** in either state — confirmed on the actual returned list, not by inspecting `_CONTENT_TYPES`'s contents.
- The legitimately-referenced allergy **does** appear.
- The **unanchored** allergy correctly does **not** appear in the population list (verified this is intentional, not an oversight: including it would mean auto-populating this patient's intake form with clinical data that might belong to someone else entirely, since the urn can't be resolved to anyone). Its presence still surfaces via `reason == CONTENT_REASON_UNANCHORED`, so the honesty guarantee (#390's territory) holds without conflating "flag the ambiguity" with "hand it to the populate engine."

Also read the reflection-point fix in `r6/sdc/expressions.py`'s `evaluate()` — already well-guarded against the two adjacent risks I'd have probed for: the expression text itself is never logged (only `link_id`), and the `warned` set caps log volume per linkId per request so a malformed expression inside a repeating list group can't be used to flood logs from caller-controlled input. Didn't find anything to push on there.

## Test file (not pushed)

`tests/test_qa_adversarial_extract_and_review_content.py` (shared with the #664 comment — same worktree, both PRs merged cleanly onto current `main`, no conflicts). Full content in a follow-up comment on #664 if useful here too.

No push, no arm.

PR #668, opened 2026-09-06
fix(sdc): $extract commit mode refuses to write clinical rows on a step-up token alone (#572)

aks129:
Part of #572, lifted out of the held part-2 stack because it closes a live guardrail gap on `main`. **This one jumps the queue**, the way #615 did.

## Measured on main, not inferred

Nothing on the human-gated path calls `$extract`: the form-fill executor never extracts, and `$extract`'s callers are the raw endpoint (step-up only) and the MCP write tool. In a scratch worktree on `main`: a Questionnaire whose root `definitionExtract` names `AllergyIntolerance`, with definitions for `clinicalStatus`, `verificationStatus`, `patient` and `code.text` answered as plain strings, POSTed to commit mode with a step-up token:

```
status=200 rows_before=0 rows_after=1
```

One AllergyIntolerance stored, its statuses and patient reference written as bare strings that the validator waved through by truthiness. A clinical write with no human confirmation, on the credential an agent holds: the class the CTO design pass for #572 named worse than the #214 header.

## What changes

- `RAIL_ONLY_TYPES` (`AllergyIntolerance`, `Condition`, `MedicationRequest`) in `r6/sdc/extract.py`.
- Commit mode on `$extract` refuses a bundle carrying any of them with a 422 OperationOutcome naming the form-fill rail and `dryRun`, before validation and before any write. `dryRun` still previews them. The refusal writes no audit row, consistent with the kernel's own refusals, so the audit trail does not cover refused attempts here.

## Evidence

- **Pin** (`tests/test_extract_refuses_clinical_rows_on_step_up_alone.py`): the measured request now answers 422 and stores nothing; dryRun previews it; the three types are named.
- **Mutation**, executed 2026-09-06: drop the refusal, red (200 and a stored row).
- SDC guardrails/routes/roundtrip/extract, the write-guard matrix, conformance and the ratchets green (247 passed). Full suite on the branch: 3664 passed, 13 skipped, 1 xfailed.

## What this leaves

The rest of part 2 of #572 (no Patient on a subject-bound response, populated rows carrying their source, the engine building the clinical rows for the executor to commit after confirmation) is built and held as branches behind #664; that stack drops this commit when it retargets.

### DIFF
--- a/r6/sdc/extract.py
+++ b/r6/sdc/extract.py
@@ -18,6 +18,14 @@ DEFINITION_EXTRACT_URL = (
 )
 
 
+#: The types the form-fill rail commits after human confirmation. Nothing on
+#: the human-gated path calls $extract, so commit mode on the raw endpoint
+#: refuses a bundle carrying them (r6/sdc/routes.py): a step-up token alone
+#: must not write clinical rows (#572).
+RAIL_ONLY_TYPES = frozenset({"AllergyIntolerance", "Condition",
+                             "MedicationRequest"})
+
+
 def extract_resources(questionnaire_response, questionnaire):
     """Return a FHIR transaction Bundle of resources extracted from `qr`."""
     subject_ref = questionnaire_response.get("subject")
--- a/r6/sdc/routes.py
+++ b/r6/sdc/routes.py
@@ -31,7 +31,7 @@ from r6.models import R6Resource
 from r6.audit import record_audit_event
 from r6.redaction import apply_redaction
 from r6.sdc.populate import NOT_POPULATED, populate_questionnaire
-from r6.sdc.extract import extract_resources
+from r6.sdc.extract import RAIL_ONLY_TYPES, extract_resources
 
 logger = logging.getLogger(__name__)
 
@@ -139,6 +139,23 @@ def register_sdc_routes(blueprint, deps):
                 "Questionnaire for the response could not be resolved"), 404
 
         bundle = extract_resources(qr, questionnaire)
+
+        if not dry_run and any(
+                e["resource"]["resourceType"] in RAIL_ONLY_TYPES
+                for e in bundle["entry"]):
+            # #572. Nothing on the human-gated path calls $extract, so commit
+            # mode here runs on a step-up token alone; a bundle carrying
+            # allergies, conditions or medications is refused rather than
+            # written on that, and rather than dropped silently after dryRun
+            # showed it. The form-fill rail commits those rows after the
+            # person confirms them. Measured on main before this: the legacy
+            # single-target engine wrote an AllergyIntolerance this way.
+            return operation_outcome(
+                "error", "business-rule",
+                "This bundle carries clinical rows (allergies, conditions or "
+                "medications). The form-fill rail commits those after human "
+                "confirmation; $extract does not. Use dryRun=true to preview "
+                "them."), 422
 
         if not dry_run:
             # H4 posture (deliberate): $extract commits clinical resources as a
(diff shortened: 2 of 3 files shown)


aks129:
Adversarial QA pass, driven against live Flask instances on real migrated SQLite DBs (not the test client), plus the real Node MCP server.

**Reproduced the gap on `main` first**, exact scenario from the PR body: root `definitionExtract=AllergyIntolerance`, `clinicalStatus`/`verificationStatus`/`patient`/`code.text` all answered as plain strings, POSTed to `$extract` commit mode with only a step-up token:

```
rows_before=0
status=200
rows_after=1
```

Stored row: `{"resourceType": "AllergyIntolerance", "clinicalStatus": "active", "verificationStatus": "confirmed", "patient": "Patient/p-668", "code": {"text": "peanut-668"}}` — confirms the measured claim, including the malformed bare-string shape.

**Same request against this PR's head**: `dryRun=true` still returns 200 with the built (still bare-string-shaped) preview; commit mode now returns 422 naming the form-fill rail and `dryRun`; `rows_after=0`.

**The two named variants:**

- **MCP write tool in commit mode.** `questionnaire_extract` (tier: write) with no `X-Step-Up-Token` refuses with `requires_step_up` before ever reaching Flask. With a valid token it forwards to the same route and gets the same 422 (`"$extract failed with status 422"`); DB count stays 0.
- **Mixed bundle (Observation + AllergyIntolerance).** Added an `observationExtract`-flagged LOINC item to the same questionnaire/response. `dryRun` previews both entries. Commit mode refuses the *entire* request with the same 422 — the otherwise-allowed Observation is not written either (the `any(...)` check runs before the commit loop, so there's no partial commit). DB counts for both `AllergyIntolerance` and `Observation` are 0 after.

Full suite on the branch: 3664 passed, 13 skipped, 1 xfailed — matches the number quoted in the PR body, no regressions found.

No bypass found on any of the four angles tried. Findings only; no push, no arm.

aks129:
## The refusal is a denylist over a type the caller chooses

This closes the case it measured, and I would take it. But the class it names is still open, and I think the shape of the guard is the reason. Both of the following are measured against 8e5fbef, not read off the diff.

**An Observation still commits on a step-up token alone.**

```
OBSERVATION status 200 stored 1
```

A questionnaire carrying the observation-extract extension with one coded item, answered 191, commits one Observation. A systolic blood pressure of 191 is a clinical row by any reading of the rule this pull request is enforcing, and it is the most commonly extracted type of the lot.

**And the target type is whatever the questionnaire says it is.**

```
CONSENT status 200 stored 1
```

`_extract_by_definition` takes `target_type` from the questionnaire's `definitionExtract` `valueCode` and builds `{"resourceType": target_type}` from it. So the same request that is refused for `AllergyIntolerance` succeeds for `Consent`, and would for `Procedure`, `Immunization`, `MedicationStatement` or anything else not among the three named. A fabricated consent record written on an agent's credential is not a smaller problem than a fabricated allergy.

**Why I think it is the shape.** Three named types is a denylist, and the field it filters on is supplied by the caller. Every type nobody thought of is permitted by default, which is the failure mode you do not find by testing the case you already know about.

An allowlist inverts it: commit mode writes only the types explicitly cleared to be written without a human confirming them, and refuses everything else with the message this pull request already writes. New extraction targets then arrive refused, and someone has to decide, which is the decision this rule exists to force.

I have not opened a branch for it. Filed as #679 so it does not ride on this pull request; the probe is in a scratch worktree and is not proposed for merge.

Issue #568, opened 2026-09-03
MCP authorization phases 2 and 3: a hosted connector still cannot reach a tenant


aks129:
Successor to #290, filed as council ruling D4 requires when phase 1 lands.

Phase 1 (PR #556) serves protected-resource metadata and an enriched challenge behind `MCP_CANONICAL_RESOURCE`. That closes #523. It does **not** make a hosted connector able to reach a tenant, which is what #290 was really about, so this issue carries the remainder rather than letting #290 close on a partial.

## What is still missing

**Phase 2 — the authorization server has to behave like one.** The council's amendments P2-a to P2-e, now in `docs/specs/2026-08-16-mcp-authorization.md`, each name a specific gap verified against `r6/oauth.py`:

- `/oauth/authorize` returns the code as JSON instead of a `302` to the redirect URI with `code`, `state` and `iss`. A hosted connector opens that URL in a browser popup, so the user sees a JSON blob and the code never reaches the client.
- The tenant comes from a request header a browser cannot set, so a browser-initiated authorize has no way to say which tenant it means.
- No `resource` parameter is read at either endpoint, so audience restriction and `invalid_target` do not exist.
- Registration always answers `client_secret_post` while discovery advertises both, and the token endpoint never verifies a secret anyway.
- There is no introspection endpoint, and the credential the MCP server would use to call one has never been provisioned.

**Phase 3 — the tool-argument credential path.** Amendment P3-a: header passthrough is stopped, but the JSON-RPC tool-argument overrides are not, so on the OAuth path a caller could hand Flask a credential in a tool argument. The property to test is that no credential carried in a tool argument reaches Flask on the OAuth path.

**The consent story.** A hosted connector reaching a tenant means a person consenting to that, and the spec does not yet say what that consent looks like or where it is recorded.

## How this closes

The spec's §8.4 end-user run, performed by someone who is not us, against the deployed server. Not a curl transcript from the author.

## Gates before any of it deploys

DNS for the canonical host (#522), then `MCP_CANONICAL_RESOURCE` set, then the phase 1 verification. Phase 2 does not start until the council's two-week hold on MCP-auth building ends, per D4's settled dissent.

## Related

#290, #523, #522, PR #556, PR #560, council ruling 2026-09-02 D4.

PR #669, opened 2026-09-06
feat(oauth): the authorization server behaves like one (#568 phase 2, P2-a to P2-d)

aks129:
## What

Phase 2 of the MCP authorization design, the authorization-server half. After this a browser can finish our OAuth flow and a token knows what it is for. Everything is inert for existing callers except where a caller was relying on a defect (a JSON authorize body, an unverified client secret).

- `/oauth/authorize` answers **302** to the registered redirect URI with `code`, `state` (URL-encoded, only when sent) and RFC 9207 `iss`. Today it returns the code in a JSON body, which no connector popup can act on (#568).
- RFC 8707 `resource` is read at both endpoints. A known audience is recorded on the code and the token as `aud`; an unknown one is `invalid_target` back at the client, never recorded as sent. The token endpoint refuses a `resource` the code was not issued for and inherits it when absent (P2-c).
- The audience map (spec §3.5.1, P2-b): the FHIR resource keeps today's header-bound tenant behind the public-tenant guard; `MCP_CANONICAL_RESOURCE`, when set, binds `MCP_OAUTH_DEMO_TENANT` and **ignores the header**, failing closed when that tenant is not public.
- `r6.read_auth` accepts a bearer only when its `aud` is this FHIR surface. A token minted at the same issuer for the MCP audience is refused on every tenant (spec §9.2, R3 in reverse). This lands with the `aud` recording, not after it.
- `/oauth/register` (RFC 7591, P2-d): honours `token_endpoint_auth_method` (`none`, `client_secret_post`, `client_secret_basic`), issues no secret to a public client, reports `client_secret_expires_at` and `client_id_issued_at`, refuses a redirect URI that is neither `https://` nor a plain-http loopback, and loopback URIs match on any port (RFC 8252 §7.3, Claude Code's ephemeral port).
- `/oauth/token` authenticates the client: a public client must send `client_id`; a confidential client must present its secret, POST body or HTTP Basic, compared through `r6.constant_time.equal`. Today no secret is ever checked.
- The metadata is served at the issuer root, `/.well-known/oauth-authorization-server`, the first place a client looks for a path-less issuer, and every published URL derives from `OAUTH_ISSUER` when it is set. `authorization_response_iss_parameter_supported` is advertised.
- First commit: spec §13, the consent surface (the founder's 2026-09-06 direction: HealthClaw stays the issuer, CareAgents is the sign-in and consent front door, revocable consent, rotating refresh, a read-only downstream mint). Design only; the PRs that build it follow this one.

## Evidence

Full suite 3664 passed, 13 skipped, 1 xfailed. `tests/test_oauth_as_conformance.py` (26 tests) plus four existing test files moved from the JSON body to the redirect. Nine mutations executed on a separate worktree, every one caught:

| mutation | failing |
|---|---|
| `aud` comparison dropped in `_oauth_authorizes` | 4 |
| unknown `resource` recorded as sent | 2 |
| token endpoint ignores a `resource` mismatch | 1 |
| demo policy binds the header tenant | 2 |
| redirect matched on host only (port ignored everywhere) | 1 |
| client secret never compared | 2 |
| `iss` omitted from the redirect | 3 |
| public client's `client_id` not required | 1 |
| root discovery rule removed | 2 |

## Not in this PR

Introspection and the read-only mint (next), the CareAgents handoff, refresh tokens, the MCP server's OAuth path. No deploy step and no DNS. `MCP_CANONICAL_RESOURCE` unset on Flask keeps the audience map at today's single entry.

Closes nothing yet; #568 closes on the spec's §8.4 end-user run.

### DIFF
--- a/docs/specs/2026-08-16-mcp-authorization.md
+++ b/docs/specs/2026-08-16-mcp-authorization.md
@@ -1132,8 +1132,18 @@ end-user run.
   port.
 - A manual client id and secret can be pasted under "Advanced settings"; that
   path is not designed for and not needed.
-- Whether Claude uses the refresh grant is not stated. The server offers it
-  (§13.5); the end-user run reports whether it was taken.
+- **Claude registers as a confidential client**: its registration body carries
+  `token_endpoint_auth_method: client_secret_post`, and at the token endpoint
+  it sends `client_id` and `client_secret` in the form body, as
+  `application/x-www-form-urlencoded`, for the code exchange and for refresh
+  (Anthropic's connector authentication notes, read 2026-09-06). The MCP
+  Python SDK's client, which Claude's user agent identifies, sends the
+  credential the way the registration answered, refuses a registration that
+  names a secret-bearing method without a secret, uses the refresh grant, and
+  sends RFC 8707 `resource` when the protocol version calls for it. So the
+  token endpoint's client authentication (P2-d, #669) is what Claude expects,
+  and the refresh chain (§13.5) will be exercised. Still measured only by the
+  §8.4 run.
 
 ### 13.2 The shape: Flask stays the issuer, CareAgents is the front door
 
--- a/docs/specs/2026-08-16-mcp-authorization.md
+++ b/docs/specs/2026-08-16-mcp-authorization.md
@@ -1102,3 +1102,183 @@ refresh, and the missing `aud` check in `_oauth_authorizes`.
   `services/agent-orchestrator/src/index.ts`). On the
   OAuth path `_tenantId`, `_stepUpToken` and `_authorization` are discarded,
   exactly as `isPublicDemo()` pins them. Asserted as R8.
+
+---
+
+## 13. Amendment 2026-09-06: the consent surface, and a hosted connector reaching a real tenant (#568)
+
+**Status: proposed by the founder on 2026-09-06; design only in this section.**
+The founder asked for the locked server to work as an ordinary claude.ai
+connector on real records, with CareAgents as the account and consent surface.
+That request overrides the "phase 2 waits" clause of ruling D4. Everything in
+§1 to §12 stands; this section fills the one hole §6.1 and #568 left open.
+
+### 13.1 What claude.ai actually does, from Anthropic's own connector notes
+
+Read, not run, on 2026-09-06. Measured against our deployment only in the §8.4
+end-user run.
+
+- Claude registers itself with Dynamic Client Registration. The client name is
+  `Claude`. The callback is `https://claude.ai/api/mcp/auth_callback`, and
+  Anthropic says it may become `https://claude.com/api/mcp/auth_callback`, so
+  both are accepted as registered redirect URIs.
+- The authorize request carries `resource=` (the PRM `resource`, RFC 8707),
+  `scope=` (from the PRM), `code_challenge` with `S256`, and `state`. Claude
+  supports the 2025-03-26 and 2025-06-18 authorization specs, which is the
+  discovery chain §3 designs to.
+- Claude Code, as a native client, registers loopback redirect URIs
+  (`http://localhost/callback`, `http://127.0.0.1/callback`) on ephemeral
+  ports; redirect-URI matching for `localhost` and `127.0.0.1` must ignore the
+  port.
+- A manual client id and secret can be pasted under "Advanced settings"; that
+  path is not designed for and not needed.
+- Whether Claude uses the refresh grant is not stated. The server offers it
+  (§13.5); the end-user run reports whether it was taken.
+
+### 13.2 The shape: Flask stays the issuer, CareAgents is the front door
+
+The issuer is unchanged (`https://app.healthclaw.io`, D4). The MCP server
+still introspects at Flask (§3.5). What changes is what happens between a
+browser arriving at `/oauth/authorize` and a code being issued:
+
+```
+claude.ai ── GET /r6/fhir/oauth/authorize?...&resource=<RESOURCE> ──▶ Flask
+                Flask validates client, PKCE, resource; parks the request
+                302 ─▶ https://careagents.cloud/authorize?req=<signed handle>
+                                                                     │
+      CareAgents: passkey sign-in; consent page lists the account's   │
+      connections (each one a tenant) plus "the demo records";        │
+      approval requires a fresh passkey assertion; Grant row written  │
+                302 ◀─ /r6/fhir/oauth/consent/return?grant=<signed>  ─┘
+                Flask verifies the grant, binds the code to its tenant,
+                audits the consent under that tenant
+                302 ─▶ https://claude.ai/api/mcp/auth_callback?code&state&iss
+```
+
+Two extra top-level redirects. From the client's side it is ordinary OAuth
+2.1 with PKCE. CareAgents runs no authorization server: no registration, no
+token endpoint, no PKCE. It authenticates a person and records a decision.
+
+**The alternative, stated so it can be chosen instead:** CareAgents as the
+issuer. It would port registration, PKCE, the stores and revocation from
+`r6/oauth.py`, re-rule D4, and move the PRM's `authorization_servers`. Nothing
+the founder asked for needs that.
+
+### 13.3 The handoff protocol
+
+Both sides already share `INTERNAL_TOKEN_MINT_SECRET` (CareAgents mints
+step-up tokens with it). The handoff key is derived from it with domain
+separation, the way `payload_digest` derives from `STEP_UP_SECRET`:
+`sha256(b'healthclaw-consent-handoff:' + INTERNAL_TOKEN_MINT_SECRET)`. One
+fewer secret to provision, and a forged grant needs the mint secret, which
+already grants everything.
+
+**Outbound (Flask to CareAgents).** Flask stores the parked request under a
+random `request_id` (10 minutes, single use) with `client_id`, `client_name`,
+`redirect_uri`, `scopes`, `code_challenge`, `state`, `resource`. The URL
+carries `req=<request_id>.<exp>.<tag>`, where `tag` is HMAC-SHA256 over
+`request_id.exp` under the handoff key. The page shows the client name and
+scopes from the parked request, fetched by CareAgents from Flask
+(`GET /r6/fhir/oauth/consent/<request_id>`, service-authenticated with the
+mint secret), so nothing about the client rides in the URL.
+
+**Inbound (CareAgents to Flask).** `grant=<base64url(JSON)>.<tag>` where the
+JSON is `{request_id, tenant_id, consent_id, nonce, exp, decision}` and
+`decision` is `approved` or `denied`. Flask: `compare_digest` on the tag;
+`exp` in the future; `nonce` unused (the nonce cache in `r6/stepup.py`);
+the parked request popped exactly once; `tenant_id` matches
+`^[a-zA-Z0-9_-]{1,64}$`. Then the code is issued bound to `tenant_id`, or
+the client is sent `error=access_denied`.
+
+**Trust, written down as an assumption.** Flask cannot verify that the
+CareAgents account owns `tenant_id`; it trusts the signed assertion. That
+trust exists today: CareAgents holds the mint secret and can act on any
+tenant. This section adds no authority, it moves an existing one onto a
+signed, single-use, expiring message.
+
+**The tenant policy of the `invalid_target` map (§3.5.1) gains a third value.**
+`demo` binds `MCP_OAUTH_DEMO_TENANT`; `careagents` sends the request through
+§13.3 and binds whatever the grant names; the FHIR resource keeps binding the
+header tenant behind the existing public-tenant guard. With
+`CAREAGENTS_CONSENT_URL` unset, the MCP resource falls back to `demo`, which
+is the §3.5.1 behaviour unchanged.
+
+### 13.4 Consent is a tenant event, and it can be taken back
+
+- Flask writes an AuditEvent under the tenant on approval: `event_type=create`,
+  `resource_type=Consent`, `resource_id=<consent_id>`, detail
+  `client_id=… scopes=… via=careagents`. No account id, no email, no label.
+- Flask stores the consent (`consent_id` to `{tenant_id, client_id, scopes,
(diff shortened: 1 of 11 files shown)


aks129:
Adversarial QA pass on `feat/oauth-as-conformance` (head `43a8194`), driven against a live Flask instance on a real migrated SQLite DB, over real HTTP — full PKCE dance (register → authorize → token) plus direct OAuth-store injection for the two scenarios no legitimate front-door flow can produce (see below). `READ_AUTH_ENABLED=true`, `PUBLIC_TENANTS=desktop-demo`, `MCP_CANONICAL_RESOURCE`/`MCP_OAUTH_DEMO_TENANT` set throughout.

**All 23 scenarios correct — no bypass.**

**(1) Authorize endpoint**
- 302 to the registered redirect URI with `code`, `state` (echoed), and `iss` (RFC 9207) — confirmed on the actual `Location` header, not a JSON body.
- Unknown `resource` → `error=invalid_target` **on the redirect**, `state` still echoed.
- `resource=<MCP_CANONICAL_RESOURCE>` binds the configured demo tenant (`desktop-demo`) and **ignores** a genuinely different `X-Tenant-Id` header sent alongside it — verified by reading the stored auth-code/token record directly, not just trusting a 200.

**(2) Token endpoint**
- `resource` at token-exchange differing from the code's stored `aud` → `400 invalid_target`, and this check is reached (not shadowed by client auth) once the confidential client's secret is supplied.
- Confidential client with no secret → `401 invalid_client`; same flow with the correct secret → succeeds. (My first pass on this hit false failures from my own harness omitting the now-required secret — worth flagging only because it's a clean forcing function: the old two-argument token exchange silently stops working the moment a client registers as confidential, which is exactly the intended behavior change.)
- Public client (`token_endpoint_auth_method: none`) → registration response carries no `client_secret`; omitting `client_id` at token exchange → `400`.

**(3) Registration**
- Plain `http://` non-loopback → `400`; `https://` and loopback (`http://127.0.0.1:9999/cb`) → `201`.
- A loopback redirect matches at authorize time on a **different port** than registered (`:9999` registered, `:12345` presented) → succeeds. A non-loopback `https://` client does **not** get that leniency — an extra path segment on an otherwise-identical URI is refused (`400`, exact match required).

**(4) Audience isolation in `r6/read_auth.py`**
- **R3 in reverse**: minted a token at the MCP audience (bound to the public demo tenant, the only tenant the `demo` policy can ever bind), replayed it as a bearer against `GET /r6/fhir/Patient` for a genuinely different, private tenant → `401`.
- Isolated the **audience check specifically** (rather than relying on the tenant mismatch above to explain the refusal, since the `demo` policy can never mint for a private tenant through the front door): injected a token directly into the OAuth store with `tenant_id` **matching** the tenant being read and `aud=<MCP>` — refused (`401`). A control token with the identical shape but `aud=<correct FHIR resource>` — accepted (`200`). Isolates that the `aud` comparison itself is doing the work, not a coincidental tenant mismatch.
- **Near-misses**, both as the `resource` param at `/authorize` and as a directly-injected stored `aud` on an otherwise-valid token: trailing slash, upper-cased scheme+path, `http` vs `https`, and percent-encoded path separators (`%2Fr6%2Ffhir`). All four rejected in both positions — `resource_policies()`'s dict lookup and `read_auth`'s `!=` comparison are both exact-string, no normalization anywhere that a near-miss could ride through.

**(5)** `GET /.well-known/oauth-authorization-server` at the host root answers 200 with `issuer` equal to the bare host and the prefixed endpoint URLs, matching the `/r6/fhir`-prefixed copy.

Full suite on the branch: 3695 passed, 13 skipped, 1 xfailed — no regressions.

Confirmed absent, as flagged: no consent binding, no refresh token — this PR doesn't claim either.

Findings only; no push, no PR, no arm.

aks129:
Adversarial QA pass on the two branches stacked on this PR's head, driven against live Flask instances on real migrated SQLite DBs, over real HTTP (plus a few direct OAuth-store injections/in-process calls for the handful of scenarios no legitimate front-door flow can produce — noted inline below).

## Branch A — `feat/oauth-introspection-and-read-mint` (head `3c3c6a9`)

**16/16 scenarios correct.**

- **Introspection, unconfigured** (`MCP_INTROSPECTION_CLIENT_ID`/`_SECRET` unset): refuses everybody with `401`, including a guess at plausible-looking creds.
- **Introspection, configured**: wrong secret → `401`. Missing token, unknown token, an expired token, and a revoked token all answer **exactly** `{"active": false}` — same shape for all four, no timing/detail leak between "doesn't exist" and "expired." A genuinely live token answers the full shape (`aud`, `tenant_id`, `scope`, `client_id`, `exp`) and the raw token value never appears anywhere in the response body. Client auth via HTTP Basic works identically to the POST-body form.
- **Read-only internal mint**: `scope="write"` and `scope="anything-else"` both `400`; `scope="read"` mints a token that reads a private tenant (`READ_AUTH_ENABLED=true`) and is refused (`401`, `"Read-scoped token cannot authorize this operation"`) on a direct `POST /r6/fhir/Observation` (with `X-Human-Confirmed`) and on `POST /r6/actions/<id>/commit`. `POST /r6/fhir/QuestionnaireResponse/$extract` (commit mode) also refuses it (`401`), though its message collapses to the generic `"Invalid step-up token"` rather than the specific reason the other two sites surface — `r6/sdc/routes.py` calls `validate_step_up_token` directly rather than through the kernel's reason-publishing path, so this is a pre-existing diagnostic-clarity gap in that route, not something this branch introduces or a security issue (still a clean `401` either way).
- One clarification for my own test design: `POST /r6/actions/propose` takes no step-up at all by contract (tenant header only) — the write-gated step in that rail is `/commit`, which is what I tested against the read-scoped token.
- The mint gate itself (`X-Internal-Secret` for a non-public tenant) is unchanged — still `403` without it.

Full suite: 3715 passed, 13 skipped, 1 xfailed.

## Branch B — `feat/oauth-consent-handoff` (head `4af8319`, on A)

**19/19 scenarios correct.**

- The park redirect's tag (`req=<request_id>.<exp>.<tag>`) matches `handoff_tag(f"{request_id}.{exp}")` computed independently from the shared key.
- `GET /oauth/consent/<request_id>` refuses both no-secret and wrong-secret with `403`; correct secret returns the parked request's `client_id`/`scopes`.
- **Forged tag**: four guessed keys (`"guessed-secret"`, empty string, `"admin"`, the real secret plus one character) all refused `400`, against a correctly-shaped payload — a control grant signed with the real key on a fresh parked request succeeds right after, so the refusals aren't from a shape problem.
- **Tampered body under a valid tag**: took a legitimately-signed grant, decoded the body, changed `tenant_id`, re-encoded, and kept the *original* tag — refused `400`. The HMAC covers the body's actual bytes, not just its presence.
- **Expired grant** → `400`.
- **Same nonce on two parked requests**: first decision succeeds (`302`), the second — different request, different tenant, same nonce — refused `400`. `mark_nonce_used` is checked before the parked-request lookup, so this is a real replay guard, not incidental.
- **Same parked request decided twice**: first decision succeeds, a second decision for the same `request_id` (fresh nonce this time) is refused `400` — the parked record is popped on first use regardless of nonce freshness.
- **`tenant_id` with path or shell metacharacters** (`../../etc/passwd`, `tenant; rm -rf /`) both refused `400` by the tenant-id shape check.
- **`X-Tenant-Id` header on the return request is ignored**: sent a grant naming `tenant-from-grant` alongside a header claiming `tenant-from-HEADER-should-be-ignored`; the issued token is bound to `tenant-from-grant` (verified against the stored token record, not just a 200).
- **Audit content**: the `Consent` `AuditEvent` written on approval has `event_type=create`, `agent_id=<client_id>`, and `detail="client_id=<id> scopes=<scopes> via=careagents"` — nothing else. Checked that neither the tenant id nor the consent id (both of which live in their own columns) leak into the free-text `detail`, and there's no account/person identifier anywhere on the row.
- **Revoke kills both at once**: minted a token through the full consent flow, confirmed `validate_bearer_token` (the one function both `read_auth._oauth_authorizes` and `/introspect` call) and `/introspect` both read it as live, called `POST /oauth/consent/<id>/revoke`, confirmed both flip to dead in the same instant — `validate_bearer_token` now returns `(False, "Consent revoked")` and introspection answers `{"active": false}`.
  - Methodology note: this consent-issued token is bound to the **MCP** audience (the flow only runs for `resource=MCP_CANONICAL_RESOURCE`), so it correctly fails a *direct* `GET /r6/fhir/Patient` bearer-read regardless of revoke state — that's PR #669's own R3 audience check doing its job, unrelated to consent, and conflating the two would have been a false read on my part. Used the shared `validate_bearer_token` call directly (in-process) to isolate the consent-revocation property specifically, and added a control confirming an ordinary FHIR-audience token (unrelated to this consent, minted via the existing auto-approve path) still reads normally — this branch doesn't regress the ordinary read path.
  - `revoke` itself needs `X-Internal-Secret` (`403` without) and is idempotent (`200` on a second call).

Full suite: 3733 passed, 13 skipped, 1 xfailed.

No bypass found on either branch. Findings only; no push, no arm.

aks129:
Adversarial QA pass on `feat/oauth-refresh-rotation` (head `143b847`, on B), driven against a live Flask instance on a real migrated SQLite DB, over real HTTP.

**17/17 scenarios correct — plus two measured near-misses worth calling out explicitly (not blocking, but real properties, not assumptions).**

- Code grant now also returns `refresh_token`; refreshing returns a **new** access+refresh pair and the presented refresh token is immediately unusable.
- **Reuse detection**: replaying the already-spent token is refused, and — the specific claim in the brief — the **current, never-yet-used** refresh token from that same chain is *also* dead afterward (verified by attempting to refresh with it right after the replay: refused).
- **Client binding**: another client's id+secret refused; a confidential client with no secret refused (`401`); a public client with no `client_id` refused (`400`).
- **Scope**: narrowing succeeds; widening → `invalid_scope`.
- **Resource**: a `resource` other than the chain's `aud` → `invalid_target`; the matching resource succeeds.
- **Consent linkage**: revoking the consent behind a chain kills the next refresh (`400`) — via `consent_is_live`, same as the read/introspection path in the previous branch.
- **`POST /oauth/revoke` with a refresh token** ends its chain (a later refresh with the same chain's current token fails).
- **Two clients racing the same refresh token** (`ThreadPoolExecutor`, `threaded=True` dev server): exactly one gets `200` with a fresh pair, the other `400` — `dict.pop` is atomic under the GIL for the in-memory store this exercises; a Redis-backed deployment would need to confirm the same atomicity on its own `pop`, which this test can't reach.
- **Store key structural check**: no route anywhere iterates `_refresh_tokens`/`_refresh_used`/`_refresh_chains` by tenant or client — the only access pattern in `r6/oauth.py` is a single-key get/set/pop keyed by the 48-byte `secrets.token_urlsafe` value itself. There's no derivable or enumerable path to another tenant's refresh record from outside.

**Two measured properties, stated as facts rather than either a pass or a fail, since they're real design tradeoffs rather than obviously-wrong behavior:**

1. **The access token issued alongside a since-revoked refresh token keeps working until its own natural 1-hour TTL.** Reuse detection (`_revoke_refresh_chain`) only touches the `refresh-token`/`refresh-chain` stores — it never reaches into `access-token`. So "the newest token dies too" is true for the refresh side of a rotation, not the access-token side. Confirmed via `validate_bearer_token` directly (the same function `read_auth` and `/introspect` both call): the access token minted in the same rotation as the now-revoked refresh token still validates as live. Worth a line in the spec if this is the intended tradeoff (a max 1h exposure window vs. the 30-day refresh danger) — it reads as deliberate given the access-token TTL is already short, but the code doesn't say so.
2. **Presenting a refresh token to the WRONG client burns it for the legitimate client too — no correct credentials needed.** `_refresh_grant` pops (spends) the presented token from the store *before* `_client_authenticates` runs, per its own docstring ("the presented token is spent whatever happens next"). So an attacker who has only observed/guessed a refresh token value — zero client credentials — can single-shot deny the real client's ability to rotate, by presenting it once with any wrong `client_id`. Distinct from the reuse-detection chain revocation (which needs a token that was already legitimately rotated once): this fires on the very first, otherwise-unauthenticated presentation. Confirmed live: a wrong-client attempt against a fresh token refuses (`400`/`401` depending on which check trips first), and the real client's immediate retry with the same token is then also refused (`400`).

Full suite on the branch: 3747 passed, 13 skipped, 1 xfailed.

aks129:
Two things checked after review, before anyone arms this: (1) no non-test consumer in the repository reads the old JSON authorize body or exchanges a code without a secret (grep of every .py/.js/.ts/.html/.md outside tests: only comments in main.py and the tests already moved), so the behaviour change has no live caller to break; the one external caller that matters, claude.ai, registers as client_secret_post and sends the secret in the form body, which is exactly what the token endpoint now requires (spec §13.1, last commit). (2) This is the one PR in the connector set that is live on merge: Flask auto-deploys from main, and the redirect and client-authentication changes take effect at once for any client registering from then on. #670 and #671 are inert until their flags and the Flask branches land.

aks129:
Re-probe on `feat/oauth-refresh-rotation` at the new head (`6568f54`), targeting only the two behaviors that changed since my earlier pass on `143b847` — everything else on that earlier comment is unchanged and still holds.

**4/4 correct:**

- **Wrong-client presentation no longer burns the token.** A wrong client_id + secret presenting a valid refresh token is refused (`401`/`400`); the real client then successfully refreshes with the SAME token right after (`200`, fresh pair). Confirmed the fix moved the authentication check ahead of the spend, as intended.
- **Control — a genuine double-refresh race still cuts the chain.** Two concurrent presentations of the *same*, never-before-used, valid refresh token (via `ThreadPoolExecutor`, real HTTP, `threaded=True` dev server): exactly one gets `200`, the other `400` — the OAuth 2.1 reuse rule holds for the case it's actually meant for, distinct from the wrong-client case above.
- **Chain revoked by reuse kills every access token it issued.** Minted a token, refreshed it once (two access tokens now exist from the same chain — verified both live via `validate_bearer_token` directly, the function both `read_auth` and `/introspect` consult), then replayed the already-spent first refresh token to trigger reuse detection. Both access tokens — the one from before the rotation and the one from after — stop validating.
- **Chain revoked by `POST /oauth/revoke` kills every access token it issued.** Same two-access-token setup, revoked via the current refresh token instead of a replay. Same result: both access tokens die.

Full suite on the branch: 3747 passed, 13 skipped, 1 xfailed.

PR #670, opened 2026-09-06
feat(careagents): the consent page for the MCP connector, with a fresh passkey at approval (#568)

aks129:
## What

The CareAgents half of the MCP connector's consent handoff (spec §13.2, §13.4, §13.7 in `docs/specs/2026-08-16-mcp-authorization.md`, the §13 amendment is on #669). HealthClaw parks a connector's OAuth request and sends the browser to `/authorize` on careagents.cloud with a signed handle. This page is what the person sees and decides on.

- **Nothing is fetched until the handle verifies.** The tag is checked under the key derived from the mint secret both services already hold (`sha256("healthclaw-consent-handoff:" + secret)`), then the expiry. A forged link is a 400 that never calls HealthClaw.
- **Sign-in resumes the request.** Signed out, the handle is parked in the session and the person goes to `/auth`; `/home` then resumes it once.
- **The page says who is asking and what they get, in the person's words**, and offers the account's own connections. Real-record connections are offered only where `CARE_REAL_RECORDS` opens them, exactly as connecting them is gated; with the gate closed the decision endpoint refuses a real one even by id.
- **Nothing is granted without a fresh passkey assertion with user verification, for this account.** `/webauthn/consent/options` asks for `userVerification: required`; `/authorize/decide` pops the challenge (no challenge, no approval), verifies the assertion with `require_user_verification=True`, and refuses a passkey that verifies as another account. An account without a passkey is offered enrolment, not approval. A denial needs no passkey.
- **The signed decision goes back through the browser** in the format HealthClaw's `decode_grant` verifies (sorted keys, no whitespace, a fresh nonce, five-minute expiry), to `<HEALTHCLAW_PUBLIC_BASE>/r6/fhir/oauth/consent/return`.
- **A `Grant` row records it** (account, connection, tenant pointer, client id and name, scopes, HealthClaw's consent id, granted and revoked times). The hub lists grants under "Shared with other apps"; **Revoke asks HealthClaw first** and marks the row only on its yes, so a revocation is never shown that did not happen (502 and still active otherwise).
- No PHI anywhere in this: a request id, a tenant pointer, a client name, connection labels.

New: `careagents/consent.py`, `careagents/templates/consent.html`, `careagents/static/consent.js`, the `Grant` model, four routes, client methods `consent_request` and `revoke_consent`.

## Evidence

`tests/test_careagents_consent.py` (15 tests). The cross-layer rows that verify a grant with HealthClaw's own decoder skip until the handoff PR is on main and then run for real; a reference decoder holds the format meanwhile. Full suite: 3646 passed, 13 skipped, 1 xfailed, one guard allowlisted with its reason (`careagents/consent.py` compares two byte strings it built itself). Eight mutations executed on a separate worktree, all caught:

| mutation | failing |
|---|---|
| handle tag never compared | 3 |
| handle expiry ignored | 1 |
| real records offered when the gate is closed | 1 |
| decide ignores the real-records gate | 1 |
| another account's passkey approves | 1 |
| approval without a fresh challenge | 1 |
| revocation claimed when HealthClaw failed | 1 |
| presence accepted instead of user verification | 1 |

## Not in this PR

The HealthClaw endpoints this page talks to are on the held branch `feat/oauth-consent-handoff` (opens after #669 and the introspection PR land). Until then the page is reachable but every handle fails to verify without a matching secret, and nothing here runs on its own. No deploy: CareAgents is a manual `railway up` from a staged directory.

### DIFF
--- a/careagents/accounts.py
+++ b/careagents/accounts.py
@@ -25,7 +25,7 @@ from webauthn.helpers.structs import (AuthenticatorSelectionCriteria,
                                       UserVerificationRequirement)
 
 from careagents import mail
-from careagents.models import (Account, Agent, Connection, EmailToken, Passkey,
+from careagents.models import (Account, Agent, Connection, EmailToken, Grant, Passkey,
                                Surface, UsageDay, make_engine,
                                make_session_factory, now)
 
@@ -234,15 +234,20 @@ class AccountService:
 
     # --- WebAuthn: authentication -------------------------------------------
 
-    def authentication_options(self) -> tuple[dict, str]:
+    def authentication_options(self, require_uv: bool = False) -> tuple[dict, str]:
+        """`require_uv` asks the authenticator for user verification (face,
+        fingerprint, PIN) rather than mere presence: the consent step wants
+        the person, not the device."""
         opts = webauthn.generate_authentication_options(
             rp_id=self.cfg.rp_id,
-            user_verification=UserVerificationRequirement.PREFERRED)
+            user_verification=(UserVerificationRequirement.REQUIRED if require_uv
+                               else UserVerificationRequirement.PREFERRED))
         return (_opts_to_dict(webauthn.options_to_json(opts)),
                 bytes_to_base64url(opts.challenge))
 
     def finish_authentication(self, credential: dict,
-                              expected_challenge: str) -> Account:
+                              expected_challenge: str,
+                              require_uv: bool = False) -> Account:
         raw_id = base64url_to_bytes(credential["rawId"])
         with self.session() as s:
             pk = s.query(Passkey).filter_by(credential_id=raw_id).first()
@@ -255,6 +260,7 @@ class AccountService:
                 expected_origin=self.cfg.origin,
                 credential_public_key=pk.public_key,
                 credential_current_sign_count=pk.sign_count,
+                require_user_verification=require_uv,
             )
             pk.sign_count = verification.new_sign_count
             acct = s.get(Account, pk.account_id)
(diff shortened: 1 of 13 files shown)


aks129:
Adversarial QA pass on `feat/careagents-consent-page` (head `02bf42e`), driven against two REAL live processes talking real HTTP: HealthClaw as a subprocess from a `feat/oauth-refresh-rotation` checkout (has the consent handoff + introspection this branch's CareAgents side needs), and CareAgents in-process from this branch's own checkout, on a real migrated SQLite DB on each side. The passkey ceremony is stubbed (`AccountService.finish_authentication` monkeypatched) exactly as flagged in the brief — no hardware authenticator involved, but every branch around it is exercised for real: real email-code signup (captured before hashing, not read back from the DB), real sessions, real cookies, real HTTP round-trips between the two apps, real HealthClaw-side OAuth/consent/introspection.

**13/13 scenarios correct.**

- **Full loop**: `authorize?resource=<MCP>` on HealthClaw → 302 to CareAgents `/authorize?req=...` → (signed in) consent page renders the offered (sample) connection → `POST /webauthn/consent/options` → `POST /authorize/decide` (approved, stubbed passkey verifying the right account) → redirect to HealthClaw's `/oauth/consent/return` → code → token → `/introspect` shows `active: true` with `tenant_id` equal to the **chosen connection's own tenant** (verified against CareAgents' own DB row, not just trusted from the response).
- **No `wa_consent_challenge` in session** (skipped the options call) → `400`.
- **A passkey verifying another account** → `403` ("That passkey belongs to another account"), tested by stubbing the verifier to return account 2 while account 1's session is driving the request.
- **A connection belonging to another account** → `404` — this check runs before the passkey stage in the route, confirmed by triggering it with no challenge set at all and still getting the connection-ownership refusal, not a challenge error.
- **A real-record connection with `CARE_REAL_RECORDS=off`** → `403`. Had to flip the switch on just long enough to create a `direct`-kind connection (the API refuses to create one while off), then flip it back off before the consent decision — `Config` is resolved once at `create_app()`, so this meant mutating the live `Config` object CareAgents' routes already hold a reference to, not just the env var. A control right after (same connection, switch back on) succeeds, confirming the 403 really tracks the current switch state and not something else about that connection.
- **Revoke, HealthClaw first**: `POST /api/grants/<id>/revoke` calls HealthClaw's revoke, HealthClaw's `/introspect` immediately shows `active: false`, and the CareAgents-side grant row flips to `revoked` — in that order.
- **A HealthClaw failure leaves the grant shown active, `502`**: pointed CareAgents' `HealthClawClient` at a dead port (mutating both `.base` and `.fhir` — the client caches `.fhir` at construction rather than re-deriving it, a gotcha worth knowing about if anyone else pokes at this client the same way) and confirmed the local grant row stays `active` rather than being marked revoked on a failed call.
- **PHI on the consent page**: swept the rendered HTML for SSN/diagnosis/medication keywords, the account's own email, and its raw account id — none present. What's shown is the connection label, provider name, and client name, as designed.

Full suite on the branch: 3677 passed, 13 skipped, 1 xfailed — no regressions.

aks129:
Review found a defect the suites could not: `Grant.connection_id` is a foreign key to `ca_connections`, and `delete_connection` removed the connection row, so on Railway Postgres the first delete of a connection that carried a grant would have raised (SQLite ignores the key, and no test deleted a connection with a grant). Fixed in the amended head 3cf9449: the delete route now revokes every active grant at HealthClaw before unlinking, marks them, and `delete_connection` detaches the grant rows (kept as the record of what was shared and taken back) before the connection goes. A revoke that cannot be confirmed answers 502 with `deleted: true, unlinked: false, grants_active: n` and keeps the connection listed so Delete can be retried. Two tests added to `tests/test_careagents_consent.py`, which runs on both CI lanes. One near-vacuous assertion in the sign-in resume test (an `or` that could not fail) tightened at the same time.

aks129:
Full suite on the amended head 3cf9449: 3648 passed, 13 skipped, 1 xfailed (both new delete-path tests included; they run on the Postgres lane in CI as well).

aks129:
Re-probe on `feat/careagents-consent-page` at the new head (`3cf9449`), targeting only the delete path — everything from my earlier pass on `02bf42e` is unchanged and still holds. Driven against the same two-real-process setup as before (HealthClaw subprocess with the consent handoff, CareAgents in-process, real HTTP between them).

**2/2 correct:**

- **`DELETE /api/connections/<id>` with an active grant, HealthClaw's revoke succeeds**: `hc.revoke_consent` called and confirmed before anything is unlinked — status `200`, `deleted: true`. Checked CareAgents' own DB afterward: the connection is gone from `get_connection` (unlinked), and the grant row still exists with `connection_id: null` and `status: "revoked"` — detached, not deleted, exactly as designed for the audit trail.
- **Same, with HealthClaw's revoke failing**: patched just `revoke_consent` on the live client instance to raise (leaving `purge_tenant` — the actual record deletion — untouched, so this isolates the *consent* revoke specifically, not the whole HealthClaw client) → `502`, `deleted: true`, `unlinked: false`, `grants_active: 1`. The connection is still listed (`get_connection` still returns it) and the grant row is still `active` — nothing was falsely marked revoked on a call that didn't confirm it.

Full suite on the branch: 3679 passed, 13 skipped, 1 xfailed.

PR #671, opened 2026-09-06
feat(mcp): the server accepts authorization-server tokens for a consented tenant, behind MCP_OAUTH_ENABLED (#568 phase 3)

aks129:
## What

Phase 3 of the MCP authorization design (spec §3.5, §3.6, §7 and the §13.6 amendment on #669). Off by default; nothing changes until `MCP_OAUTH_ENABLED=true`, and enabling it without every dependency refuses to start.

- **A bearer that is not `MCP_AUTH_TOKEN` is introspected at Flask** (`POST <FHIR_BASE_URL>/oauth/introspect`, the pre-registered client credential) and accepted only when it is live, its `aud` is exactly `MCP_CANONICAL_RESOURCE`, it carries a read scope and it has not expired. Every other case is the same 401 as today, never a 403 (a wrong audience is a token that was never for us). Flask unreachable is a 401: introspection fails closed.
- **The result is cached under the token's SHA-256 only**, for at most five minutes and never past the token's expiry, so a revoked consent stops being served within the window and the token value is never held.
- **On that path the caller's credentials never reach Flask.** The tenant comes from introspection and nowhere else; the inbound `Authorization`, any step-up token and the SHARP bring-your-own-FHIR headers are dropped; the six credential tool arguments are discarded on both transports (P3-a, R8), with the SSE override logic lifted into a pure helper so the property is asserted directly.
- **The downstream credential is a read-scoped step-up token this server mints** for the consented tenant (`scope: "read"`, the mint's new argument from the introspection branch), cached under the token hash and never past its expiry; a mint that answers without the read scope is a refusal, not a wider call. **A write-tier tool is refused before anything is sent**: a consent to read is not a consent to write.
- **The 401 and the transport preflight are readable from a browser** (`Access-Control-Allow-Origin: *`, `WWW-Authenticate` exposed), which is part 2 of #523. The transport's own origin allowlist is untouched, so nothing the transport returns becomes readable by any other page.
- `/health` reports `oauth.enabled`. Boot: `MCP_OAUTH_ENABLED=true` requires a canonical `MCP_CANONICAL_RESOURCE`, `MCP_INTROSPECTION_CLIENT_ID`, `MCP_INTROSPECTION_CLIENT_SECRET` and `INTERNAL_TOKEN_MINT_SECRET`.

## Evidence

`src/oauth-path.test.ts` (23 tests, Flask mocked at the fetch layer so every assertion is about what this server sends and refuses): R1, R2, R3, R4 with the §8.3 near-misses, R5, R6, R7, R8 on both transports, the positive chain with the exact headers Flask receives, the write refusal, the fail-closed mint, the caches. Whole jest suite 264 passed; `tsc --noEmit` clean. Mutations executed on the same worktree:

| mutation | result |
|---|---|
| audience never checked | caught (3) |
| writes allowed on a consented connection | caught (2) |
| a full-capability mint accepted as the read credential | caught |
| OAuth path on regardless of the flag | caught |
| expiry not enforced at the resource | caught |
| 401 unreadable from a browser | caught |
| SSE overrides honoured on the OAuth path | caught |
| extractHeaders falls through and the grant dispatch is gone (two-ended) | caught |
| extractHeaders alone keeps `Authorization` on the grant path | **inert by construction**: the grant path builds Flask's headers from scratch and never copies it, so the test stays green; the two-ended row above is the one that measures the property |

## Not in this PR

Flask's introspection endpoint and the read-only mint argument are on the held branch `feat/oauth-introspection-and-read-mint` (opens after #669). Until both are deployed and the flag is set, this server behaves exactly as today. No deploy: the MCP server deploy is a founder-authorized step, and production enablement comes after the staging walkthrough (spec §7).

### DIFF
--- a/services/agent-orchestrator/src/index.ts
+++ b/services/agent-orchestrator/src/index.ts
@@ -314,6 +314,44 @@ function challengeHeader(req: express.Request): string {
   return `Bearer ${params.join(", ")}`;
 }
 
+// --- Rate Limiting (in-memory, per IP) ---
+
+const rateLimitMap = new Map<string, { count: number; resetAt: number }>();
+const RATE_LIMIT_WINDOW_MS = 60_000;
+const RATE_LIMIT_MAX = parseInt(process.env.RATE_LIMIT_MAX || "120", 10);
+const SESSION_TTL_MS = 30 * 60 * 1000;
+
+function checkRateLimit(ip: string): boolean {
+  const now = Date.now();
+  const entry = rateLimitMap.get(ip);
+  if (!entry || now > entry.resetAt) {
+    rateLimitMap.set(ip, { count: 1, resetAt: now + RATE_LIMIT_WINDOW_MS });
+    return true;
+  }
+  entry.count++;
+  return entry.count <= RATE_LIMIT_MAX;
+}
+
+app.use((req, res, next) => {
+  // Preflight answers below without reaching a handler, and it did not consume
+  // budget when the limiter sat further down the chain. Keep it that way: a
+  // browser client spends its allowance on requests, not on asking permission.
+  if (req.method === "OPTIONS") return next();
+  const clientIp = req.ip || req.socket.remoteAddress || "unknown";
+  if (!checkRateLimit(clientIp)) {
+    return res.status(429).json({
+      jsonrpc: "2.0",
+      error: { code: -32000, message: "Rate limit exceeded" },
+    });
+  }
+  next();
+});
+
+// Rate limiting comes before the credential check on purpose. The check now
+// makes a network call of its own (introspection), so a limiter placed after
+// it lets an unauthenticated flood drive unbounded requests at the
+// authorization server. A caller over the limit is refused here, having cost
+// this server one map lookup and Flask nothing.
 // Health probes and CORS preflight remain public. When configured, every MCP
 // network transport requires the deployment-scoped bearer credential.
 app.use(async (req, res, next) => {
@@ -387,35 +425,6 @@ app.use((req, res, next) => {
       res.setHeader("Access-Control-Allow-Origin", "*");
     }
     return res.sendStatus(204);
-  }
-  next();
-});
-
-// --- Rate Limiting (in-memory, per IP) ---
-
-const rateLimitMap = new Map<string, { count: number; resetAt: number }>();
-const RATE_LIMIT_WINDOW_MS = 60_000;
-const RATE_LIMIT_MAX = parseInt(process.env.RATE_LIMIT_MAX || "120", 10);
-const SESSION_TTL_MS = 30 * 60 * 1000;
-
-function checkRateLimit(ip: string): boolean {
-  const now = Date.now();
-  const entry = rateLimitMap.get(ip);
-  if (!entry || now > entry.resetAt) {
-    rateLimitMap.set(ip, { count: 1, resetAt: now + RATE_LIMIT_WINDOW_MS });
-    return true;
-  }
-  entry.count++;
-  return entry.count <= RATE_LIMIT_MAX;
-}
-
-app.use((req, res, next) => {
-  const clientIp = req.ip || req.socket.remoteAddress || "unknown";
-  if (!checkRateLimit(clientIp)) {
-    return res.status(429).json({
-      jsonrpc: "2.0",
-      error: { code: -32000, message: "Rate limit exceeded" },
-    });
   }
   next();
 });
@@ -1120,6 +1129,10 @@ if (require.main === module) {
   });
 }
 
+function resetRateLimitForTests(): void {
+  rateLimitMap.clear();
+}
+
 function closeMCPServerForTests(): void {
   clearInterval(sessionCleanupInterval);
   rateLimitMap.clear();
@@ -1143,6 +1156,7 @@ function getRuntimeStateForTests(): {
 
 export {
   app,
+  resetRateLimitForTests,
   assertMCPAuthConfigured,
   assertMCPCanonicalResourceConfigured,
   assertMCPOAuthConfigured,
--- a/services/agent-orchestrator/src/index.ts
+++ b/services/agent-orchestrator/src/index.ts
@@ -28,6 +28,7 @@ import {
 import crypto from "crypto";
 import { FHIRTools } from "./tools";
 import { executeMCPTool } from "./mcp-tool-result";
+import { fetchWithTimeout } from "./fetch-timeout";
 
 const { version: SERVER_VERSION } = require("../package.json") as {
   version: string;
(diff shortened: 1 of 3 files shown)


aks129:
Review found a gap the first cut left open: an SSE session captures its headers at connect, so a later POST to `/messages` authenticated with a different valid token (another tenant's, or the static one) would have executed as the tenant the session was opened for. Amended head f9cdabc binds every SSE session to the credential that opened it (the static token, or one OAuth token's hash) and refuses a mismatch with 401; streamable HTTP resolves its headers per request and needs no binding. Tested over a real socket: the stream is opened with one token, then `/messages` is posted under another accepted token and under the static token (both 401) and under the opening token (not refused). Suite now 266.

aks129:
Adversarial QA pass on `feat/mcp-server-oauth-path` (head `afeb59a`), driven against two REAL live processes: Flask on a `feat/oauth-refresh-rotation` checkout (subprocess, real migrated SQLite DB) and the Node MCP server (subprocess, `npm run dev`) on this branch, talking real HTTP to each other. `Host: mcp.healthclaw.io` set explicitly per-request (not `/etc/hosts`) — `hostIsCanonical()` only compares the request's `Host` header against `MCP_CANONICAL_RESOURCE`'s hostname, so that's sufficient to enrich the challenge.

**Methodology note**: used the `/mcp/rpc` JSON-RPC HTTP bridge rather than a full Streamable-HTTP `initialize` handshake for `tools/list`/`tools/call` — `/mcp/rpc` matches `isMCPTransportPath` (`/mcp/`-prefixed) so it passes through the exact same auth middleware, `extractHeaders`, and `executeAsGrant` code as the real transport; the `initialize` handshake itself is protocol negotiation, not part of the authorization surface this PR changes. Didn't attempt a real MCP SDK client session for that reason.

**10/10 scenarios correct — no bypass.**

- **Full loop**: registered a public client, `authorize?resource=<MCP>` (demo policy, no `CAREAGENTS_CONSENT_URL` set) → 302 → token → `tools/list` (27 tools) → `fhir_search` (`resource_type: Patient`) succeeds → the resulting `AuditEventRecord` row (read straight from Flask's SQLite file) has `agent_id = oauth:<client_id>`.
- **A token minted at the FHIR audience, replayed at the MCP server** → `401`, and specifically `401` not `403` (confirmed the exact status, not just "not 200" — the brief calls this out because a `403` would tell a prober their token is recognized here at all).
- **`MCP_OAUTH_ENABLED=false` with the same otherwise-valid token** → `401`, and Flask's log shows the identical count of `oauth/introspect` lines before and after the attempt — the server genuinely never asked, not just discarded the answer.
- **Write tools on the OAuth session** (`fhir_commit_write`, `action_propose`), each called with `_stepUpToken`/`_authorization` carrying a unique marker string in their arguments: both refused before `tool.validate()` or the handler ever runs (`executeAsGrant`'s write-tier check returns immediately), confirmed by the marker appearing in neither Flask's log tail nor any recent `AuditEventRecord` row.
- **Revoke + cache window**: token worked, `POST /oauth/revoke`, restarted the MCP process (the explicitly-allowed stand-in for waiting out the 5-minute cache), same token now `401`.
- **Flask down**: a freshly-minted, never-before-used token (so nothing could be served from cache) against a freshly-restarted MCP process with Flask killed → `401`.
- **Browser-facing**: `OPTIONS /mcp` with `Origin: https://claude.ai` → `204` with `Access-Control-Allow-Origin: *`; an unauthenticated `POST /mcp/rpc` from the same origin → `401` carrying both `WWW-Authenticate` (naming `resource_metadata` and `scope`) and `Access-Control-Allow-Origin`, so a browser can actually read the challenge rather than having the CORS layer swallow it.

**Deferred, not run**: the CareAgents-consent-surface variant of this same chain (through `/authorize` on #670's CareAgents app rather than the demo-tenant auto-approve). I validated that consent flow in isolation on #670's own thread and the MCP-server OAuth path here in isolation; wiring all three processes (Flask + CareAgents + Node) together for one combined pass is a larger lift I didn't take on this round — flagging it explicitly rather than silently skipping it.

Node suite on the branch: 264 passed (10 suites).

aks129:
Re-probe on the new head (`f9cdabc`) for the one behavioral delta since `afeb59a`: SSE session binding. Everything from the earlier comment on this PR is unchanged and still holds (didn't re-run all of it — the diff between the two heads is exactly the 16-line SSE-binding addition plus its own tests).

Driven live: `GET /sse` with OAuth token A (real streaming connection held open in a background thread — necessary, since a first pass of this test broke `resp.iter_lines()` early to capture the `sessionId` line and re-entered the generator, which released the underlying connection back to the pool and tore down the SSE session before the `/messages` POST ever ran; fixed by keeping one continuous iterator for the connection's whole lifetime).

**4/4 correct:**

- `GET /sse` with token A → session opens, `sessionId` captured from the endpoint event.
- The **same** credential (token A) `POST /messages?sessionId=...` → accepted (`202`).
- A **second OAuth token, for a different tenant**, `POST /messages` on the session token A opened → `401 Unauthorized`.
- The **static `MCP_AUTH_TOKEN`** `POST /messages` on that same OAuth-opened session → also `401` — the delta's own comment names this case too (`session.credential !== sessionCredential(...)`), confirmed live rather than just read off the diff.

Node suite: 266 passed (10 suites, +2 from this delta's own tests).

aks129:
## CodeQL found something real, and the fix does not clear the check

**The finding.** Two `js/missing-rate-limiting` alerts, both high. Phase 3 gave the credential check a network call: an unresolved bearer token costs a POST to the authorization server's introspection endpoint. The limiter sat *below* that check, so an unauthenticated flood was introspected once per request and this server amplified it at Flask. That is a genuine defect this branch introduced, not a scanner artifact.

**The fix** (b9d76f8). The limiter moved above the credential check. Past the limit a caller is refused after one map lookup and Flask is not called at all. Preflight is exempt, matching what it cost before, when CORS answered above the limiter.

Two tests pin it, and the ordering one is mutation-checked: moving the limiter back below the check turns it red. Full server suite 268/268, `tsc --noEmit` clean.

**The check stays red anyway.** Both alerts came back at the new line numbers after the fix. The query recognizes rate limiting from known middleware packages, and this server's limiter is hand-rolled, so the arrangement is invisible to it. No ordering of correct code clears this alert.

**Two ways to close it, owner's call.**

1. Dismiss both alerts as false positives, citing this comment. Nothing changes in the code.
2. Replace the hand-rolled limiter with `express-rate-limit`, which the query recognizes. That adds a dependency to the server and changes its limiter behaviour, so it is not something to slip into this branch unasked.

Until then the check is red for a reason that is written down, and `mergeStateStatus` stays UNSTABLE rather than BLOCKED because CodeQL is not a required check.

aks129:
Two notes for the record, neither blocking.

**Required checks are green on b9d76f8.** All eighteen pass, including `postgres-tests`, `node-tests` and `playwright-tests`. CodeQL is the only red one and it is not required, for the reason written up above.

**A rejected token still costs one call, up to the limit.** There is no negative cache on introspection, so an unknown bearer token is asked about once per request until the caller runs out of allowance: at the default that is 120 calls per minute per address. The limiter bounds it, which is what this pull request owes. A short-lived cache of negative answers would bound it further and is worth its own issue rather than a change here.

aks129:
Filed the negative-cache follow-up as #675, so it is tracked rather than carried in this thread.

Issue #417, opened 2026-08-05
care-gaps tells patients their DOB and sex are missing when they are on file, and hides partial results


aks129:
Two defects in `r6/caregaps/report.py`, both found by adversarial QA driving the production caller shape. Both are what PR #393 (#389 half one) left behind: the silence is fixed, the *reasons* are not.

## 1. A false reason: "your date of birth and sex were not available"

Production callers pass no `subject`. On that path the tenant-default fallback **resolves** — `subjectResolution: {"state": "tenant-default", "subject": "Patient/dfe2…"}` — but the route does not hand the resolved Patient to the evaluator. All seven rules go `indeterminate`, and `_unevaluated_reason` (`r6/caregaps/report.py:27-30`) returns `demographics-unavailable`.

The patient reads:

> Your date of birth and sex were not available to this check.

The same Patient carries `birthDate 1985` and `gender female`. With an explicit subject the identical record scores `due=4, up_to_date=2, indeterminate=0`.

The emptiness is honest. The reason is false, and it is actively harmful: it sends someone to re-enter demographics that are already in their record, and it points suspicion at their data rather than at our plumbing.

This is a sharper version of the pattern. We stopped saying "you have no screenings due" and started saying "your record is missing basics". The second is a different false statement about the patient's data, not an absence of one.

## 2. A partial list presented as complete

`r6/caregaps/report.py:88` attaches the `unevaluated` marker only when the consumer list is **entirely** empty:

```python
if not lines:
```

A Patient with a `birthDate` and no `gender` — routine in real feeds — produces:

```
buckets: due=4 up_to_date=0 n/a=1 INDETERMINATE=2 total=7
consumer lines shown: 4
unevaluated marker  : (NONE)
```

The two that vanish are the sex-gated screenings, cervical and mammography. A woman with no `gender` coded sees four screenings and is told nothing about the two that were not checked — and those two are the ones where a missed screening matters most.

"All-or-nothing" is the wrong granularity for a completeness marker. Any indeterminate rule must be visible, not only the case where every rule is indeterminate.

## Fix

1. Hand the resolved Patient to the evaluator on the fallback path, so a resolved subject is actually evaluated. Then `demographics-unavailable` becomes true only when the demographics really are absent.
2. Attach the marker whenever `indeterminate > 0`, naming how many rules were not evaluated and why. A partial answer must say it is partial.

## Proof required

[...]

PR #691, opened 2026-09-10
fix(labs, caregaps): an unscored reading is a statement, and up-to-date means timing (#689)

### LINKED ISSUE #689: labs: an unscored analyte vanishes from the consumer summary, so a panel of nothing-but-unscored reads as nothing wrong
Found while reviewing a recorded clinician demo against the code. The demo showed a patient whose only vitals on file were four consecutive stage 2 blood pressures, the most recent 164/98 — and the assistant's summary of that patient led with "0 high, 0 critical".

## Why it says that

`r6/labs/interpret.py` scores an analyte only when its LOINC is in `LOINC_RANGES`. Blood pressure is not in that table and cannot be — it is a multi-component observation, and the interpreter declines to guess rather than scoring a panel. That is the right call, and `_indeterminate` records it with `flag: None`.

What happens next is the problem.

- `build_interpretation_summary` counts it under `indeterminate`, but **never adds it to `flagged`**, so `flagged` is `[]`.
- `build_consumer_summary` iterates `if r.get("flag")`, so an indeterminate result produces **no line at all**. The analyte disappears.

A caller — a model, a clinician skimming, a patient — reads `high: 0, critical: 0, flagged: []` and a consumer summary with nothing in it. For a panel where every analyte was unscored, "we did not evaluate this" and "we evaluated this and it is fine" are rendered identically. The one the reader takes is the wrong one.

## Why this is the guardrail's own rule

This project already refuses the same shape elsewhere: a screening we cannot fully check is not one you are due for, and care gaps render an explicit unevaluated note rather than silence. "No known allergies is never inferred" is the same principle. An unscored analyte is the lab-interpreter instance of it, and it is currently silent.

It is worse here than in care gaps, because the number that would have prompted a clinician — a stage 2 reading — is on the screen at the same moment the summary says nothing is flagged.

## What to change

- `build_consumer_summary` emits a line for an indeterminate result that names the analyte and says it was not evaluated and why, rather than dropping it.
- The interpretation summary makes an unscored count impossible to miss when it is non-zero — a caller reading only `flagged` must not be able to conclude "nothing to report".
- Neither is a clinical judgement: nothing here starts scoring blood pressure. Declining to score stays; only the silence goes.

## Tests

A panel containing only unscored analytes: the consumer summary is not empty, and it names what was not evaluated. A mixed panel: the unscored analyte appears alongside the scored ones. And a pin that `flagged` alone cannot be read as completeness while `indeterminate` is non-zero.

Related: the care-gaps unevaluated note, and #214's family of "the gate that reads stronger than it is".

aks129:
Closes #689. Both defects were found by reviewing a recorded clinician demo against the code.

## 1. An unscored analyte left no trace

The demo showed a patient whose only readings were four consecutive stage 2 blood pressures, most recent 164/98, and the assistant's summary read "0 high, 0 critical".

Blood pressure is multi-component, so it is not in `LOINC_RANGES` and cannot be; the interpreter is right to decline rather than guess. What was wrong is what followed: an undecided result never entered `flagged`, and `build_consumer_summary` skipped anything without a flag. A panel where nothing could be scored rendered exactly like a panel where everything came back normal, and of those two readings the reader took the wrong one.

- `build_consumer_summary` attaches `unevaluated`, `unevaluated_count`, `unevaluated_analytes` and `unevaluated_note` on **any** undecided result, not only a wholly undecided panel. Care gaps paid for that granularity once already (#417).
- `build_interpretation_summary` names the undecided analytes, so a caller reading `flagged` cannot take it for completeness.
- `_indeterminate` carries `indeterminate_reason`, a slug, so the report layer groups on a fact instead of parsing prose. Same field, same posture as `r6/caregaps/report.py`, whose docstring already cited this module as the pattern it followed.
- The undecided analyte's name comes from `r6/terminology.py` keyed by code — never from the Observation's own `display` or `code.text`. Where the terminology has no label, the note carries the LOINC, which is not PHI.

Nothing here starts scoring blood pressure, and no note says anything about the person: it states what was not evaluated and why, and sends the reader to their clinician.

## 2. "Looks up to date" read as reassurance

Same demo, other surface: the assistant told the patient "your blood pressure and A1c monitoring both look up to date". True about cadence, taken as a verdict about control, and the words were ours. The line now says which it is: *"is up to date on timing (last on …). This checks when it was done, not what it showed."*

## Tests and mutations

Seven pins in the labs report suite and one in care gaps. The four describing new labs behaviour were red before the fix.

| Mutation | Result |
|---|---|
| Drop the marker from the consumer summary | red, 3 tests |
| Attach the marker only when the panel is wholly undecided | red |
| Empty the clinician summary's analyte names | red |
| Name the analyte "an analyte" instead of via terminology | red, 3 tests |
| Count an unscored result as normal | red, 3 tests |
| Drop the "not what it showed" clause | red |

Full suite 3668 passed, 13 skipped, 1 xfailed. `uv run ruff check .` clean.


aks129:
Third commit, after checking the two surfaces that showed the defect rather than only the data behind them.

**The marker arriving is necessary, not sufficient.** The MCP server passes the Flask JSON through whole, so a connected client already receives `unevaluated_note`. CareAgents also receives it, but had no instruction to use it, and care gaps found out the hard way (#417) that a model handed lines plus a note leads with the lines. Handed `high: 0, critical: 0` plus a note, it leads with the zeros — which is the summary on camera. `get_labs` now carries the same shape of instruction `get_care_gaps` does: the answer is incomplete, zero flagged means nothing was *scored* abnormal rather than nothing is abnormal, and an unevaluated result is never called normal. Two tests; disabling the guard turns one red.

Two copy fixes from review: the labs note pluralises its cause for more than one unscored result, and the timing line no longer reads oddly after a vaccine.

Full suite 3670 passed. Ruff clean.

Issue #685, opened 2026-09-07
Board: held branches, fork points and the order they have to land in


aks129:
Shared execution board for Claude Code (builder), Astra (verification and partner evidence), and the founder. Updated September 12, 2026. Verified remote main: f6365aa00bff9c01b33e019b0f16e243c5703ec3.

## Delivery objective
Reach a repeatable, supervised synthetic beta first. General 2.0 release remains a separate gate against the implementation playbook. Recruitment preparation runs alongside engineering; publication, spending, merges and deployment stay with the founder.

## Priority execution order

| Order | Owner | Work | Acceptance / dependency |
| --- | --- | --- | --- |
| P0.1 | Claude builder | Reproduce and resolve existing #679 on current main, one bounded PR | Failing characterization before fix; authorized cases and preview preserved; refusal class tested; mutation caught; required checks pass; Astra reviews exact SHA |
| P0.2 | Claude builder | Resolve existing #681 and #666 as separate form-correctness slices | Sequence shared SDC paths after P0.1; each slice has tests and mutation evidence. Forms stay outside pilot scope until their acceptance recipe passes |
| P0.3 | Claude prepares; founder executes | Release candidate and deployment checklist, tracking #644 and #568 | Exact candidate SHA, configuration prerequisites, migration/rollback steps and smoke commands prepared before approval. Verify deployed SHA afterward |
| P1.1 | Astra | Execute #677 first-user acceptance on the candidate | Sign-in → explicit source → useful read → honest missing-data handling → reopen/history → applicable review artifact. Record statuses only; account data stays private |
| P1.2 | Astra; Claude coordinates environment ownership | Credentialed Medplum and supported OpenClaw acceptance | Verify server identity, authenticated read, refusal and operation-specific audit evidence. Agree sandbox/host ownership first; fixture writes scoped to a test environment. Simulated tests do not complete this gate |
| P1.3 | Claude builder; Astra reviews | Next audit/kernel slice; existing #659 for any included push-approval surface | Use current ratchets and existing issue contracts; unverified actions excluded from pilot. One surface does not prove all surfaces |
| P1.4 | Astra prepares; founder sends | First five usability testers, three engineering partners, two clinician reviewers | Tailored drafts, current destination rules, working synthetic entry point, approved posts/incentives. Measure groups separately |
[...]

PR #672, opened 2026-09-06
qa(mcp): the OAuth walkthrough, A1 to A8 and R1 to R8 as a script that fails loudly (#568, spec §8)

aks129:
## What

The artifact spec §8 names: `services/agent-orchestrator/qa/oauth-walkthrough.sh`, in the pattern of the Aidbox walkthrough. It runs against a deployed pair (Flask with #669 plus the three held branches, the MCP server with #671) under the demo policy, so no person is needed for the authorize step, and no PHI is reachable by construction.

Asserted, each with the guarantee it names:

- **A1** unauthenticated initialize is 401 with `resource_metadata` and `scope`, no `error=`
- **A2/A3** the URL from that header serves the document, readable from a browser, and its `resource` string-equals the URL the client connected to
- **A4** the authorization server's metadata resolves where a client looks for a path-less issuer; `issuer` matches; every endpoint is https; S256; RFC 9207 advertised
- **A5** dynamic registration answers 201
- **A6** authorize is a 302 carrying `code`, `state` and `iss`, read from the Location header; the token endpoint answers a Bearer (and a refresh token, noted when absent)
- **A7** the token initializes a session; tools/list answers; the two privileged tools are withheld
- **A8** a read reaches Flask through the consented tenant; a write-tier tool with a forged step-up argument is refused as read-only. `CALL_EVERY_TOOL=1` walks the read tier with the fixtures the script has and reports called/answered/skipped counts rather than implying coverage it did not measure
- **R1, R2, R3, R5, R8** measured live; **R4, R6, R7** named as deferred to `src/oauth-path.test.ts` with the reason (a token older than its TTL, a boot property, a second deployment with the flag off)
- **§9.4** the preflight answers with an allow-origin and the 401 exposes `WWW-Authenticate`

Exit status is the verdict: any failed assertion exits 1. The spec's §8 gets a one-line pointer.

## Evidence

`bash -n` clean. Not yet executed against a pair: the pair does not exist deployed, and the Mac mini session is asked to run it against its local pair and post the transcript here. That run is the evidence; until it lands this PR is the script and the claim that it asserts, not a claim that anything passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #701, opened 2026-09-12
docs(runbook): enabling the MCP connector on real records, in the owner's order

### LINKED ISSUE #568: MCP authorization phases 2 and 3: a hosted connector still cannot reach a tenant

aks129:
The connector chain is complete on `main` (#669, #695, #697, #698, #670, #671, #672) and switched off in production behind `MCP_OAUTH_ENABLED` and `CAREAGENTS_CONSENT_URL`. Nothing turns it on by itself.

This runbook is the owner's sequence: DNS for the canonical host (#522), the variables per service (the same secret in three places, byte-identical), the two manual deploys with the flag still unset, enable with the demo policy first, then the consent surface, then the spec §8.4 end-user run that closes #568. §5 is the rollback; §6 the three things that look like bugs and are not.

Docs only. The §0 table now records that every code piece has merged; the "what is not done" line is everything below it. Linked from the deploy notes in `docs/development.md`. Link guard green.

Refs #568, #522, board #685 (held branch reconciled: this was `a53cb87` cut from `957847f`, now rebased onto `f6365aa`).


aks129:
**Owner decision needed: which tenant classes may be read through a harness that retains session state without ZDR.**

Context: #723 reviews the OpenAI Agents API, Grok Bot and Meta Muse as clients of the locked MCP server. All three consume the same endpoint and token; the guardrails hold server-side. The one thing the connector cannot settle on its own is data retention after a result leaves us. The Agents API states US-only residency, no Zero Data Retention, and that a self-hosted sandbox does not change that. The claude.ai connector is in the same position for the same reason.

Recommendation, per tenant class:

| Tenant class | Harness that retains state, no ZDR | Note |
|---|---|---|
| Synthetic demo tenant | Yes | already the connector's demo policy |
| Supervised beta on sample records (#718) | Yes | sample records only |
| Real-records tenant | **No, until a BAA covers the harness** | consent page should say so in plain words |

If you rule this way, the consent page (spec §13) needs one line stating which connections a given client may see, and the OAuth mint refuses a real connection for a client without a covering agreement. That is a small follow-up slice on my side; I will not start it without the ruling.

Credentials: no static deployment token into a harness vault or onto a shared cloud computer (Grok Bot shares one machine across all of a user's bots). The OAuth path with the rotating refresh token (#698) is the credential for these surfaces.

aks129:
**Decision table for the owner** (Codex asked that the decision be stated separately from the merged runbook, with the exact configuration each choice implies; nothing below is enabled by this comment).

| Ruling | Who may read through a retaining harness (Agents API, claude.ai connector, Grok Bot, Muse) | Exact configuration if chosen | Residual limitations |
|---|---|---|---|
| **A — as recommended above** | demo tenant and supervised-beta sample tenants only; real-records connections refused for any client without a covering agreement | `MCP_OAUTH_ENABLED=1` + `CAREAGENTS_CONSENT_URL` per the runbook; **one new knob** `MCP_COVERED_CLIENTS` (comma-separated OAuth client ids with a BAA in place; default empty) — the consent page lists only demo/sample connections for a client not on it, and the mint refuses a real connection for that client (small follow-up slice, not started) | a harness can still retain what a *sample* record says; the wording on the consent page is the only thing telling the person which of their connections a client may see |
| **B — demo tenant only** | the synthetic demo tenant; no beta, no real records | `MCP_OAUTH_ENABLED=1`, `CAREAGENTS_CONSENT_URL` **unset** (demo policy stays) | the supervised beta cannot use a connector client at all; #718's harness rows are moot until changed |
| **C — real records permitted with consent text** | any connection the person consents to, with the retention sentence on the consent page | runbook as written, no new knob | HIPAA exposure sits with the harness vendor's retention terms, not with anything we enforce; not recommended without a BAA |

Common to all: no static deployment token in any harness vault or shared cloud computer; the OAuth path with the rotating refresh token (#698) is the credential. Retention terms are the vendor's as published on 14 Sep (Agents API: US-only, no ZDR) and should be re-read before enabling.

The implementation side (#701 runbook) is merged; this table is the open decision.

PR #706, opened 2026-09-12
fix(mcp): a rejected token is remembered, so a stale one costs one round trip per window (#675)

### LINKED ISSUE #675: MCP OAuth: a rejected token is introspected once per request, with no negative cache
Follow-up from #671, which fixed the ordering defect but not this.

The MCP server resolves a bearer token by asking the authorization server's introspection endpoint. Live grants are cached. A **rejected** answer is not, so an unknown or expired token costs one request to Flask every time it is presented, up to the rate limit: 120 per minute per address at the default.

That is bounded now, because the limiter runs above the credential check (#671). It is still an asymmetry worth closing: a client looping on a stale token, or a scanner walking the endpoint, turns one useless request into one useless request plus one authorization-server round trip.

Shape of the fix: cache the negative answer under the same key the positive one uses, with a short time to live, well under the shortest token lifetime so a token that becomes valid is not refused for longer than that window. A token cannot go from inactive back to active in the current design, but a cache that assumes so should say why rather than rely on it.

Wants a test that a second presentation of a rejected token makes no second call, and one that the cache expires.

Refs #671, #568.

aks129:
Closes #675. Refs #671, #568.

## The asymmetry

Live grants were cached under the token's SHA-256; a rejected answer was not, so a client looping on a stale token or a scanner walking the endpoint turned one useless request into one useless request plus one authorization-server round trip, up to the rate limit (120/min per address). Bounded since #671 put the limiter above the credential check; still worth closing.

## The fix

The negative answer is remembered under the same key for **thirty seconds** — well under the shortest token lifetime (`OAUTH_TOKEN_TTL`, 3600s by default), so a token that becomes valid is refused for at most that window. A token cannot go from inactive back to active in the current design; the comment says so, and the cache does not rely on it.

One distinction the issue leaves implicit, made explicit: `introspect()` now returns the grant, `"rejected"` (the server's answer, or a token failing a condition) or `"unavailable"` (not configured, unreachable, not 2xx, unparseable). Both are 401 to the caller; only `"rejected"` is remembered. Otherwise an outage blip would refuse a good token for the whole window — fail closed now, ask again next time.

## Tests (`oauth-path.test.ts`)

- a second presentation of a rejected token makes no second call
- the memory expires (Date.now stubbed past the window; the token is then accepted, with a second call)
- an authorization server that could not be asked is not remembered as a no

The limiter test now presents a distinct token per request, since a repeated one no longer costs a call; its property (past the limit, no introspection) is unchanged.

**Mutations**, each compiling (an env-var guard, per the trap in #702), tests ran: never remember → 1 failed / 29 passed; remember the unavailable too → 1 failed / 29 passed. `tsc --noEmit` clean; full node suite green.


aks129:
Note on `node-tests`: the suite flakes in full parallel runs on `main` too (2 of 15 runs at the current node tree, same two limiter tests; on this branch the rate was the same). Filed as #707 with the measurements; a red `node-tests` here is a rerun, not this change. The limiter test's change to a distinct token per request is unrelated to that flake and is required by this PR.

PR #731, opened 2026-09-14
feat(action-rail): a person who says no is recorded as one, not as a timeout (#520)

### LINKED ISSUE #520: actions: a human who declines and a human who never looked produce the same record

aks129:
## What

`awaiting_confirmation` could only become `executing` or `expired`, so a refusal and a lapse produced the same record: the trail could prove nobody approved, never that somebody objected (#520, human-gate spec §4.4).

- `declined`: terminal, reachable only from `awaiting_confirmation`, and only through the new route. Never inferred: a lapsed window stays `expired` even when Decline is tapped after the fact.
- `POST /r6/actions/<id>/decline`: the same action-bound, payload-bound, single-use credential as confirm (`require_grant` with `ACTION_APPROVAL_AUDIENCE` and `approval_operation`, nonce consumed), so only the surface that showed the person the card can record their answer, and a declined credential cannot be replayed to approve. Body `{declined_via}` over the same surface vocabulary as `approved_via`. Guarded transition with the expiry predicate; audited PHI-free (surface + payload digest) via `add_audit_event` in the same session, per the post-commit-shim ratchet.
- The review page's status poll gains a `declined` arm (the drift test requires every producible status to have one); the MCP `action_status` description lists the word and says never to re-propose unasked.
- Matrix row and non-ASCII-token row for the new gate.

Not in this slice, on purpose: the Decline button and its CareAgents relay (the review page posts through the relay, which holds the credential). That is a consumer-app slice on top of this state; the spec's `ActionConfirmation`-shaped refusal record is a schema change that should ride with the approve page (#215), and the `ActionEvent` + AuditEvent here already carry who-via/when.

Closes #520.

## Tests

`tests/actions/test_decline_is_a_state.py`: transition table; decline recorded as decline with a PHI-free audit row and no dial; declined action cannot then be approved (409); the credential is spent (replay to confirm is 401); a generic write token cannot decline (401); a lapsed window is `expired` not `declined` (410); decline before commit conflicts (409); unknown surface 400.

## Mutation evidence

- Route records `expired` instead of `declined`: 2 failed.
- Gate weakened to a generic write token: 2 failed.
- Nonce not consumed: 1 failed.
- Restored: 3921 passed, 44 skipped, 1 xfailed. Ruff clean; node status tests 5 passed in band.


aks129:
Two fix-ups pushed after review: (1) `3abc832` regenerates the MCP tool manifest (the `action_status` description changed) and lifts the landing page's test figure, which CI caught; (2) the decline route now mirrors confirm's TOCTOU branch: a window that lapses between the snapshot check and the guarded claim is flipped to `expired` and answered 410, not 409 naming the same state twice; pinned with the `is_expired → False` pattern from `test_expired_row_cannot_win_commit_claim`. Mutation: branch removed → 1 failed.

